Issue #107, opened 2023-04-03
Support more dynamic secret paths


adrianmoisey:
**Describe the bug**
The Vault Secrets Operator fails to work on dynamic secret paths other than AWS.
The operator has "/creds/" hardcoded in the path: https://github.com/hashicorp/vault-secrets-operator/blame/f617b4aa2f2572414d7039774ee1170f8f6e3a08/controllers/vaultdynamicsecret_controller.go#L209

Some backends, such as GCP, don't use "/creds/"

**To Reproduce**
1. Configure Vault with a GCP backend
2. Install and configure the Vault Secrets Operator
3. Configure a "VaultDynamicSecret" CR pointing at a secret in Vault
5. See error in the vault-operator logs:

```
2023-04-03T14:49:40Z	ERROR	Reconciler error	{"controller": "vaultdynamicsecret", "controllerGroup": "secrets.hashicorp.com", "controllerKind": "VaultDynamicSecret", "VaultDynamicSecret": {"name":"vaultdynamicsecret-sample","namespace":"default"}, "namespace": "default", "name": "vaultdynamicsecret-sample", "reconcileID": "1321a336-0f52-4f27-9e62-2fa69fc28e41", "error": "Error making API request.\n\nURL: GET https://vault.default.svc.cluster.local:8200/v1/gcp/creds/roleset/bucket/token\nCode: 403. Errors:\n\n* 1 error occurred:\n\t* permission denied\n\n"}
``` 

The main part of that output is the following:
> Error making API request.\n\nURL: GET https://vault.default.svc.cluster.local:8200/v1/gcp/creds/roleset/bucket/token\nCode: 403. Errors:\n\n* 1 error occurred:\n\t* permission denied\n\n"

This is due to the GCP backend not having a secret on the path "gcp/creds/roleset/bucket/token".
The problem is that "creds" is injected into the path.

Application deployment:

Deployed using the Helm chart
```yaml
apiVersion: secrets.hashicorp.com/v1alpha1
kind: VaultConnection
metadata:
  name: vaultconnection-sample
  namespace: default
spec:
  address: https://vault.default.svc.cluster.local:8200
  caCertSecretRef: vault-tls
---
apiVersion: secrets.hashicorp.com/v1alpha1
kind: VaultAuth
metadata:
  name: vaultauth-sample
  namespace: default
spec:
  vaultConnectionRef: vaultconnection-sample
  method: kubernetes
  mount: kubernetes
  kubernetes:
[...]

PR #108, opened 2023-04-03
Remove hardcoded "creds" from path

adrianmoisey:
Related issue: https://github.com/hashicorp/vault-secrets-operator/issues/107

Not all secrets engines have "creds" in their path.

I honestly don't know if the approach is the correct one.

I assume another approach would be to detect the engine, and then dynamically figure out which path is needed.
I don't know how possible that is.

### DIFF
--- a/controllers/vaultdynamicsecret_controller.go
+++ b/controllers/vaultdynamicsecret_controller.go
@@ -206,7 +206,7 @@ func (r *VaultDynamicSecretReconciler) isRenewableLease(resp *secretsv1alpha1.Va
 }
 
 func (r *VaultDynamicSecretReconciler) syncSecret(ctx context.Context, vClient vault.Client, o *secretsv1alpha1.VaultDynamicSecret) (*secretsv1alpha1.VaultSecretLease, error) {
-	path := fmt.Sprintf("%s/creds/%s", o.Spec.Mount, o.Spec.Role)
+	path := fmt.Sprintf("%s/%s", o.Spec.Mount, o.Spec.Role)
 	resp, err := vClient.Read(ctx, path)
 	if err != nil {
 		return nil, err


adrianmoisey:
I'm wondering if a better approach would be to replace "Mount" and "Role" (these: https://github.com/hashicorp/vault-secrets-operator/blob/f617b4aa2f2572414d7039774ee1170f8f6e3a08/api/v1alpha1/vaultdynamicsecret_types.go#L20-L22) with a single Path string?

kschoche:
Hi @adrianmoisey !
Thanks for submitting this PR, I think you have a very valid point about not assuming the `/creds/` path and that should be fixed. I know that with the current code in this PR we're going to fail the integration tests because they are assuming the `/creds/` path as well in the terraform code. But I think that is just something we'll fix once we reach a consensus on the approach.

I'm digging into this internally to see what we can do within the CRD to maybe unify the various dynamic secrets engines and their paths under this particular schema. I was wondering about introducing a secret engine type that we could base the scheme off of, or as you mentioned maybe just using the entire path without the separate role field would be sufficient? Any other ideas?

kschoche:
Hey @adrianmoisey - 
What do you think of this [approach](https://github.com/hashicorp/vault-secrets-operator/compare/VAULT-15169/add_path_to_VDS_CRD?expand=1)? I was even thinking of removing Mount+Role like you mentioned, but left it for now.
Given we are in a beta stage, we have a bit of freedom to help flush out these pain points so feedback on this is much appreciated!

adrianmoisey:
> What do you think of this [approach](https://github.com/hashicorp/vault-secrets-operator/compare/VAULT-15169/add_path_to_VDS_CRD?expand=1)? I was even thinking of removing Mount+Role like you mentioned, but left it for now.

That looks good! That's more in line with what I'd expect, and it should work with every secrets engine.
I think it's worth considering removing Mount+Role at some stage, having Mount, Role and Path all required fields is a bit confusing. However, adding Path unblocks me for my use case, which is nice.

Something that's worth noting, is that all the other popular tools I've looked at (vault-agent, external-secrets, Vault CSI Provider, etc) use a path to determine which secret to fetch. 
I think that "path" is the best option going forward.

> Given we are in a beta stage, we have a bit of freedom to help flush out these pain points so feedback on this is much appreciated!

No worries, happy to help!

ameyp:
@kschoche Any updates?

kschoche:
> @kschoche Any updates?

Hi @ameyp - thanks for the checking back in, we are actively working on this and should have a PR up for it soon!

Cheers,
~Kyle

PR #110, opened 2023-04-03
Test with k8s 1.26 and latest Vault versions

tvoran:
Also remove the extra go caching steps in favor of setup-go@v3's [built-in caching support](https://github.com/actions/setup-go/tree/v3.5.0#caching-dependency-files-and-build-outputs).


kschoche (approved):
Awesome!

Issue #99, opened 2023-03-31
Avoid waiting for last lease to regenerate dynamic credentials


dnlopes:
**Is your feature request related to a problem? Please describe.**

When using dynamic secrets (e.g., MySQL credentials), I have both defined a `lease-time` and `TTL` (let's assume a `1m` lease, `60m` TTL). The Operator will keep leasing without issues, until eventually it receives a `400 BadRequest` because he can't lease anymore. At this point, it will request new credentials and redeploy the app (I'm using the `rolloutRestartTargets` setting for this).

To me, this has potential to cause downtime if the app needs those credentials during the timeframe in which the TTL was reached and the app is being redeployed.

**Describe the solution you'd like**

I think the Operator has all the information needed to detect that the next lease will fail. When it detects it, he could generate a new credential and order the redeploy of the apps.

This would ensure that at any point in time, apps were using expired credentials.

Another improvement would be to also be able to apply this behavior earlier, to avoid waiting for the last lease. For instance, one could have a setting to force the credential regeneration at 75% of the TTL. This would give some extra slack if for some reason Vault is unable to generate the new credentials (it would have 25% of the TTL time to retry the operation, for instance).

**Describe alternatives you've considered**

None.

**Logs**
```
2023-03-31T11:55:15Z    DEBUG    events    Renewed lease, lease_id=database/creds/demo/nLJW1XkrEXftmUumvV83U8M0, horizon=841.645753ms    {"type": "Normal", "object": {"kind":"VaultDynamicSecret","namespace":"demo","name":"vault-dynamic-secret-1","uid":"5b6f3ab3-9cd4-4d27-bb44-e82179778df1","apiVersion":"secrets.hashicorp.com/v1alpha1","resourceVersion":"762"}, "reason": "SecretLeaseRenewal"}
2023-03-31T11:55:15Z    DEBUG    events    Could not renew lease, lease_id=database/creds/demo/nLJW1XkrEXftmUumvV83U8M0, err=Error making API request.

URL: PUT http://host.docker.internal:8200/v1/sys/leases/renew
Code: 400. Errors:

[...]

PR #112, opened 2023-04-04
fix: avoid reconciling after lease expires

dnlopes:
This PR is a first attempt to address issue #99. 

I removed the `if leaseDuration < 1 {` because of the new validation afterwards, I'm not 100% sure it's safe though.

As I said in the original issue, I wonder if this change is enough, maybe not; it seems a bit fragile. But please, edit this PR as you see fit.

I would really like to have this change in the component to better streamline the adoption of dynamic credentials without hitting runtime issues because of expired credentials.


dnlopes:
@tvoran @benashz @adrianmoisey 

Can anyone shed some light on this issue? This PR is a tentative fix for #99.

**I'm totally fine if you decide this is not the correct approach**, but I would like to understand if there are plans in the near future to address the issue I raised on #99, just to manage my expectations, as this issue is really a blocker for my use case of dynamic credentials.

Thanks.

dnlopes:
Closing this in favor of #170.

PR #114, opened 2023-04-05
fix: docker cache improvement

blakepettersson:
The build arg `LD_FLAGS` gets dynamically computed on each invocation of `docker build`. This implies that `go mod download` always gets run, even if no dependencies have changed.

By moving `LD_FLAGS` just before `go build`, we ensure that the Docker layer caching is used and that Go dependencies does not unnecessarily have to be downloaded every time.


tvoran (approved):
Thanks!

PR #116, opened 2023-04-06
Run integrationTest as part of the build workflow 

tvoran:
This way the integration tests use the docker image that would be promoted through to staging and production, instead of building one just for the tests.

This will make the build workflow take more time of course, and if it becomes too long to wait for at release time we may want to use a shorter "smoke test" style job here instead.

Also updates the `if: ${{ matrix.vault-version == '1.13.1' }}` check for OSS tests, and added it to the Helm OSS job.

Updates kind v0.17.0 => v0.18.0

Sets the image tag based on the env variable `VERSION` in the build workflow for the integration tests, so that the tests are using that instead of the default from values.yaml or config/manager/kustomization.yaml.


kschoche (approved):
This is a great change, thanks!

tvoran:
I added a bit more here since I realized the integration tests were still running on the default image set in values.yaml and config/manager/kustomization.yaml for the helm and kustomize tests respectively.

tvoran:
Thanks!

PR #127, opened 2023-04-12
Check version of the binary in the image

tvoran:
As part of the `build` workflow, after building each docker image, run the `--version` command and check that the output matches the expected version.

Also adds yaml and json output modes for `--version` as a `--output` option.


.github/workflows/build.yaml
```diff
@@ -133,3 +133,14 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
             public.ecr.aws/hashicorp/${{env.repo}}:${{env.version}}
+      - name: Check binary version in container
+        shell: bash
+        run: |
+          versionOutput=$(docker run hashicorp/${{env.repo}}:${{env.version}} --version --output=json)
```
benashz, line 139: nit: I think we normally use snake-case for shell code.

benashz (approved):
Looks great! Just one minor nit.

PR #130, opened 2023-04-12
PKI: ensure alt_names are in the proper form

### LINKED ISSUE #115: VaultPKISecret AltNames not of type string with Helm templating
**Describe the bug**
`2023-04-06T09:34:29Z	ERROR	Reconciler error	{"controller": "vaultpkisecret", "controllerGroup": "secrets.hashicorp.com", "controllerKind": "VaultPKISecret", "VaultPKISecret": {"name":"front-certificate","namespace":"ns"}, "namespace": "ns", "name": "front-certificate", "reconcileID": "c64d90ee-c03e-4cc9-a275-f4a9a08c0c3e", "error": "Error making API request.\n\nURL: PUT http://platform-vault.ns.svc.cluster.local:8200/v1/pki/issue/redacted\nCode: 400. Errors:\n\n* Field validation failed: error converting input [redacted.fqdn] for field \"alt_names\": '' expected type 'string', got unconvertible type '[]interface {}', value: '[redacted.fqdn]'"}`

**To Reproduce**
1. Deploy application with the following yaml file with the following VSO custom resources as a template in a Helm dependency.
```yaml
---
kind: VaultConnection
metadata:
  namespace: ns
  name: vso-connection
spec:
  address: http://platform-vault.ns.svc.cluster.local:8200
---
kind: VaultAuth
metadata:
  namespace: ns
  name: vso-auth
spec:
  vaultConnectionRef: vso-connection
  method: kubernetes
  mount: kubernetes
  kubernetes:
    role: default
    serviceAccount: default
---
kind: VaultPKISecret
metadata:
  namespace: ns
  name: front-certificate
spec:
  vaultAuthRef: vso-auth
  mount: pki
  name: redacted
  commonName: {{ (.Values.global).fqdn | quote }}
  altNames:
    - {{ (.Values.global).fqdn | quote }}
  format: pem
  privateKeyFormat: pem
  expiryOffset: 1s
  ttl: 60s
  namespace: ns
  revoke: true
  clear: true
  destination:
    create: true
    name: front-cert
    type: kubernetes.io/tls
```
2. Look into operator logs to find above error.

**Expected behavior**
The file is ok and the secret is created.

**Environment**
* Kubernetes version: 
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): k3s
  * Other configuration options or runtime services (istio, etc.): Helm dependency, deployed by ArgoCD
* vault-secrets-operator version: 0.1.0-beta

Don't hesitate if anything is unclear!
Thanks in advance and have a great day :)

benashz:
Previously we were passing alt_names as a string slice to Vault, which was invalid, since Vault expects a CSV string.

Closes #115


api/v1alpha1/vaultpkisecret_types.go
```diff
@@ -130,7 +132,6 @@ type VaultPKISecret struct {
 func (v *VaultPKISecret) GetIssuerAPIData() map[string]interface{} {
 	m := map[string]interface{}{
 		"common_name":          v.Spec.CommonName,
-		"alt_names":            v.Spec.AltNames,
 		"ip_sans":              v.Spec.IPSans,
 		"uri_sans":             v.Spec.URISans,
 		"other_sans":           v.Spec.OtherSans,
```
tvoran, line 137: It looks like we may need to do the same for ip_sans, uri_sans, and other_sans, since they also need to be comma-separated lists: https://developer.hashicorp.com/vault/api-docs/secret/pki#ip_sans

And on a related note, `VaultPKISecretSpec.OtherSans` may also need to be updated to be a `[]string` in vaultpkisecret_types.go, to follow the other fields.

api/v1alpha1/vaultpkisecret_types.go
```diff
@@ -139,6 +140,10 @@ func (v *VaultPKISecret) GetIssuerAPIData() map[string]interface{} {
 		"exclude_cn_from_sans": v.Spec.ExcludeCNFromSans,
 	}
 
+	if len(v.Spec.AltNames) > 0 {
```
tvoran, line 143: It looks like strings.Join is safe to use with 0-length elements since it checks the len of the items, so this check probably isn't necessary.

benashz: Good call!

api/v1alpha1/vaultpkisecret_types.go
benashz: Good call!

api/v1alpha1/vaultpkisecret_types.go
benashz: Fixed in https://github.com/hashicorp/vault-secrets-operator/commit/3a3be5f080719d6de65a0366ce4531b539eab877

kschoche (approved):
Thanks for fixing this!
I think `other_sans` also takes a string array, however the vaultpkisecret_type crd treats it as a string, instead of a `[]string` like ip_sans/uri_sans/alt_names, I was thinking it might make sense for them to all be the same data types (`[]string`) in the crd spec, and then apply the same logic also to other_sans here as well? I suppose that could be another PR. Thoughts?

tvoran:
@kschoche Yeah I had the same thought about `other_sans`; while we're still in beta we might be able to change it without changing the apiVersion, needing a conversion, etc.

benashz:
> @kschoche Yeah I had the same thought about `other_sans`; while we're still in beta we might be able to change it without changing the apiVersion, needing a conversion, etc.

@tvoran I am working on a PR for that change now. I have a few other "breaking" changes forthcoming as well. I am going to look to see how easy the conversion/apiVersion bump is to do. It may not be worth it at, since as you say we are still in beta, but could be good practice for the future.

api/v1alpha1/vaultpkisecret_types.go
```diff
@@ -139,6 +140,10 @@ func (v *VaultPKISecret) GetIssuerAPIData() map[string]interface{} {
...
+		m["alt_names"] = strings.Join(v.Spec.AltNames, ",")
```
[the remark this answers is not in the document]
benashz: Will do that in a follow-on PR.

PR #131, opened 2023-04-12
feat: add support for the JWT auth method

thyton:
- implement a new [CredentialProvider()](https://github.com/hashicorp/vault-secrets-operator/blob/f617b4aa2f2572414d7039774ee1170f8f6e3a08/internal/vault/credential_provider.go#L28) interface
- add a VaultAuth CR helm template for JWT
- add vault auth integration tests


internal/vault/credentialproviders/jwt.go
```diff
@@ -0,0 +1,107 @@
...
+		Name:      l.authObj.Spec.Kubernetes.ServiceAccount,
+	}
+	sa := &corev1.ServiceAccount{}
+	if err := client.Get(ctx, key, sa); err != nil {
+		return nil, err
+	}
+	return sa, nil
+}
+
+func (l *JwtCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.Client) (map[string]interface{}, error) {
```
tvoran, line 67: Will this support grabbing a jwt token from a k8s secret as well?

internal/vault/credentialproviders/kubernetes.go
```diff
@@ -0,0 +1,109 @@
...
+
+func NewKubernetesCredentialProvider(authObj *secretsv1alpha1.VaultAuth, providerNamespace string,
+	uid types.UID,
+) *KubernetesCredentialProvider {
+	return &KubernetesCredentialProvider{
+		authObj,
+		providerNamespace,
+		uid,
+	}
+}
```
tvoran, line 33: I wonder if this function is necessary, since KubernetesCredentialProvider is already public?

api/v1alpha1/vaultauth_types.go
```diff
@@ -44,6 +61,8 @@ type VaultAuthSpec struct {
 	Headers map[string]string `json:"headers,omitempty"`
 	// Kubernetes specific auth configuration, requires that the Method be set to kubernetes.
 	Kubernetes *VaultAuthConfigKubernetes `json:"kubernetes,omitempty"`
+	// Jwt specific auth configuration, requires that the Method be set to jwt/oidc.
+	Jwt *VaultAuthConfigJwt `json:"jwt,omitempty"`
```
tvoran, line 65: We'll probably want to add `jwt` to the validation enum for `Method` above.

tvoran (commented):
Left some thoughts, but this is coming along nicely!

tvoran:
BTW I think the `build-pre-checks` job is failing because `make manifests` is generating some uncommitted changes.

internal/vault/credentialproviders/kubernetes.go
thyton: All fields from `KubernetesCredentialProvider` are currently private. There were no usage of referencing fields when the struct was private, prior to the PR. I also see only [tests](https://github.com/hashicorp/vault-secrets-operator/pull/131/files#diff-43563f2031dc36837155f26a710e12785ebba21517a705364a25bdb08a24cf62) need to instantiate with custom values for the fields. I wonder if adding `SetUID()` would make more sense since only `uid` is used to create the object.

internal/vault/credential_provider.go
```diff
@@ -7,23 +7,23 @@ import (
...
 	"fmt"
 
-	authv1 "k8s.io/api/authentication/v1"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	secretsv1alpha1 "github.com/hashicorp/vault-secrets-operator/api/v1alpha1"
+	"github.com/hashicorp/vault-secrets-operator/internal/vault/credentialproviders"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/utils/pointer"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
thyton: Thanks!

api/v1alpha1/vaultauth_types.go
```diff
@@ -25,6 +25,23 @@ type VaultAuthConfigKubernetes struct {
 	TokenExpirationSeconds int64 `json:"tokenExpirationSeconds,omitempty"`
 }
 
+// VaultAuthConfigJwt provides VaultAuth configuration options needed for authenticating to Vault.
+type VaultAuthConfigJwt struct {
+	// Role to use for authenticating to Vault.
+	Role string `json:"role"`
+	// Token to use when authenticating to Vault's JWT authentication backend.
+	Token string `json:"token,omitempty"`
```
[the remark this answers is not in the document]
thyton: I updated it to a Secret reference

internal/vault/credentialproviders/jwt.go
thyton: Yes, it will.

chart/templates/vault-auth-method-jwt.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: jwt
+  namespace: {{ .Release.Namespace }}
+  labels:
+    control-plane: controller-manager
+    component: controller-manager
+    app.kubernetes.io/component: controller-manager
+  {{- include "chart.labels" . | nindent 4 }}
+spec:
+  namespace: {{ .Values.jwtAuthMethod.namespace }}
+  method: {{ .Values.jwtAuthMethod.method }}
```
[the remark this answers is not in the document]
thyton: I removed`vault-auth-method-jwt.yaml` due to moving jwt spec under default-vault-auth-method.yaml.

internal/vault/credentialproviders/kubernetes.go
tvoran: Oh yep, sorry, I missed that this was moved to a new package.

api/v1alpha1/vaultauth_types.go
```diff
@@ -25,6 +26,28 @@ type VaultAuthConfigKubernetes struct {
 	TokenExpirationSeconds int64 `json:"tokenExpirationSeconds,omitempty"`
 }
 
+// Token represents the token if present in a VaultAuthConfigJwt.
+type Token struct {
+	ValueFrom *corev1.EnvVarSource `json:"valueFrom"`
```
tvoran, line 31: [EnvVarSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#envvarsource-v1-core) supports more sources than just secrets, so I wonder if we should restrict it to only secrets? It feels a little odd to me to allow tokens to be retrieved from non-secrets.

Perhaps this could just be our own object that takes a secret name and key within the secret?

thyton: Thanks! I replaced field Token for `TokenSecretKeySelector` with type `*corev1.SecretKeySelector`

api/v1alpha1/vaultauth_types.go
```diff
@@ -25,6 +26,24 @@ type VaultAuthConfigKubernetes struct {
 	TokenExpirationSeconds int64 `json:"tokenExpirationSeconds,omitempty"`
 }
 
+// VaultAuthConfigJwt provides VaultAuth configuration options needed for authenticating to Vault.
+type VaultAuthConfigJwt struct {
+	// Role to use for authenticating to Vault.
+	Role string `json:"role"`
+	// TokenSecretKeySelector to use when referencing the secret containing the JWT token
+	// to authenticate to Vault's JWT authentication backend.
+	TokenSecretKeySelector *corev1.SecretKeySelector `json:"tokenSecretKeySelector,omitempty"`
```
thyton, line 35: Should  I rename `TokenSecretKeySelector` to `SecretKeyRef` as in [EnvVarSource](https://pkg.go.dev/k8s.io/kubernetes@v1.27.1/pkg/apis/core#EnvVarSource)?
```
        // Selects a key of a secret in the pod's namespace.
	SecretKeyRef *SecretKeySelector

File: internal/vault/credentialproviders/jwt.go [hunk 9]
```diff
@@ -0,0 +1,135 @@
...
+		// credentials needed for Jwt auth
+		return map[string]interface{}{
+			"role": l.authObj.Spec.Jwt.Role,
+			"jwt":  tr.Status.Token,
+		}, nil
+	}
+
+	return map[string]interface{}{
+		"role": l.authObj.Spec.Jwt.Role,
+		"jwt":  string(l.tokenSecret.Data[l.authObj.Spec.Jwt.TokenSecretKeySelector.Key]),
```
tvoran, line 113: If the token secret is updated, will this pick it up? I wonder if we should fetch the token secret in `GetCreds()` too.

internal/vault/credentialproviders/jwt.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+		Status: authv1.TokenRequestStatus{},
+	}
+
+	if err := client.SubResource("token").Create(ctx, sa, tr); err != nil {
+		return nil, err
+	}
+
+	return tr, nil
+}
```
tvoran, line 135: Since this is pretty similar to KubernetesCredentialProvider.requestSAToken(), should we move this into a common function that both can call?

internal/vault/cache_metrics_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hashicorp/vault-secrets-operator/internal/vault/credentialproviders"
```
tvoran, line 10: Looks like this one escaped the local import block.

internal/vault/credentialproviders/jwt.go
thyton: Thanks. I refactored both to call new `requestSAToken()` from `credentialproviders/helpers.go`

internal/vault/credentialproviders/jwt.go
thyton: This is fixed to fetch the token in `GetCreds()`.

tvoran (approved):
Nice work!

tvoran:
And don't forget a changelog entry 🙂

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -0,0 +1,223 @@
...
+			"VaultStaticSecret", secret)
+	}
+
+	for x, tt := range auths {
+		t.Run(tt.Spec.Method, func(t *testing.T) {
+			putKV(t, secrets[x])
+			require.Nil(t, crdClient.Create(ctx, secrets[x]))
+			assertSync(t, secrets[x])
+			t.Cleanup(func() {
+				deleteKV(t, secrets[x])
```
benashz, line 219: I don't see where we are cleaning up the `VaultStaticSecrets`, we should probably add that if it is not already being done somewhere else.

benashz (commented):
Looks great! Just have a few comments/nits to take a look at.

internal/vault/credentialproviders/helpers.go
```diff
@@ -32,3 +32,17 @@ func requestSAToken(ctx context.Context, client ctrlclient.Client, sa *corev1.Se
 
 	return tr, nil
 }
+
+// getSecret for the provided namespace and name
```
benashz, line 36: 👍

internal/vault/credential_provider.go
```diff
@@ -7,23 +7,24 @@ import (
...
 	providerMethodKubernetes string = "kubernetes"
+	providerMethodJwt        string = "jwt"
 )
 
-var providerMethodsSupported = []string{providerMethodKubernetes}
+var (
+	providerMethodsSupported = []string{providerMethodKubernetes, providerMethodJwt}
+
+	_ CredentialProvider = (*credentialproviders.KubernetesCredentialProvider)(nil)
+	_ CredentialProvider = (*credentialproviders.JwtCredentialProvider)(nil)
```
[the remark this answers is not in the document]
thyton: If we do the declaration in the provider package, we will have a circular import issue as `NewCredentialProvider()` references individual provider structs in the provider package.

I think the work around could be moving `credential_provider.go` to the provider package and possibly renaming the file to `interface.go`. Let me know how that sounds

test/integration/vaultauthmethods_integration_test.go
thyton: Thanks for catching that!

benashz (approved):
Nice work!

PR #143, opened 2023-04-17
Adds support for revocation of VDS secrets on delete.

### LINKED ISSUE #125: Drop AWS IAM_USER when the VaultDynamicSecret is deleted
**Describe the bug**
When i delete the VaultDynamicSecret that is linked to an AWS Role backend of type `iam_user`, the AWS user is not deleted.
And i can't re-create the resource before the lease period.

**To Reproduce**
1. create a vault_aws_backend_role with type `iam_user`
2. generate a  dynamic secret with the operator
3. wait the K8s secret creation
5. Delete the Dynamic secret and the K8s secret

The AWS User is not deleted

**Expected behavior**
can we have an option like provided on the Vault agent `vault.hashicorp.com/agent-revoke-on-shutdown` to revoke Dynamic credential when it is deleted.

**Environment**
* Kubernetes version:  1.22.2
* vault-secrets-operator version: 0.11.0

kschoche:
Adds support for revocation of VDS secrets on delete:
* Adds the field `Revoke` to the VDS secret Spec.
* Adds logic to revoke the VDS secret on deletion through a finalizer.
* Updates tests to validate revocation of VDS secrets.

Note: Introducing this behaviour *requires* the following policy added to each VaultAuthMethod which is referenced by VDS secrets that enable revocation:
```
path "sys/leases/revoke" {
    capabilities = ["update"]
}
```
In the absence of the above policy, revocation will fail on VDS secret deletion.
This should fix #125


controllers/vaultdynamicsecret_controller.go
```diff
@@ -84,14 +84,29 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if o.Spec.Revoke {
+		if o.GetDeletionTimestamp() == nil {
+			if err := r.addFinalizer(ctx, o); err != nil {
+				return ctrl.Result{}, err
+			}
+		} else {
+			logger.Info("Got deletion timestamp", "obj", o)
+			if err := r.handleDeletion(ctx, o); err != nil {
+				msg := "Failed to handle deletion"
+				r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonSecretLeaseRevoked, msg+": %s", err)
```
tvoran, line 101: I wonder if it would be better to use a reason that's not in the past tense here, since in this err handling the lease may not have been revoked yet. And it looks like this error could be either from removing the finalizer or revoking the lease.

It may be useful to emit an event on successful revoke too.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -84,14 +84,29 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+				return ctrl.Result{}, err
+			}
```
tvoran, line 103: Since `handleDeletion()` also checks `Spec.Revoke`, I wonder if the else clause should be out on it's own (instead of under the `if o.Spec.Revoke` here).

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -20,6 +20,8 @@ type VaultDynamicSecretSpec struct {
 	Mount string `json:"mount"`
 	// Role in Vault to get the credentials for.
 	Role string `json:"role"`
+	// Revoke the credential when the resource is deleted.
```
tvoran, line 23: Should we also revoke credentials when they're rotated if Revoke==true?

kschoche: Oh that is a great question, let me look into that. I had assumed it'd just time out and that would be okay but we should probably also include that.

kschoche: Updated in 451a3268045f475de39ef228db34c225ff4b5512

controllers/vaultdynamicsecret_controller.go
kschoche: Good point thanks, fixed in 451a3268045f475de39ef228db34c225ff4b5512

controllers/vaultdynamicsecret_controller.go
```diff
@@ -288,3 +304,55 @@ func isLeaseNotfoundError(err error) bool {
...
+	return nil
+}
+
+// revokeLease revokes the VDS secret's lease.
+// NOTE: Enabling revocation requires the VaultAuthMethod referenced by `o.Spec.VaultAuthRef` to have a policy
+// that includes `path "sys/leases/revoke" { capabilities = ["update"] }`, otherwise this will fail with permission
+// errors.
+func (r *VaultDynamicSecretReconciler) revokeLease(ctx context.Context, o *secretsv1alpha1.VaultDynamicSecret, id string) error {
+	logger := log.FromContext(ctx)
+	logger.Info("Revoking lease for credential ", "id", o.Status.SecretLease.ID)
```
tvoran, line 337: When `id != ""`, this ends up logging that the new credential is going to be revoked, instead of the old one, so we should probably define `leaseID` before and use it in this log call.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -288,3 +304,55 @@ func isLeaseNotfoundError(err error) bool {
...
+// handleDeletion will handle the deletion path of the VDS secret:
+// * revoking any associated outstanding leases
+// * removing our finalizer
+func (r *VaultDynamicSecretReconciler) handleDeletion(ctx context.Context, o *secretsv1alpha1.VaultDynamicSecret) error {
+	logger := log.FromContext(ctx)
+	if err := r.revokeLease(ctx, o, ""); err != nil {
+		// Do not return error, otherwise we may fail to remove the finalizer.
+		// Worst case at this point we will leave a dangling lease instead of a secret which
+		// cannot be deleted.
+	}
```
tvoran, line 317: The comment is great, but should we log something here? Otherwise the `if err := ...` seems extraneous (and go-staticcheck is complaining to me about it 🙂).

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -336,10 +334,37 @@ func TestVaultDynamicSecret(t *testing.T) {
...
+	// Get a Vault client so we can validate that all leases have been removed.
+	cfg := api.DefaultConfig()
+	cfg.Address = vaultAddr
+	c, err := api.NewClient(cfg)
+	assert.NoError(t, err)
+	c.SetToken(vaultToken)
+	// Check to be sure all leases have been revoked.
+	retry.DoWithRetry(t, "waitForAllLeasesToBeRevoked", 30, time.Second, func() (string, error) {
+		// ensure that all leases have been revoked.
+		resp, err := c.Logical().ListWithContext(ctx, fmt.Sprintf("sys/leases/lookup/%s/creds/readonly", outputs.DBRole))
```
tvoran, line 355: Remind me how this works? I was thinking it would be more like `sys/leases/lookup/<outputs.DBPath>/creds/<output.DBRole>`.

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gruntwork-io/terratest/modules/retry"
+
 	"github.com/cenkalti/backoff/v4"
 	"github.com/gruntwork-io/terratest/modules/files"
 	"github.com/gruntwork-io/terratest/modules/k8s"
```
tvoran, line 21: import escapee:

```suggestion
	"github.com/cenkalti/backoff/v4"
	"github.com/gruntwork-io/terratest/modules/files"
	"github.com/gruntwork-io/terratest/modules/k8s"
	"github.com/gruntwork-io/terratest/modules/retry"
```

tvoran (approved):
Seems to be working for me!

tvoran:
Oh and don't forget a changelog entry 🙂

controllers/vaultdynamicsecret_controller.go
kschoche: Good call, we're actually emitting an Event and logging inside this function in those cases, I'm going to change the function signature a bit so we're not "hiding anything" and move the comment inside `revokeLease`

test/integration/vaultdynamicsecret_integration_test.go
kschoche: Ah, I grabbed this from https://developer.hashicorp.com/vault/docs/commands/lease#examples
I think you're right on that path, I'll look into it.
I have validated in the controller logs though that the leases are revoked, but this does look off thx!

kschoche: Yup that was it, thanks! Did confirm that I can see the # of leases getting smaller through that loop on shutdown!

PR #145, opened 2023-04-18
remove vds kubebuilder fields which are unnecessary

kschoche:
Removing unnecessary kubebuilder fields from the vds controller:
* All of the verbs are duplicated on line 52 except for `delete` which we aren't using.

### DIFF
--- a/controllers/vaultdynamicsecret_controller.go
+++ b/controllers/vaultdynamicsecret_controller.go
@@ -55,9 +55,6 @@ type VaultDynamicSecretReconciler struct {
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;patch
-//
-// needed for managing cached Clients, duplicated in vaultconnection_controller.go
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;delete;update;patch
 
 // Reconcile ensures that the VaultDynamicSecret Custom Resource is synced from Vault to its
 // configured Kubernetes secret. The resource will periodically be reconciled to renew the


controllers/vaultdynamicsecret_controller.go
```diff
@@ -55,9 +55,6 @@ type VaultDynamicSecretReconciler struct {
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;patch
-//
-// needed for managing cached Clients, duplicated in vaultconnection_controller.go
```
tvoran, line 59: Might be nice to keep the comment in here somewhere? And is the reason `delete` is not needed that we rely on k8s garbage collection to cleanup secrets?

tvoran (approved):

controllers/vaultdynamicsecret_controller.go
kschoche: I think this comment was leftover from the original cached client implementation, the reason that we don't need delete is we have an `OwnerReference` set on the Kube secret that is created so that we're not issuing any deletes, the Kube API server is doing that for us, yup!
I'm not entirely sure if the comment is relevant at this point, I could add a comment on the existing kubebuilder line which covers the verbs for secrets? I figured that bit was obvious though :-)
Happy to change things around, Thoughts?

tvoran: Hmm, so it looks like if you follow the `ClientFactory.Get()` call from vaultdynamicsecret_controller.go down far enough, it is deleting K8s secrets as part of the cache storage mechanism (since they're immutable). So now I'm thinking we should leave the delete capability in (but still dedup those two lines).

call:
https://github.com/hashicorp/vault-secrets-operator/blob/b3cd80d52300af0fe4865fb1ccacd6729d964774/controllers/vaultdynamicsecret_controller.go#L113

delete:
https://github.com/hashicorp/vault-secrets-operator/blob/b3cd80d52300af0fe4865fb1ccacd6729d964774/internal/vault/cache_storage.go#L259

I guess the other thing we need secrets permissions for is for creating the destination secret if requested, but there we do seem to rely on garbage collection for cleanup.

kschoche: Good call out, let me dig into this, I ran it locally and in CI and it passed, so either it failed to delete it silently due to rbac or we're getting the rbac policies overridden elsewhere.
In either case, agree that we can dedupe the policies in the end.

benashz: @tvoran  is correct, the Operator requires delete for Secrets.

kschoche:
Closing this since it doesn't actually affect the RBAC policy generated for the ClusterRole.

PR #152, opened 2023-04-21
Adds approle as an auth method

kschoche:
* Adds approle as an auth method


kschoche:
Way too many merge conflicts with the base branch, creating #155 to replace this.

PR #154, opened 2023-04-24
Adding testing framework for EKS 

gautambaghel:
Vault 13852/test against EKS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -389,8 +389,12 @@ ci-ecr-build: ## Build operator binary for ECR (without generating assets).
 
 .PHONY: port-forward
 port-forward:
-	@echo "Starting port forwarding..."
-	@bash -c 'trap exit SIGINT; while true; do kubectl port-forward -n $(K8S_VAULT_NAMESPACE) statefulset/vault 38300:8200; done'
+	@if lsof -Pi :38300 -sTCP:LISTEN -t >/dev/null ; then \
+	    echo "Port 38300 is already in use, please free it up and try again."; \
+	else \
+	    echo "Starting port forwarding..."; \
+	    bash -c 'trap exit SIGINT; while true; do kubectl port-forward -n $(K8S_VAULT_NAMESPACE) statefulset/vault 38300:8200; done'; \
+	fi
 
 # Currently only supports amd64
 .PHONY: ci-ecr-build-push
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,9 @@ VAULT_PATCH_ROOT = $(INTEGRATION_TEST_ROOT)/vault
 TF_INFRA_SRC_DIR ?= $(INTEGRATION_TEST_ROOT)/infra
 TF_VAULT_STATE_DIR ?= $(TF_INFRA_SRC_DIR)/state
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -368,6 +371,53 @@ helm-chart: manifests kustomize helmify
...
+integration-test-eks: ## Run unit tests in the EKS cluster
+	@$(eval KIND_CLUSTER_CONTEXT := $(shell $(TERRAFORM) -chdir=$(TF_EKS_DIR) output -raw cluster_arn))
+	@$(eval IMAGE_TAG_BASE := $(shell $(TERRAFORM) -chdir=$(TF_EKS_DIR) output -raw ecr_url))
+	$(MAKE) setup-vault KIND_CLUSTER_CONTEXT=$(KIND_CLUSTER_CONTEXT) IMAGE_TAG_BASE=$(IMAGE_TAG_BASE)
+	kubectl port-forward vault-0 -n $(K8S_VAULT_NAMESPACE) 38300:8200 &
+	cd $(CONFIG_MANAGER_DIR) && $(KUSTOMIZE) edit set image controller=$(IMG)
+	SUPPRESS_TF_OUTPUT=$(SUPPRESS_TF_OUTPUT) SKIP_CLEANUP=$(SKIP_CLEANUP) OPERATOR_NAMESPACE=$(OPERATOR_NAMESPACE) \
+	OPERATOR_IMAGE_REPO=$(IMAGE_TAG_BASE) OPERATOR_IMAGE_TAG=$(VERSION) \
+    INTEGRATION_TESTS=true KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) KIND_CLUSTER_CONTEXT=$(KIND_CLUSTER_CONTEXT) CGO_ENABLED=0 \
+	go test github.com/hashicorp/vault-secrets-operator/test/integration/... $(TESTARGS) -timeout=30m
```
tvoran, line 409: I wonder if we could just call `make integration-test` here with the cluster context. Something like this?

```suggestion
	$(MAKE) integration-test KIND_CLUSTER_CONTEXT=$(KIND_CLUSTER_CONTEXT)
```

Though I'm also thinking we may want to test Vault Enterprise on EKS instead of OSS.

tvoran (commented):
Really great to have this working!

Makefile
gautambaghel: @tvoran - Should it be either OSS or Ent? or should I look at doing both tests perhaps? I'll try out this recommendation, I believe I'll also need to pass the Operator Image repo and Image tag in that make command

tvoran: Doing both would be fine too, but I think we should at least be testing Ent.

gautambaghel: Add the enterprise tests + OSS tests in the CI jobs using env var and the same make command, working fine 👍🏾  I'll add it in the readme file

Fixed the port-forward, and made it universal so that when I work with GKE and AKS it should work there as well.

now integration-test-eks calls integration-test as it should! 👍🏾

gautambaghel:
This works!

tvoran (approved):
Thanks!

PR #155, opened 2023-04-26
Adds approle auth method support

kschoche:
* Updates the VaultAuth spec to add approle as a supported Auth Method. 
* Updates the CRDs / chart templates to document this new auth method, and updates bats tests.
* Adds an approle credential provider.
* Removes SecretKeyRef in the JWT auth method spec and replaces it with a SecretRef (string), introducing mandatory secret Keys.
* Small cleanup of the JWT auth method and some common functions used internally.

Note: There are a ton of commits from merging main, will squash when we merge.


config/manager/kustomization.yaml
```diff
@@ -16,4 +16,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: hashicorp/vault-secrets-operator
-  newTag: 0.1.0-beta
\ No newline at end of file
+  newTag: 0.0.0-dev
```
kschoche, line 19:
```suggestion
  newTag: 0.1.0-beta
```

test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -117,6 +117,54 @@ resource "vault_jwt_auth_backend_role" "dev" {
...
+  # role_id is auto-generated, and we use this to do the Login
+  token_policies = [vault_policy.approle.name]
+}
+
+# Creates the Secret ID for the AppRole
+resource "vault_approle_auth_backend_role_secret_id" "id" {
+  namespace  = local.namespace
+  backend    = vault_auth_backend.approle.path
+  role_name  = vault_approle_auth_backend_role.role.role_name
+  depends_on = [kubernetes_namespace.tenant-1, vault_approle_auth_backend_role.role]
```
kschoche, line 140:
```suggestion
```

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	secretsv1alpha1 "github.com/hashicorp/vault-secrets-operator/api/v1alpha1"
+)
+
+type ApproleCredentialProvider struct {
```
benashz, line 17: nit: I think the auth engined is called `AppRole`, may want to rename to `AppRole*`

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+
+	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	secretsv1alpha1 "github.com/hashicorp/vault-secrets-operator/api/v1alpha1"
+)
+
```
benashz, line 16: would be good to declare that `AppRoleCredentialProvider` implements `CredentialProvider` interface e.g:  https://github.com/hashicorp/vault-secrets-operator/blob/2081cc71f0cc9fee11a5e3a7e2b8bb1939ef8697/internal/vault/credentials/kubernetes.go#L17

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (l *ApproleCredentialProvider) GetUID() types.UID {
+	return l.uid
+}
+
+func (l *ApproleCredentialProvider) Init(ctx context.Context, client ctrlclient.Client, authObj *secretsv1alpha1.VaultAuth, providerNamespace string) error {
+	l.authObj = authObj
+	l.providerNamespace = providerNamespace
+	l.uid = types.UID(uuid.New().String())
```
benashz, line 34: I believe that this will was cause a log of churn in the Client, since the cache will be updated after each call to `Init()` e.g: Operator restart. This could lead to more frequent lease renewal failures. I wonder if we could rely on the Secret's UID and some other identifier here?

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	return creds, nil
+}
+
+func (l *ApproleCredentialProvider) getSecretID(ctx context.Context, client ctrlclient.Client) (string, error) {
+	logger := log.FromContext(ctx)
+
+	secret, err := getSecret(ctx, client, l.authObj.Namespace, l.authObj.Spec.AppRole.SecretKeyRef.Name)
+	if err != nil {
+		logger.Error(err, "Failed to get secret when fetching secret_id ", "role_id", l.authObj.Spec.AppRole.RoleID)
```
benashz, line 60:
```suggestion
		logger.Error(err, "Failed to get secret when fetching secret_id", "role_id", l.authObj.Spec.AppRole.RoleID)
```

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil
+}
+
+func (l *ApproleCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.Client) (map[string]interface{}, error) {
+	logger := log.FromContext(ctx)
+	// Fetch the secret_id each time we call GetCreds in case the secret_id has changed since
+	// the last time the client token was generated. In the case of AppRole this is assumed to be common.
+	secretID, err := l.getSecretID(ctx, client)
+	if err != nil || secretID == "" {
+		logger.Error(err, "Failed to get secret_id for ", "role_id", l.authObj.Spec.AppRole.RoleID)
```
benashz, line 44:
```suggestion
		logger.Error(err, "Failed to get secret_id, "role_id", l.authObj.Spec.AppRole.RoleID)
```

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (l *ApproleCredentialProvider) getSecretID(ctx context.Context, client ctrlclient.Client) (string, error) {
+	logger := log.FromContext(ctx)
+
+	secret, err := getSecret(ctx, client, l.authObj.Namespace, l.authObj.Spec.AppRole.SecretKeyRef.Name)
+	if err != nil {
+		logger.Error(err, "Failed to get secret when fetching secret_id ", "role_id", l.authObj.Spec.AppRole.RoleID)
+		return "", err
+	}
+	secretID := string(secret.Data[l.authObj.Spec.AppRole.SecretKeyRef.Key])
```
benashz, line 63: We may want to return an error if `Key` is not in `Data`

test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -117,6 +116,53 @@ resource "vault_jwt_auth_backend_role" "dev" {
...
+}
+
+resource "vault_policy" "approle" {
+  name      = "approle"
+  namespace = local.namespace
+  policy    = <<EOT
+path "${vault_mount.kvv2.path}/*" {
+  capabilities = ["read","list","update"]
+}
+path "auth/approle/login" {
```
benashz, line 159: nit: suggest replacing `auth/approle` with `vault_auth_backend.approle.path` (unfortunately that attribute is missing the leading `auth` component.

```suggestion
path "auth/${vault_auth_backend.approle.path}/login" {
```

api/v1alpha1/vaultauth_types.go
```diff
@@ -51,6 +51,15 @@ type VaultAuthConfigJWT struct {
 	TokenExpirationSeconds int64 `json:"tokenExpirationSeconds,omitempty"`
 }
 
+// VaultAuthConfigAppRole provides VaultAuth configuration options needed for authenticating to Vault.
+type VaultAuthConfigAppRole struct {
+	// RoleID (role_id) to use for authenticating to Vault.
+	RoleID string `json:"roleid"`
```
benashz, line 57: nit: I think the json flag should be mixed case: `roleId` , like `serviceAccount` above.

api/v1alpha1/vaultauth_types.go
```diff
@@ -51,6 +51,15 @@ type VaultAuthConfigJWT struct {
...
+	// Selects a key of a secret in the AuthMethod's namespace which holds the secret_id of the approle used
```
benashz, line 58: I wonder if we might be more opinionated here and  require a specific key be present in the secret. We could at a minimum set a `default` value.

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -158,13 +158,33 @@ func TestVaultAuthMethods(t *testing.T) {
...
+		{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      "vaultauth-test-approle",
+				Namespace: testK8sNamespace,
+			},
+			Spec: secretsv1alpha1.VaultAuthSpec{
+				// No VaultConnectionRef - using the default.
+				Namespace: testVaultNamespace,
+				Method:    "approle",
+				Mount:     "approle",
```
benashz, line 170: I wonder if these values should either be TF vars, or outputs. That would prevent the test from becoming out of sync with the TF.

benashz (commented):
Made a first pass. The one thing that might be problematic is the use of a UUID4 as the credential provider's UID.

api/v1alpha1/vaultauth_types.go
kschoche: Hrm, I suppose we could do this, I need to figure out how to use kubebuilder to define defaults for a custom object.
What do you envision that `key` would be?
I personally feel like a user is probably used to the concept of a secretRef's name+key convention enough to probably not forget the key when creating the reference?
Happy to make the change :)

internal/vault/credentials/approle.go
kschoche: Good catch, thanks!

internal/vault/credentials/approle.go
kschoche: Fixed these in a pending upcoming commit, thanks for the catch!

benashz (approved):
I added some additional comments/suggestions, but otherwise looks great!

api/v1alpha1/vaultauth_types.go
```diff
@@ -25,21 +25,13 @@ type VaultAuthConfigKubernetes struct {
...
 // VaultAuthConfigJWT provides VaultAuth configuration options needed for authenticating to Vault.
 type VaultAuthConfigJWT struct {
 	// Role to use for authenticating to Vault.
 	Role string `json:"role"`
-	// SecretKeyRef to use when referencing the secret containing the JWT token
-	// to authenticate to Vault's JWT authentication backend.
-	SecretKeyRef *SecretKeySelector `json:"secretKeyRef,omitempty"`
+	// SecretRef is the name of a Kubernetes secret in the consumer's (VDS/VSS/PKI) namespace which provides the JWT token
+	// to authenticate to Vault's JWT authentication backend. The secret must have a key named `jwt` which holds the JWT token.
+	SecretRef string `json:"secretRef,omitempty"`
```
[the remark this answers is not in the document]
kschoche: New laptop + new install of goland gave me a 120 character wide default .. fixing this thanks!

api/v1alpha1/vaultauth_types.go
```diff
@@ -51,6 +43,17 @@ type VaultAuthConfigJWT struct {
...
 }
 
+// VaultAuthConfigAppRole provides VaultAuth configuration options needed for authenticating to Vault via an
+// AppRole AuthMethod.
+type VaultAuthConfigAppRole struct {
+	// RoleID `role_id` of the AppRole Role to use for authenticating to Vault.
+	RoleID string `json:"roleId"`
+
+	// SecretRef is the name of a Kubernetes secret in the consumer's (VDS/VSS/PKI) namespace which provides the
+	// `secret_id` associated with the AppRole Role. The secret must have a key named `id` which holds the `secret_id`.
```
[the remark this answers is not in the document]
kschoche: Good call-out, I've changed the wording a bit, got rid of `secret_id` and kept `id`

api/v1alpha1/vaultauth_types.go
```diff
@@ -51,6 +44,18 @@ type VaultAuthConfigJWT struct {
...
 
+// VaultAuthConfigAppRole provides VaultAuth configuration options needed for authenticating to
+// Vault via an AppRole AuthMethod.
+type VaultAuthConfigAppRole struct {
+	// RoleID `role_id` of the AppRole Role to use for authenticating to Vault.
+	RoleID string `json:"roleId"`
+
+	// SecretRef is the name of a Kubernetes secret in the consumer's (VDS/VSS/PKI) namespace which
+	// provides the AppRole Role's SecretID. The secret must have a key named `id` which holds the
+	// ppRole Role's secretID.
```
tvoran, line 55:
```suggestion
	// AppRole Role's secretID.
```

test/integration/integration_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/vault-secrets-operator/internal/vault/credentials"
```
tvoran, line 22: This could be grouped with the other local imports below.

internal/vault/credentials/jwt.go
```diff
@@ -95,13 +97,27 @@ func (l *JWTCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.
...
+	key := ctrlclient.ObjectKey{
+		Namespace: l.providerNamespace,
+		Name:      l.authObj.Spec.JWT.SecretRef,
+	}
+	l.tokenSecret, err = getSecret(ctx, client, key)
 	if err != nil {
 		return nil, err
 	}
-
+	jwtData := l.tokenSecret.Data[ProviderSecretKeyJWT]
```
benashz, line 108: nit: couldn't you do the get + existence check e.g: `jwtData, ok := l.tokenSecret.Data[ProviderSecretKeyJWT]`

tvoran (approved):
Nice work!

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,82 @@
...
+	key := ctrlclient.ObjectKey{
+		Namespace: l.providerNamespace,
+		Name:      l.authObj.Spec.AppRole.SecretRef,
+	}
+	secret, err := getSecret(ctx, client, key)
+	if err != nil {
+		logger.Error(err, "Failed to get secret", "secret_name", l.authObj.Spec.AppRole.SecretRef)
+		return nil, err
+	}
+	secretID := secret.Data[ProviderSecretKeyAppRole]
```
benashz, line 66: nit: couldn't you do the get + existence check e.g: `secretID, ok := l.tokenSecret.Data[ProviderSecretKeyJWT]`

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,82 @@
...
+	if secretID == nil {
```
benashz, line 67: I think we can probably do a length check on the byte slice, then we don't need the extra string evaluation below.

internal/vault/credentials/jwt.go
```diff
@@ -95,13 +97,27 @@ func (l *JWTCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.
...
+	if jwtData == nil {
+		logger.Error(err, "Failed to get jwt token from secret, no key found", "secret_name",
```
benashz, line 109: looks like we are shadowing `err` here, which would be nil.

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,81 @@
...
+		Namespace: l.providerNamespace,
+		Name:      l.authObj.Spec.AppRole.SecretRef,
+	}
+	secret, err := getSecret(ctx, client, key)
+	if err != nil {
+		logger.Error(err, "Failed to get secret", "secret_name", l.authObj.Spec.AppRole.SecretRef)
+		return nil, err
+	}
+	if secretID, ok := secret.Data[ProviderSecretKeyAppRole]; !ok {
+		logger.Error(err, "Failed to get secretID from secret, no key found", "secret_name",
```
benashz, line 67: looks like we are shadowing `err` here, which would be nil.

internal/vault/credentials/jwt.go
kschoche: good catch thanks! fixed both of these issues in b335b37

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+	if secretID, ok := secret.Data[ProviderSecretKeyAppRole]; !ok {
+		err = fmt.Errorf("no key %q found in secret", ProviderSecretKeyAppRole)
+		logger.Error(err, "Failed to get secretID from secret", "secret_name",
+			l.authObj.Spec.AppRole.SecretRef)
+		return nil, err
+	} else if len(secretID) == 0 {
+		err = fmt.Errorf("no data found in secret key %q", ProviderSecretKeyAppRole)
+		logger.Error(err, "Failed to get secretID from secret", "secret_name",
```
benashz, line 73: nit: this message is a bit unclear to me, what is the `secretID`?

benashz (approved):
Looks good, only a few nits related to the log messages.

internal/vault/credentials/approle.go
```diff
@@ -0,0 +1,83 @@
...
+	// Fetch the AppRole Role's SecretID from the Kubernetes Secret each time there is a call to
+	// GetCreds in case the SecretID has changed since the last time the client token was
+	// generated. In the case of AppRole this is assumed to be common.
+	key := ctrlclient.ObjectKey{
+		Namespace: l.providerNamespace,
+		Name:      l.authObj.Spec.AppRole.SecretRef,
+	}
+	secret, err := getSecret(ctx, client, key)
+	if err != nil {
+		logger.Error(err, "Failed to get secret", "secret_name", l.authObj.Spec.AppRole.SecretRef)
```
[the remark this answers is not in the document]
kschoche: Good call, I'll add this in a follow-on PR. thanks!

rjchicago:
First, I'm very excited this feature has been developed! 👏 
Now, I'm working on testing this out - is `0.1.0-beta.1` the best version to test? If so, I've run into a few discrepancies between docs and implementation. The [Helm chart default values](https://artifacthub.io/packages/helm/hashicorp/vault-secrets-operator/0.1.0-beta.1?modal=values&path=defaultAuthMethod.appRole) for this version has the following:

``` yaml
  appRole:
    roleId: ""
    secretName: ""
```

However, it appears these should be `roleid` (case) and `secretRef` per the [template](https://artifacthub.io/packages/helm/hashicorp/vault-secrets-operator/0.1.0-beta.1?modal=template&template=default-vault-auth-method.yaml). Also `appRole.method` documentation should be updated to include `appRole` method now.

kschoche:
Hi @rjchicago!
First, thanks for filing this issue and bringing this to our attention, and apologies that you're running into these issues here. I'll fix this right now, and it will be fixed in our GA release (very soon).

We just released `0.1.0-rc.1` a couple days ago so that would be the latest that you can use, but it will have the same issue you mentioned.

The intended fields are `roleId`, `secretRef` and `appRole` but I'll put notes in the changelogs to indicate anything that breaks.
Cheers,
~Kyle

rjchicago:
@kschoche, thanks for jumping on this so quickly! I wasn't expecting such a quick reply on a Friday afternoon! 😉

kschoche:
> @kschoche, thanks for jumping on this so quickly! I wasn't expecting such a quick reply on a Friday afternoon! 😉

You're welcome, happy to help:D Thanks for testing this out and raising the issue! 🥂

PR #156, opened 2023-04-26
Add integration-test-helm-ent and make it the default for matrix

kschoche:
We currently restrict the scope of our testing matrix to the latest release of Vault OSS (1.13.1) for the majority of our test environments, with the exception of the `ENT tests using Kustomize` which run against the full test matrix of Vault versions and Kubernetes versions.
This PR changes this so that we deploy using Helm instead of Kustomize for our most expansive test and only run the existing `ENT tests using Kustomize` against `1.13.1`.

* Adds a new Makefile target `integration-test-helm-ent` which runs the integration test with Helm and Vault ENT
* Use this test as the basis for our more expansive tests by introducing and running `ENT tests using Helm` instead of `ENT tests using Kustomize` across all versions of Kubernetes and Vault.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -237,6 +237,8 @@ jobs:
         run: |
           make integration-test-ent SUPPRESS_TF_OUTPUT=true EXPORT_KIND_LOGS_ROOT=${{ steps.create_kind_export_log_root.outputs.log_root }}
       - name: ENT tests using Helm
+        # TODO: When the VDS integration test supports Helm swap the matrix filter so this test runs the full matrix.
+        if: ${{ matrix.vault-version == '1.13.1' }}
         env:
           INTEGRATION_TESTS: true
           VAULT_LICENSE_CI: ${{ secrets.VAULT_LICENSE_CI }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -228,8 +228,8 @@ jobs:
         run: |
           make integration-test SUPPRESS_TF_OUTPUT=true EXPORT_KIND_LOGS_ROOT=${{ steps.create_kind_export_log_root.outputs.log_root }}
       - name: ENT tests using Kustomize
-        # Ideally we only need to test the latest release of OSS Vault.
-        if: ${{ matrix.vault-version == '1.13.1' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -228,12 +228,21 @@ jobs:
         run: |
           make integration-test SUPPRESS_TF_OUTPUT=true EXPORT_KIND_LOGS_ROOT=${{ steps.create_kind_export_log_root.outputs.log_root }}
       - name: ENT tests using Kustomize
+        # Ideally we only need to test the latest release of OSS Vault.
```
tvoran, line 231:
```suggestion
        # Ideally we only need to test the latest release of ENT Vault with Kustomize.
```

.github/workflows/build.yaml
```diff
@@ -228,12 +228,21 @@ jobs:
...
+        if: ${{ matrix.vault-version == '1.13.1' }}
```
tvoran, line 232: It doesn't look like the VaultDynamicSecret tests run with helm? Maybe we should hold off on this condition until they do?

tvoran (approved):

.github/workflows/build.yaml
kschoche: Yeah, I was on the fence about that too.
The reason I went for this approach is that I wanted to avoid having 2 sets of "expansive" tests which don't apply the matrix filter just to keep the runtime down.

Do we expect that VDS test would be somehow more prone to potential issues from combinations of Vault version  and k8s version than the other tests (which already support helm)?

I suppose there isn't a big reason to introduce this change here unless it starts to make tests run too long, thoughts? I'm fine reverting it.

tvoran: Yeah, I dunno, I'm just worried we're losing coverage. Maybe shift the condition to the ENT Helm tests for now, with a todo to switch once the dynamic tests work with helm?

kschoche: Okay took out the conditional so we're running the full matrix against Helm and Kustomize for Vault ENT, and left a TODO comment to put it in in the future.
I think they run in parallel anyways, so shouldnt increase the time right now, I'll keep an eye out though!

PR #157, opened 2023-04-26
Add support for cloning a namespaced Vault Client

### LINKED ISSUE #117: namespace field in SecretSpecs
Thank you for creating this operator, I really enjoy using it so far.

I have  short question regarding the status of the [`namespace` field in the "SecretSpecs"](https://developer.hashicorp.com/vault/docs/platform/k8s/vso/api-reference).

I currently store the Vault secrets in the Vault root namespace. According to the documentation, the field configures "Namespace to get the secret from in Vault". However, in my experiments with the operator I could set this field to any value (like "abc", where there exists neither a Kubernetes nor a Vault namespace with that value).

First I was wondering, if the `namespace` field was used to configure the namespace of the `vaultAuthRef`. But it does not seem the case here.

Is the `namespace` field not implemented yet?

benashz:
Client's can now be cloned for use in namespaces other than the one configured in the VaultAuth. A new sub-cache has been added to the clientCache. The sub-cache is responsible for caching cloned Clients. A cached clone is pruned from the cache whenever its parent is evicted or removed from the main cache. The spec of the ClientCacheKey has been extended to support a new "clone" suffix which differentiates it from a "parent" client. Client clones share the parent's token.

Closes #117


internal/vault/client.go
```diff
@@ -163,6 +169,50 @@ type defaultClient struct {
...
+	}
+
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	clone, err := c.client.Clone()
+	if err != nil {
+		return nil, err
+	}
+	clone.SetToken(c.client.Token())
```
tvoran, line 199: Does setting `config.CloneToken = true` in MakeVaultClient() not actually clone the token?

internal/vault/client_factory.go
```diff
@@ -381,6 +425,10 @@ func (m *cachingClientFactory) cacheClient(c Client) (ClientCacheKey, error) {
 		return "", err
 	}
 
+	if c.IsClone() {
+		return "", fmt.Errorf("cannot cache cloned Clients")
+	}
```
tvoran, line 430: Why are clone clients rejected here, when it looks like they're added to the cache in `cachingClientFactory.Get()`?

benashz: Updated in 1a28323f25fc18e3791844589c865f76376b9dde

internal/vault/client.go
tvoran: Hmm, I don't see 1a28323f25fc18e3791844589c865f76376b9dde here yet.

benashz: Oops, should be there now. Forgot to push it :(

internal/vault/client_factory.go
```diff
@@ -249,21 +251,58 @@ func (m *cachingClientFactory) Get(ctx context.Context, client ctrlclient.Client
...
+			if clone, ok := m.cache.Get(cacheKeyClone); ok {
+				return clone, nil
+			}
+
+			clone, err := c.Clone(ns)
+			if err != nil {
+				return nil, err
+			}
+
+			logger.Info("Cloned Client",
```
kschoche, line 277: Does it make sense to track metrics here of the # of namespaced clients, or is it covered elsewhere?

kschoche (approved):
Looks great, I left just a couple questions but I think the approach looks good.

tvoran (approved):

internal/vault/cache_key.go
```diff
@@ -21,6 +23,9 @@ var (
 	errorKeyLengthExceeded = errors.New("cache-key length exceeded")
 	errorDuplicateUID      = errors.New("duplicate UID")
 	errorInvalidUIDLength  = errors.New("invalid UID length")
+	cacheKeyRe             = regexp.MustCompile(
```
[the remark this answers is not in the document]
benashz: regex

internal/vault/client_factory.go
benashz: Do you mean including namespaces in the set of supported labels? Seems like something worth considering. I wounder if there might be any potential security issues around exposing the namespace.

config/manager/kustomization.yaml
```diff
@@ -16,4 +16,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: hashicorp/vault-secrets-operator
-  newTag: 0.1.0-beta
+  newTag: 0.0.0-dev
```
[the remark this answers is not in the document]
benashz: Yeah, I think we need to see why we keep getting these set. I think it happens when we run kustomize.

internal/vault/cache.go
```diff
@@ -103,21 +133,44 @@ func (c *clientCache) Prune(filterFunc ClientCachePruneFilterFunc) []ClientCache
...
 func NewClientCache(size int, callbackFunc onEvictCallbackFunc, metricsRegistry prometheus.Registerer) (ClientCache, error) {
-	lruCache, err := lru.NewWithEvict(size, callbackFunc)
-	if err != nil {
-		return nil, err
-	}
-
 	cache := &clientCache{
-		cache: lruCache,
+		// cache:      lruCache,
+		// cloneCache: lruCloneCache,
```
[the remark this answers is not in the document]
benashz: Good catch. Removed in 9fe1377055e8c2ea58bb1a29bd0b4eb794e65d74

config/manager/kustomization.yaml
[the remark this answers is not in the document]
benashz: Going to bump the version in `Makefile`, since that seems to be the source of the recurring reversion.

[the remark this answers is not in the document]
tvoran: I think it's probably because of the kustomize set image call from `integration-test`, so running the tests locally keeps modifying it:

https://github.com/hashicorp/vault-secrets-operator/blob/9fe1377055e8c2ea58bb1a29bd0b4eb794e65d74/Makefile#L250-L255

We may want to move that `set image` command into the build.yaml workflow where the integration tests are run?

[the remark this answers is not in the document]
tvoran: > We may want to move that set image command into the build.yaml workflow where the integration tests are run?

Eh on second thought, that's probably not what we want either. It would use the wrong image in the tests.

PR #160, opened 2023-04-27
VPS: make commonName optional

### LINKED ISSUE #150: PKI secret should not Require commonName
**Describe the bug**
Use of the Common Name field has been [deprecated since 2000](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1), and the requirement for a CN can be disabled in the Vault PKI backend role. Still, the Operator sets the field as Required. The Operator should only Require fields which are necessary to produce a valid Secret (in Vault and Kubernetes).

**To Reproduce**
1. Attempt to create a VaultPKISecret with no commonName value

Error: The VaultPKISecret "\<name>" is invalid: spec.commonName: Required value

**Expected behavior**
A certificate should be generated successfully without a Common Name.

**Environment**
* Kubernetes version: v1.25.9
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): kubeadm
  * Other configuration options or runtime services (istio, etc.):
* vault-secrets-operator version: v0.1.0-beta

benashz:
The Vault PKI role should have require_cn  to set false if commonName is not provided.

Closes #150

### DIFF
--- a/api/v1alpha1/vaultpkisecret_types.go
+++ b/api/v1alpha1/vaultpkisecret_types.go
@@ -60,7 +60,7 @@ type VaultPKISecretSpec struct {
 	Destination Destination `json:"destination"`
 
 	// CommonName to include in the request.
-	CommonName string `json:"commonName"`
+	CommonName string `json:"commonName,omitempty"`
 
 	// AltNames to include in the request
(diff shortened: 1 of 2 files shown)


kschoche (approved):
Looks good!

PR #164, opened 2023-05-01
Set the correct version in the Makefile

benashz:
Move the images block up to prevent 'kustomize edit' from stripping the end of file newline which resulted in a perpetual diff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,15 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.1.0-beta
+VERSION ?= 0.0.0-dev
 
 GO_VERSION = $(shell cat .go-version)
 
-CONFIG_MANAGER_DIR ?= config/manager
-KUSTOMIZE_BUILD_DIR ?= config/default
+CONFIG_SRC_DIR ?= config
+CONFIG_BUILD_DIR ?= build/config
+CONFIG_MANAGER_DIR ?= $(CONFIG_BUILD_DIR)/manager
+KUSTOMIZATION ?= default
+KUSTOMIZE_BUILD_DIR ?= $(CONFIG_BUILD_DIR)/$(KUSTOMIZATION)
 
 VAULT_IMAGE_TAG ?= latest
 VAULT_IMAGE_REPO ?=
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.0-dev
+VERSION ?= 0.1.0-beta
```
tvoran, line 6: FWIW this was intentionally set to 0.0.0-dev to match our other projects (vault-k8s, csi-provider) and to also have one less place to bump the version at release time, but I get that kustomize is making this awkward.

When developing locally, this might make it harder to tell if the released image or the built image is in use? If we can't keep it at 0.0.0-dev, maybe something else besides the released version? Whatever we choose it'll just have to match the kustomize version I guess.

tvoran (approved):

Makefile
benashz: Yeah, I fine with either approach. I think this may have been inadvertently introduced when we did the initial release. There be a way to have `kustomize edit` not munge the file in question.

kschoche: Also fine with either approach, maybe lets stick with precedent from the other projects?

benashz: I think we should probably set the dev version in the `kustomization.yaml`, then we can avoid the perpetual diff. We can then document the steps needed to install a production version using kustomize.

benashz: > Also fine with either approach, maybe lets stick with precedent from the other projects?

@kschoche which other projects?

kschoche: Vault-k8s and csi-provider I believe is what Theron had mentioned.

benashz: Ah, I missed that.

It looks like vault-k8s does not test the kustomize, so it probably doesn't suffer from the same issue. It looks like the vault-k8s kustomize is tied to a release. One thing we could do is to copy the kustomize root to `build/.` before applying it (for the dev case). Then we wouldn't have to an edit in place.

Makefile
```diff
@@ -246,9 +249,17 @@ ci-docker-build: ## Build docker image with the operator (without generating ass
 ci-test: vet envtest ## Run tests in CI (without generating assets)
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... $(TESTARGS) -coverprofile cover.out
 
-.PHONY: integration-test
-integration-test:  setup-vault ## Run integration tests for Vault OSS
+.PHONY: copy-config
+copy-config: ## Copy kustomize config to CONFIG_BUILD_DIR
+	@rm -rf $(CONFIG_BUILD_DIR)
+	@cp -a $(CONFIG_SRC_DIR) $(CONFIG_BUILD_DIR)
```
tvoran, line 255: May need a `mkdir -p $(CONFIG_BUILD_DIR)` in here somewhere?

tvoran (approved):
I like the approach!

Makefile
benashz: Good catch, fixed in https://github.com/hashicorp/vault-secrets-operator/pull/164/commits/c2c295d1d279c9c86301d09857d60c602fe1ce40

tvoran (approved):

PR #170, opened 2023-05-03
dynamic secrets: renew earlier and 

### LINKED ISSUE #99: Avoid waiting for last lease to regenerate dynamic credentials

tvoran:
Adds a new `RenewalPercent` field to VaultDynamicSecret for specifying at what percentage of a lease the renewal should happen, defaults to 67 with an allowed range of 0-90.

Adds a new horizon function that computes the horizon as the specified percentage of the lease, plus jitter of up to 10% of the lease, so that the horizon falls after the specified percentage of the lease. (This is inspired by the [lifetimewatcher behavior](https://github.com/hashicorp/vault/blob/f4689b1d85265dcb092e45bbde465d85b2cf1ec4/api/lifetime_watcher.go#L380-L393) from the vault api.)

Also returns an error from lease renewal if the requested ttl was capped, so that new credentials can be generated. Otherwise the renewal ttl will be [whatever is left on the max_ttl](https://github.com/hashicorp/vault/blob/f4689b1d85265dcb092e45bbde465d85b2cf1ec4/sdk/framework/lease.go#L101-L108), which can become very short before the renewal fails and and new credentials are generated.

For example, with a database engine that has `default_ttl = 15s` and `max_ttl = 200s`, the renewal horizons can end up in the millisecond range at the end:

```
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/IA6KjWdJ9o1B8x83JigI1hIJ, horizon=10.575193236s
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/IA6KjWdJ9o1B8x83JigI1hIJ, horizon=4.854674841s
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/IA6KjWdJ9o1B8x83JigI1hIJ, horizon=1.413186895s
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/IA6KjWdJ9o1B8x83JigI1hIJ, horizon=752.020265ms
DEBUG	events	Could not renew lease, lease_id=demo-db/creds/dev-postgres/IA6KjWdJ9o1B8x83JigI1hIJ, err=Error making API request.
* failed to renew entry: resp: (*logical.Response)(nil) err: past the max TTL, cannot renew
DEBUG	events	Rollout restart triggered for {Deployment vso-db-demo}
DEBUG	events	Secret synced, lease_id=demo-db/creds/dev-postgres/AkIihHlScSbsmFwzlsPpbW6d, horizon=12.736277687s
```

Also changes the ordering of "secret synced" and rollout events so that "secret synced" comes before rollout. 

With these changes, the renewal events look something like this:

```
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/z5N1kfESQLaZV41BGs41zdPB, horizon=11.22873807s
DEBUG	events	Renewed lease, lease_id=demo-db/creds/dev-postgres/z5N1kfESQLaZV41BGs41zdPB, horizon=10.426730023s
DEBUG	events	Lease renewal duration was truncated from 15s to 6s, requesting new credentials
DEBUG	events	Secret synced, lease_id=demo-db/creds/dev-postgres/qXH76AK21dI2RrUKGKZdvs0H, horizon=10.206722711s
DEBUG	events	Rollout restart triggered for {Deployment vso-db-demo}
```

Fixes #99


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Features:
 * VaultDynamicSecrets: CRD is extended with `Revoke` field which will result in the dynamic secret lease being revoked on rotation and CR deletion. Note: The VaultAuthMethod referenced by the VDS Secret must have a policy which provides `["update"]` on `sys/leases/revoke`. [GH-143](https://github.com/hashicorp/vault-secrets-operator/pull/143)
 * VaultAuth: Adds support for the JWT authentication method which either uses the JWT token from the provided secret reference, or a service account JWT token that VSO will generate using the provided service account. [GH-131](https://github.com/hashicorp/vault-secrets-operator/pull/131)
 
+Improvements:
+* VaultDynamicSecrets: Start lease renewal after 2/3 of lease duration [GH-170](https://github.com/hashicorp/vault-secrets-operator/pull/170)
```
tvoran, line 8: This change to renew at 2/3 may not be strictly necessary, especially with the fix for capped renewal below, but I left it in since it's similar to Vault Agent's behavior. Happy to remove it if y'all like.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -272,6 +268,14 @@ func (r *VaultDynamicSecretReconciler) renewLease(ctx context.Context, c vault.C
 	if err != nil {
 		return nil, err
 	}
+	// The renewal duration can come back as less than the requested increment
+	// if the time remaining on max_ttl is less than the increment. If there are
+	// rollout restart targets and the renewal was capped in this manner, return
+	// an error to ensure the targets have valid credentials while going through
+	// the rollout restart.
+	if len(o.Spec.RolloutRestartTargets) > 0 && resp.LeaseDuration < o.Status.SecretLease.LeaseDuration {
+		return nil, fmt.Errorf("lease renewal was capped to %ds, less than the requested %ds", resp.LeaseDuration, o.Status.SecretLease.LeaseDuration)
```
tomhjp, line 277: I wonder whether we should do this for all dynamic secrets, not just ones that have rollout restart targets defined? For example if db creds are read from disk (mounted via k8s secret), that pod would equally want a decent grace period where it can load some new credentials before the old ones expire.

Also one uber nit: This will cause a warning event to be emitted, but normal seems more fitting.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -193,6 +189,11 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}
```
tomhjp, line 189: Do we maybe need to delay revoking the old lease so that consumers have a grace period to load in and start using the newly created lease?

tomhjp (commented):
Looking good! Just a couple of questions on high level behaviour.

controllers/vaultdynamicsecret_controller.go
tomhjp: I'm going back and forth a little on this one. Maybe the answer is just that people who want overlap shouldn't use the `Revoke` option, as to get to here, the lease has to be expiring relatively soon anyway.

dnlopes: Sorry to pop up (I was the one who reported this issue):

> I'm going back and forth a little on this one. Maybe the answer is just that people who want overlap shouldn't use the Revoke option

Maybe I'm missing something (or I'm simply lacking knowledge of the Vault internals), but isn't the overlap given for free by requesting new credentials at 2/3 of the TTL? The revoke of the old credentials will still happen at 3/3 of the TTL, what this change will do is simply ensure that a new credential is generated and available for apps at 2/3 of the TTL. I wouldn't imagine the revoke flow to be changed here. 🤔 The grace period would be 1/3 of the TTL.

benashz: > I'm going back and forth a little on this one. Maybe the answer is just that people who want overlap shouldn't use the `Revoke` option, as to get to here, the lease has to be expiring relatively soon anyway. Although `Revoke` does also control behaviour on deleting the VDS too, so maybe that's not a perfect answer.

We could perhaps support both. I do like the idea of supporting deferred lease revocation. Definitely something worth considering for the future.

tomhjp: I may be misreading the flow of the code as I haven't tested it locally, but it looks to me like every time we hit the line `secretLease, err := r.syncSecret(ctx, vClient, o)` (which generates a new lease), we always then revoke the old lease 2 lines above this comment if `o.spec.Revoke` is set. So from that reading, I would expect the `r.revokeLease(ctx, o, oldLease.ID)` line to be called at 2/3 of the TTL as the code currently stands.

Take it with a pinch of salt though - I would want to test it to be certain.

benashz (commented):
Great work. I really like this change. I added a few comments/suggestions fwiw.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -355,3 +359,12 @@ func (r *VaultDynamicSecretReconciler) revokeLease(ctx context.Context, o *secre
...
 }
+
+// inRenewalWindow checks if 2/3 of the lease duration of a VDS has elapsed
+func inRenewalWindow(status secretsv1alpha1.VaultDynamicSecretStatus) bool {
+	leaseDuration := time.Duration(status.SecretLease.LeaseDuration) * time.Second
+	startRenewingAt := time.Duration(float64(leaseDuration.Nanoseconds()) * 2 / 3)
+	ts := time.Unix(status.LastRenewalTime, 0).Add(startRenewingAt).Unix()
+	now := time.Now().Unix()
+	return ts-now <= 0
+}
```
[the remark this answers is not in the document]
tvoran: I think that's certainly clearer, thanks.

CHANGELOG.md
tvoran: Yeah, like a `RenewalPercent` field that's out of 100? Probably restricted to 0-90 to leave room for up to 10% jitter. Trying that in 523b636.

controllers/vaultdynamicsecret_controller.go
tvoran: Yep, all good points. In 523b636 it now handles a truncated lease renewal for any dynamic secret, and emits a more friendly event. Not sure if we want to requeue though, since it's just falling through to refresh the credentials.

controllers/common.go
```diff
@@ -34,6 +34,21 @@ func computeHorizonWithJitter(minDuration time.Duration) time.Duration {
 	return minDuration - (time.Duration(jitterMax) + time.Duration(uint64(random.Int63())%u))
 }
 
+// computeDynamicHorizonWithJitter returns a time.Duration that is 2/3 of the
+// lease duration, plus additional random jitter (up to 10% of the lease), to
+// ensure the horizon falls within the 2/3 renewal window
+func computeDynamicHorizonWithJitter(leaseDuration time.Duration) time.Duration {
```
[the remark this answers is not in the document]
tvoran: Yep, good point, I rearranged things a bit in eacc2c5.

controllers/vaultdynamicsecret_controller.go
tomhjp: > However, I currently cannot think of a scenario of a dynamic secret that we want to rotate after 2/3 of the TTL (to avoid the issue this PR is addressing,) and at the same time, be willing to revoke the old credential at 2/3 of the TTL, as it would defeat the purpose of renewing earlier.

I agree, having overlap feels like a sensible default. So perhaps `Revoke` should only apply when deleting the VDS. We haven't done a release with `Revoke` yet, so it wouldn't be a breaking change.

> We could perhaps support both. I do like the idea of supporting deferred lease revocation. Definitely something worth considering for the future.

Yeah, deferred lease revocation strikes me as the more advanced use-case so we can save that part for a later date IMO.

benashz: > I agree, having overlap feels like a sensible default. So perhaps `Revoke` should only apply when deleting the VDS. We haven't done a release with `Revoke` yet, so it wouldn't be a breaking change.

I agree, we should only revoke the lease on delete.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -155,14 +151,17 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-			horizon := computeHorizonWithJitter(leaseDuration)
+			horizon := computeDynamicHorizonWithJitter(leaseDuration, o.Spec.RenewalPercent)
 			r.Recorder.Eventf(o, corev1.EventTypeNormal, consts.ReasonSecretLeaseRenewal,
 				"Renewed lease, lease_id=%s, horizon=%s", leaseID, horizon)
 			return ctrl.Result{RequeueAfter: horizon}, nil
 		} else {
 			// The secretLease was not renewed or failed, continue through Reconcile and do a rollout restart.
 			doRolloutRestart = true
-			if !isLeaseNotfoundError(err) {
+			if e, ok := err.(*LeaseTruncatedError); ok {
```
benashz, line 161: We may want to handle the case were err is wrapped:

```suggestion
			if e, ok := err.(*LeaseTruncatedError); ok || e != nil && errors.As(err, &e)
```

controllers/common.go
```diff
@@ -17,21 +17,51 @@ import (
...
+type LeaseTruncatedError struct {
+	Expected int
+	Actual   int
+}
+
+func (l *LeaseTruncatedError) Error() string {
+	return fmt.Sprintf("lease renewal duration was truncated from %ds to %ds",
+		l.Expected, l.Actual)
+}
+
```
benashz, line 31: Suggest declaring that `LeaseTruncatedError` must implement `error`

```suggestion
var _ error = (*LeaseTruncatedError)(nil)
```

controllers/vaultdynamicsecret_controller.go
benashz: I think we can probably do this in a subsequent PR.

benashz (commented):
I really like how this is coming along!

controllers/common.go
```diff
@@ -17,21 +17,51 @@ import (
...
+	return minDuration - (time.Duration(max) + time.Duration(jitter))
+}
+
+// computeDynamicHorizonWithJitter returns a time.Duration that is the specified
+// percentage of the lease duration, plus additional random jitter (up to 10% of
+// the lease), to ensure the horizon falls within the specified renewal window
+func computeDynamicHorizonWithJitter(leaseDuration time.Duration, renewalPercent int) time.Duration {
+	max, jitter := computeMaxJitter(leaseDuration)
+
+	startRenewingAt := time.Duration(float64(leaseDuration.Nanoseconds()) * float64(renewalPercent) / 100)
```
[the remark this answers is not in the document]
tvoran: Sure, that wouldn't hurt. `renewalPercent` is capped at 90 in the CRD spec (to ensure the up-to-10% jitter doesn't overrun the lease) but can double check in this function too.

Example from terraform apply:
```
│ Error: spec.renewalPercent
│ 
│   with kubernetes_manifest.vault-dynamic-secret,
│   on app.tf line 57, in resource "kubernetes_manifest" "vault-dynamic-secret":
│   57: resource "kubernetes_manifest" "vault-dynamic-secret" {
│ 
│ Invalid value: 1000: spec.renewalPercent in body should be less than or equal to 90
```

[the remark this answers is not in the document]
benashz: Either way, happy that we are enforcing it on the schema.

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"testing"
+	"time"
+
+	secretsv1alpha1 "github.com/hashicorp/vault-secrets-operator/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_inRenewalWindow(t *testing.T) {
```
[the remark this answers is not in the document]
tvoran: Sure added that case and some testing around computeDynamicHorizonWithJitter() as well in 73610bc and 4d33de5.

controllers/common.go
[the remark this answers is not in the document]
tvoran: Added some internal capping in 73610bc

controllers/vaultdynamicsecret_controller.go
tvoran: Added in 73610bc.

controllers/common_test.go
```diff
@@ -0,0 +1,59 @@
...
+			expectedMin:    time.Duration(13.5 * float64(time.Second)),
+			expectedMax:    time.Duration(15 * time.Second),
+		},
+		"renewalPercent negative": {
+			leaseDuration:  time.Duration(15 * time.Second),
+			renewalPercent: -60,
+			expectedMin:    time.Duration(0),
+			expectedMax:    time.Duration(1.5 * float64(time.Second)),
+		},
+		"renewalPercent over 90": {
```
benashz, line 44: nit: should this test be named `renewalPercent over 100`

PR #171, opened 2023-05-05
Extend vault Client validation checks to handle failed renewals

benashz:
Previously, the Vault ClientFactory only checked to see if the Client's token's expiry to determine whether or not to replace the Client. This PR adds the following validation checks:
- Client's lastLifetimeWatcher error is set
- Client's lifetimeWatcher unset

Relates to #207


internal/vault/client.go
```diff
@@ -173,6 +174,37 @@ type defaultClient struct {
...
+			return errors.New("lifetime watcher not set")
+		}
+	}
+
+	if expired, err := c.checkExpiry(0); expired || err != nil {
+		var errs error
+		if expired {
+			errs = errors.Join(errs, errors.New("client token expired"))
+		}
+		return errors.Join(errs, err)
```
kschoche, line 202: This feels a little clunky to read, unless I'm misreading `checkExpiry()` it will never return `expired=true && err != nil`
```suggestion
		if expired {
			return errors.New("client token expired"))
		}
		return err
```

kschoche (approved):
Looks great!

internal/vault/client.go
benashz: Fair point, but I kind of like the idea of getting the additional error that may have been encountered when checking the expiration. It could provide a bit more context.

PR #172, opened 2023-05-05
VDS: allow for specifying an arbitrary secrets path

benashz:
This PR ensure that VDS can support any Vault secret engine path.

- drops hard coded `creds` path component.

Example of an updated VDS CR:
```yaml
kind: VaultDynamicSecret
metadata:
  name: vso-db-demo-create
  namespace: demo-ns
spec:
  destination:
    create: true
    name: vso-db-demo-created
  mount: demo-db
  path: creds/dev-postgres
  rolloutRestartTargets:
  - kind: Deployment
    name: vso-db-demo
```

**Breaking change:**
- Spec.Role is now Spec.Path

Close #108 #107

### DIFF
--- a/demo/infra/app/variables.tf
+++ b/demo/infra/app/variables.tf
@@ -30,5 +30,9 @@ variable "k8s_db_secret_count" {
   default = 50
 }
 
+variable "db_role" {
+  default = "dev-postgres"
+}
+
 variable "vault_address" {}
 variable "vault_token" {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,14 @@ Bugs:
 * Helm Chart: fix deployment templating so setting `controller.kubernetesClusterDomain` works as defined in values.yaml. [GH-183](https://github.com/hashicorp/vault-secrets-operator/pull/183)
 
 Features:
-* VaultDynamicSecrets: CRD is extended with `Revoke` field which will result in the dynamic secret lease being revoked on rotation and CR deletion. Note: The VaultAuthMethod referenced by the VDS Secret must have a policy which provides `["update"]` on `sys/leases/revoke`. [GH-143](https://github.com/hashicorp/vault-secrets-operator/pull/143)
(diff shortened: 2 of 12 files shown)


demo/infra/app/variables.tf
```diff
@@ -30,5 +30,9 @@ variable "k8s_db_secret_count" {
   default = 50
 }
 
+variable "db_role" {
```
kschoche, line 33: Not 100% on this, but I think there is a `local.db_role` which is now unused due to this change that can be removed?

kschoche (approved):
This looks great, I left a couple smallish comments that I'd like to circle back on before merging if you get a chance.
Also a reminder to add a changelog entry (especially since this is a breaking change). I think it would be helpful to disclose the spec change and how to migrate in that entry.

demo/infra/app/variables.tf
benashz: Oops, need to drop `local.db_role`

benashz:
> This looks great, I left a couple smallish comments that I'd like to circle back on before merging if you get a chance. Also a reminder to add a changelog entry (especially since this is a breaking change). I think it would be helpful to disclose the spec change and how to migrate in that entry.

@kschoche  Good call on the changelog updates. For the breaking changes + migration instructions, I was thinking we could do that as a part of the upcoming release, since I think we have a few others in the works?

kschoche:
> > This looks great, I left a couple smallish comments that I'd like to circle back on before merging if you get a chance. Also a reminder to add a changelog entry (especially since this is a breaking change). I think it would be helpful to disclose the spec change and how to migrate in that entry.
> 
> @kschoche Good call on the changelog updates. For the breaking changes + migration instructions, I was thinking we could do that as a part of the upcoming release, since I think we have a few others in the works?

I think that sounds fair, I'd still prefer to have breaking changes noted in the changelog + duplicated in release notes so users can get that info directly from the changelog. Migration steps could be elsewhere indeed.

benashz:
> 



> > > This looks great, I left a couple smallish comments that I'd like to circle back on before merging if you get a chance. Also a reminder to add a changelog entry (especially since this is a breaking change). I think it would be helpful to disclose the spec change and how to migrate in that entry.
> > 
> > 
> > @kschoche Good call on the changelog updates. For the breaking changes + migration instructions, I was thinking we could do that as a part of the upcoming release, since I think we have a few others in the works?
> 
> I think that sounds fair, I'd still prefer to have breaking changes noted in the changelog + duplicated in release notes so users can get that info directly from the changelog. Migration steps could be elsewhere indeed.

Yeah, I was only referring to the instruction bit.

kschoche (approved):
Looks great!

Issue #173, opened 2023-05-06
Role and RoleBinding are deployed in a different namespace from the service account


nnstt1:
**Describe the bug**

I tried installing VSO with Kustomize and Helm on ArgoCD.
But, due to the lack of definition for `metadata.namespace` in [leader-election-rbac.yaml](https://github.com/hashicorp/vault-secrets-operator/blob/v0.1.0-beta/chart/templates/leader-election-rbac.yaml), Role `leader-election-role` and RoleBinding `leader-election-rolebinding` are deployed in a different namespace from VSO service account, causing a permission error in VSO.

**To Reproduce**

1. Deploy the following manifest with ArgoCD.
2. See vault-secrets-operator error logs.

Application deployment:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
helmCharts:
  - name: vault-secrets-operator
    repo: https://helm.releases.hashicorp.com
    version: 0.1.0-beta
    releaseName: vault-secrets-operator
    namespace: vault-secrets-operator
```

Vault Secrets Operator Log:

```console
I0506 17:08:00.662082       1 request.go:690] Waited for 1.005256759s due to client-side throttling, not priority and fairness, request: GET:https://10.96.0.1:443/apis/monitoring.coreos.com/v1?timeout=32s
2023-05-06T17:08:00Z    INFO    controller-runtime.metrics      Metrics server is starting to listen    {"addr": "127.0.0.1:8080"}
{"level":"info","ts":"2023-05-06T17:08:02Z","logger":"initCachingClientFactory","msg":"Initializing the CachingClientFactory"}
2023-05-06T17:08:02Z    INFO    setup   Starting manager        {"clientCachePersistenceModel": "none", "clientCacheSize": 10000}
2023-05-06T17:08:02Z    INFO    Starting server {"path": "/metrics", "kind": "metrics", "addr": "127.0.0.1:8080"}
2023-05-06T17:08:02Z    INFO    Starting server {"kind": "health probe", "addr": "[::]:8081"}
I0506 17:08:02.206680       1 leaderelection.go:248] attempting to acquire leader lease vault-secrets-operator/b0d477c0.hashicorp.com...
[...]

PR #174, opened 2023-05-06
Add metadata.namespace to Role `leader-election-role` and RoleBinding `leader-election-rolebinding`

nnstt1:
Fix issue #173

### DIFF
--- a/chart/templates/leader-election-rbac.yaml
+++ b/chart/templates/leader-election-rbac.yaml
@@ -7,6 +7,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ include "chart.fullname" . }}-leader-election-role
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/component: controller-manager
   {{- include "chart.labels" . | nindent 4 }}
@@ -47,6 +48,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: {{ include "chart.fullname" . }}-leader-election-rolebinding
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/component: controller-manager
   {{- include "chart.labels" . | nindent 4 }}


kschoche (approved):
Hi @nnstt1, thanks for filing this PR.

I believe that the PR itself is valid and that we should be templating the namespace, so I'll approve.
I'm not 100% sure it will fix your problem in #173 since I believe helm applies a namespace to resources when you pass the namespace flag (maybe I'm incorrect there?). 

Anyways I'd love to get to the bottom of this and understand why this behaves this way with Argo CD while our CI (without Argo CD) deploys to a custom namespace always and that still works. 
Silly question but did you confirm this fixes things on your end? 
Any additional info about how things behave differently with Argo+helm vs native helm? 

Thanks!

nnstt1:
Thank you for your response, @kschoche !

Sorry, I misconfigured the `spec.destination.namespace` in ArgoCD's Application resource, so the `leader-election-role` and `leader-election-rolebinding` were not deployed in the desired namespace.

I tried to deploy the Vault Secrets Operator using the Application resource below.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: vault-secrets-operator
  namespace: argocd
spec:
  project: default
  source:
    repoURL: <REPOSITORY_URL>
    path: manifests/vault-secrets-operator
  destination:
    namespace: default
    server: https://kubernetes.default.svc
```

The above Application resource calls the following manifest (manifests/vault-secrets-operator/kustomization.yaml).

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
helmCharts:
  - name: vault-secrets-operator
    repo: https://helm.releases.hashicorp.com
    version: 0.1.0-beta
    releaseName: vault-secrets-operator
    namespace: vault-secrets-operator
```

The `spec.helmCharts` in this manifest has the same settings as the command below.

`helm install --namespace vault-secrets-operator vault-secrets-operator hashicorp/vault-secrets-operator --version 0.1.0-beta`

ArgoCD seems to output and deploy the Vault Secrets Operator manifest from kustomization.yaml with `kustomize build . --enable-helm`.

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app.kubernetes.io/instance: vault-secrets-operator
    app.kubernetes.io/name: vault-secrets-operator
    app.kubernetes.io/version: 0.1.0-beta
    helm.sh/chart: vault-secrets-operator-0.1.0-beta
  name: vault-secrets-operator-controller-manager
  namespace: vault-secrets-operator
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    app.kubernetes.io/instance: vault-secrets-operator
    app.kubernetes.io/name: vault-secrets-operator
    app.kubernetes.io/version: 0.1.0-beta
    helm.sh/chart: vault-secrets-operator-0.1.0-beta
  name: vault-secrets-operator-leader-election-role
rules:
(snip)
```

ArgoCD seems to switch the deployment namespace using the Application's `spec.destination.namespace` if there is no `metadata.namespace` in the manifest output by kustomize.

In this issue, `spec.destination.namespace` of Application and `helmCharts.namespace` of kustomization.yaml were different due to my misconfiguration, so only `leader-election-role` and `leader-election-rolebinding` were deployed to different namespaces.

On the other hand, if `spec.destination.namespace` is not specified, ArgoCD will throw an error like this:

image

Also, if the manifest is generated by calling Helm from kustomize without using ArgoCD, the manifest of `leader-election-role` and `leader-election-rolebinding` without the specified namespace will be output.
This manifest is not necessarily deployed in the same namespace as the Vault Secrets Operator.

So, if there is no special reason, I think it is better to set `metadata.namespace` for `leader-election-role` and `leader-election-rolebinding`.

kschoche:
Hi @nnstt1 - Thanks for the super detailed response, this will surely be helpful for us tracking ArgoCD integrations going forward. I went ahead and merged the PR as I believe it's the right approach regardless of other factors :).

PR #182, opened 2023-05-16
Testing Vault Secrets Operator on GKE

gautambaghel:


gautambaghel:
GKE automation works!

Makefile
```diff
@@ -114,7 +114,8 @@ VAULT_PATCH_ROOT = $(INTEGRATION_TEST_ROOT)/vault
 TF_INFRA_SRC_DIR ?= $(INTEGRATION_TEST_ROOT)/infra
 TF_VAULT_STATE_DIR ?= $(TF_INFRA_SRC_DIR)/state
 
-# directory for eks infrastructure for running tests
+# directory for cloud hosted k8s infrastructure for running tests
```
tvoran, line 117:
```suggestion
# directories for cloud hosted k8s infrastructure for running tests
```

Makefile
```diff
@@ -386,7 +387,41 @@ helm-chart: manifests kustomize helmify
 unit-test: ## Run unit tests for the helm chart
 	PATH="$(CURDIR)/scripts:$(PATH)" bats test/unit/
 
+##@ GKE
+
+.PHONY: create-gke
+create-gke: ## Create a new GKE cluster
```
tvoran, line 393: Would it make sense to add `gcloud` as a dependency when it's used in a rule?

```suggestion
create-gke: gcloud ## Create a new GKE cluster
```

tvoran (approved):

Makefile
gautambaghel: Hi @tvoran - gcloud CLI would be required for the other steps too, should I add them there too? might need to do if for AWS as well in a different PR

benashz: I wonder if we could follow the docker container `run` approach for these tools?

benashz: e.g.: https://hub.docker.com/r/google/cloud-sdk/ with something like: https://github.com/hashicorp/vault-secrets-operator/blob/0c7cc967d39bce5d63c22b5fbac03d488b7c1fdc/scripts/yq#L11

test/integration/infra/gke/main.tf
```diff
@@ -0,0 +1,105 @@
...
+  role    = "roles/storage.objectViewer"
+  member  = "serviceAccount:${google_service_account.default.email}"
+}
+
+# Give it appropriate permissions to pull the image
+resource "google_project_iam_member" "default_gar_reader" {
+  project = var.project_id
+  role    = "roles/artifactregistry.reader"
+  member  = "serviceAccount:${google_service_account.default.email}"
+}
```
benashz, line 104: nit: add end of file newline.

test/integration/infra/gke/outputs.tf
```diff
@@ -0,0 +1,27 @@
...
+
+output "gar_name" {
+  value       = google_artifact_registry_repository.vault-secrets-operator.name
+  description = "Google artifact registry repository"
+}
+
+output "oidc_discovery_url" {
+  value       = format("https://container.googleapis.com/v1/projects/%s/locations/%s/clusters/%s", var.project_id, var.region, local.cluster_name)
+  description = "GKE OIDC discovery URL"
+}
```
benashz, line 27: nit: add end of file newline.

test/integration/infra/gke/providers.tf
```diff
@@ -0,0 +1,16 @@
...
+      source  = "hashicorp/google"
+      version = "4.63.1"
+    }
+  }
+}
+
+provider "google" {
+  project = var.project_id
+  region  = var.region
+}
```
benashz, line 16: nit: add end of file newline.

test/integration/infra/gke/variables.tf
```diff
@@ -0,0 +1,16 @@
...
+}
+
+variable "project_id" {
+  description = "project id"
+}
+
+variable "region" {
+  description = "region"
+  default     = "us-west1"
+}
```
benashz, line 16: nit: add end of file newline.

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -30,7 +30,10 @@ func TestVaultAuthMethods(t *testing.T) {
 	testK8sNamespace := "k8s-tenant-" + testID
 	testKvv2MountPath := consts.KVSecretTypeV2 + testID
 	testVaultNamespace := ""
-	k8sConfigContext := "kind-" + clusterName
+	k8sConfigContext := os.Getenv("KIND_CLUSTER_CONTEXT")
```
benashz, line 33: Seems a bit odd that `KIND_CLUSTER_CONTEXT` is being used as any k8s config context. Perhaps we should rename that variable? Or add a boolean for KIND? 

``` if KIND; ... ```

File: Makefile [hunk 8]
```diff
@@ -511,6 +546,24 @@ bundle-build: ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
 
+.PHONY: gcloud
+GCLOUD = ./bin/gcloud
```
benashz, line 553: suggest using the gcloud docker image in place of the binary fetch.

Makefile
```diff
@@ -386,7 +387,41 @@ helm-chart: manifests kustomize helmify
...
+	$(GCLOUD) container clusters get-credentials $$($(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw kubernetes_cluster_name) \
+	--region $$($(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw region)
+
+# Currently only supports amd64
+.PHONY: ci-gar-build-push
+ci-gar-build-push: ## Build the operator image and push it to the GAR repository
+	@$(eval GCP_REGION := $(shell $(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw region))
+	@$(eval GCP_PROJ_ID := $(shell $(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw project_id))
+	@$(eval GCP_GAR_NAME := $(shell $(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw gar_name))
+	@$(eval IMAGE_TAG_BASE := $(GCP_REGION)-docker.pkg.dev/$(GCP_PROJ_ID)/$(GCP_GAR_NAME)/$(BIN_NAME))
```
benashz, line 408: that's a log of TF calls, I wonder if we could capture the JSON output once, the use jq to extract the values that we want?

Something like:
```
terraform output -json | jq -re '.region.value'
```

Makefile
gautambaghel: tbh, it doesn't really matter because the gcloud CLI inside the GitHub Action runner is configured via actions anyway, I've included the gcloud setup here so that if anybody from the oss community wants to try it out they can

https://github.com/hashicorp/vault-secrets-operator-ci/blob/main/.github/workflows/gke.yaml#L19

let me come back to this later, unless you guys think it's important to do it this way now?

benashz (commented):
Looks good. Had few nits/suggestions.

test/integration/vaultauthmethods_integration_test.go
gautambaghel: yeah 😅 I didn't want to hinder you guys with the new naming but ideally, it should be a generalized name? Happy to make it a part of this PR if you'd like.

Makefile
gautambaghel: good idea, let me try that

test/integration/vaultauthmethods_integration_test.go
benashz: go for it!

Makefile
gautambaghel: sure! let me give that a shot

Makefile
benashz: What you have there is fine. We just need to ensure that it is easy for the community to make contributions, although I don't see a lot of people running these make targets outside of our CI pipeline.

Makefile
gautambaghel: @benashz - I've changed this to a simple error message since this works primarily on GitHub actions where gcloud is already installed we'll leave it to other folks using this to install individually because the container + script approach doesn't work something as complex as gcloud where it sets up the docker credentials in the running machine

Makefile
```diff
@@ -386,7 +390,37 @@ helm-chart: manifests kustomize helmify
...
+##@ GKE
+
+.PHONY: create-gke
+create-gke: gcloud ## Create a new GKE cluster
+	$(TERRAFORM) -chdir=$(TF_GKE_DIR) init -upgrade
+	$(TERRAFORM) -chdir=$(TF_GKE_DIR) apply -auto-approve
+	$(GCLOUD) container clusters get-credentials $$($(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw kubernetes_cluster_name) \
+	--region $$($(TERRAFORM) -chdir=$(TF_GKE_DIR) output -raw region)
+
+include $(TF_GKE_DIR)/output.env
```
gautambaghel, line 402: @benashz - I'm loading the terraform outputs this way, I've also committed an empty env file so make doesn't throw an error.

kschoche (approved):
Looks great!

gcp.mk
```diff
@@ -0,0 +1,46 @@
...
+		-var project_id=$(GCP_PROJECT) || exit 1 \
+	rm -f $(TF_GKE_STATE_DIR)/*.tfvars
+	gcloud container clusters get-credentials $$($(TERRAFORM) -chdir=$(TF_GKE_STATE_DIR) output -raw kubernetes_cluster_name) --region $(GCP_REGION)
+
+.PHONY: import-gcp-vars
+import-gcp-vars:
+-include $(TF_GKE_STATE_DIR)/outputs.env
+
+# Currently only supports amd64
+.PHONY: ci-gar-build-push
```
benashz, line 32: Now that we have all of this logic in a sub Makefile we could we simplify the name here, something like `ci-build-push` or `build-push`

gautambaghel: ci-build-push sounds legit! Will do 👍🏾

PR #184, opened 2023-05-16
fix pki integration test flakiness

kschoche:
Fixes some techdebt around waiting for the rollout-restart to trigger in integration tests.

Previously, in CI we would wait 5 seconds at the end of the test before proceeding with attempting to validate the controller had reconciled a secret and triggered a rollout-restart. In CI this was often not enough of a wait, and simply bumping this up to 10s instead of 5s seems to allow enough time for the controller to reconcile.
https://github.com/hashicorp/vault-secrets-operator/actions/runs/4887209384/jobs/8723591069?pr=155
You can see in the run above, and its associated controller logs, that the controller gets stuck in a short reconcile loop due to upstream changes in the API server:
```
Failed to update the status     {"controller": "vaultpkisecret", "controllerGroup": "secrets.hashicorp.com", "controllerKind": "VaultPKISecret", "VaultPKISecret": {"name":"existing-only-existing-0","namespace":"k8s-tenant-izmqsh"}, "namespace": "k8s-tenant-izmqsh", "name": "existing-only-existing-0", "reconcileID": "b5dc0ac3-3963-4d97-aa0a-41719c738246", "error": "Operation cannot be fulfilled on vaultpkisecrets.secrets.hashicorp.com \"existing-only-existing-0\": the object has been modified; please apply your changes to the latest version and try again"} 
```
This never reproduces locally but does so quite frequently in CI due to lower resources.

The test would subsequently fail because the controller isn't able to reconcile the VPS secret status change when triggering the rollout-restart before 5s is up.
This PR adds a timed retry loop that has a 10s timeout around the path which checks for the rollout-restart trigger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,17 @@
 ## Unreleased
 
+Bugs:
+* Helm Chart: fix deployment templating so setting `controller.kubernetesClusterDomain` works as defined in values.yaml. [GH-183](https://github.com/hashicorp/vault-secrets-operator/pull/183)
+
 Features:
-* VaultDynamicSecrets: CRD is extended with `Revoke` field which will result in the dynamic secret lease being revoked on rotation and CR deletion. Note: The VaultAuthMethod referenced by the VDS Secret must have a policy which provides `["update"]` on `sys/leases/revoke`. [GH-143](https://github.com/hashicorp/vault-secrets-operator/pull/143)
+* VaultDynamicSecrets (VDS): CRD is extended with `Revoke` field which will result in the dynamic secret lease being revoked on rotation and CR deletion. Note: The VaultAuthMethod referenced by the VDS Secret must have a policy which provides `["update"]` on `sys/leases/revoke`. [GH-143](https://github.com/hashicorp/vault-secrets-operator/pull/143)
 * VaultAuth: Adds support for the JWT authentication method which either uses the JWT token from the provided secret reference, or a service account JWT token that VSO will generate using the provided service account. [GH-131](https://github.com/hashicorp/vault-secrets-operator/pull/131)
-* VaultDynamicSecrets: New `RenewalPercent` field to control when a lease is renewed [GH-170](https://github.com/hashicorp/vault-secrets-operator/pull/170)
+* VaultDynamicSecrets (VDS): New `RenewalPercent` field to control when a lease is renewed [GH-170](https://github.com/hashicorp/vault-secrets-operator/pull/170)
 
 Improvements:
-* VaultDynamicSecrets: Generate new credentials if lease renewal TTL is truncated [GH-170](https://github.com/hashicorp/vault-secrets-operator/pull/170)
+* VaultDynamicSecrets (VDS): Generate new credentials if lease renewal TTL is truncated [GH-170](https://github.com/hashicorp/vault-secrets-operator/pull/170)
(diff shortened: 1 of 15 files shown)


test/integration/integration_test.go
```diff
@@ -442,7 +443,21 @@ func exportKindLogs(t *testing.T) {
 	}
 }
 
-func assertRolloutRestarts(t *testing.T, ctx context.Context, client ctrlclient.Client, obj ctrlclient.Object, targets []secretsv1alpha1.RolloutRestartTarget) {
+func waitForRolloutRestartsAndAssertRollout(t *testing.T, ctx context.Context,
```
benashz, line 446:
```suggestion
func waitForRolloutRestarts(t *testing.T, ctx context.Context,
```

test/integration/vaultpkisecret_integration_test.go
```diff
@@ -311,10 +311,7 @@ func TestVaultPKISecret(t *testing.T) {
 						"VaultPKISecret", secret)
 
 					if len(vpsObj.Spec.RolloutRestartTargets) > 0 {
-						// TODO(tech-debt): add method waiting for rollout-restart, for now we
-						//  can provide an artificial grace period.
-						time.Sleep(5 * time.Second)
-						assertRolloutRestarts(t, ctx, crdClient, vpsObj, vpsObj.Spec.RolloutRestartTargets)
+						waitForRolloutRestartsAndAssertRollout(t, ctx, crdClient, vpsObj, vpsObj.Spec.RolloutRestartTargets)
```
benashz, line 314:
```suggestion
						waitForRolloutRestarts(t, ctx, crdClient, vpsObj, vpsObj.Spec.RolloutRestartTargets)
```

test/integration/vaultstaticsecret_integration_test.go
```diff
@@ -383,10 +383,7 @@ func TestVaultStaticSecret_kv(t *testing.T) {
 			}
 
 			if !expectInitial && len(obj.Spec.RolloutRestartTargets) > 0 {
-				// TODO(tech-debt): add method waiting for rollout-restart, for now we
-				//  can provide an artificial grace period.
-				time.Sleep(5 * time.Second)
-				assertRolloutRestarts(t, ctx, crdClient, obj, obj.Spec.RolloutRestartTargets)
+				waitForRolloutRestartsAndAssertRollout(t, ctx, crdClient, obj, obj.Spec.RolloutRestartTargets)
```
benashz, line 386:
```suggestion
				waitForRolloutRestarts(t, ctx, crdClient, obj, obj.Spec.RolloutRestartTargets)
```

test/integration/integration_test.go
```diff
@@ -442,7 +443,21 @@ func exportKindLogs(t *testing.T) {
...
+	client ctrlclient.Client, obj ctrlclient.Object, targets []secretsv1alpha1.RolloutRestartTarget,
+) {
+	t.Helper()
+	require.NoError(t, backoff.Retry(
+		func() error {
+			return assertRolloutRestarts(t, ctx, client, obj, targets)
```
benashz, line 452: Suggest handling the case where the `assertRolloutRestarts` failed the test. Break out of the retry early.
```suggestion
			err := assertRolloutRestarts(t, ctx, client, obj, targets)
			if t.Failed() {
				e := fmt.Errorf("assertRolloutRestarts failed")
				if err != nil {
					e = fmt.Errorf("%s, err=%w", e.Error(), err)
				}
				return backoff.Permanent(e)
			}
			return err
```

benashz (commented):
Had a few suggestions, otherwise looks good!

PR #185, opened 2023-05-17
VDS: sync CRD config to Helm chart

benashz:
When working on #172 we missed updating the Helm chart to reflect the new CRD schema.

### DIFF
--- a/chart/crds/vaultdynamicsecret-crd.yaml
+++ b/chart/crds/vaultdynamicsecret-crd.yaml
@@ -1,6 +1,7 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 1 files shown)


chart/crds/vaultdynamicsecret-crd.yaml
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
+---
```
tvoran, line 4: Not sure if it matters, but the other chart crd yaml's don't have the `---` separator.

tvoran (approved):

chart/crds/vaultdynamicsecret-crd.yaml
benashz: good catch. I will update those in another PR.

benashz: Fixed in #188

PR #186, opened 2023-05-17
VDS: Add support for specifying extra params.

benashz:
Previously it was not possible to specify the parameters when getting dynamic secrets/credentials from Vault. It should now be possible for a user to secret engine specific parameters on their VaultDynamicSecret CR.


chart/crds/vaultdynamicsecret-crd.yaml
```diff
@@ -74,6 +75,18 @@ spec:
               namespace:
                 description: Namespace where the secrets engine is mounted in Vault.
                 type: string
+              params:
+                additionalProperties:
+                  type: string
+                description: Params that can be passed when requesting credentials/secrets.
+                  Please consult https://developer.hashicorp.com/vault/docs/secrets
+                  if one is uncertain about what 'params' should/can be set to.
```
tomhjp, line 83: Our documentation style should address the reader directly as "you" (and same comment for path and the other places this doc string crops up)

```suggestion
                  if you are uncertain about what 'params' should/can be set to.
```

config/crd/bases/secrets.hashicorp.com_vaultdynamicsecrets.yaml
```diff
@@ -75,6 +75,13 @@ spec:
               namespace:
                 description: Namespace where the secrets engine is mounted in Vault.
                 type: string
+              params:
+                additionalProperties:
+                  type: string
+                description: Params that can be passed when requesting credentials/secrets.
+                  Please consult https://developer.hashicorp.com/vault/docs/secrets
+                  if one is uncertain about what 'params' should/can be set to.
```
tomhjp, line 83: Perhaps worth adding that it implies a `POST`? As HTTP verbs are the terms in which our API is written.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -221,11 +221,23 @@ func (r *VaultDynamicSecretReconciler) isRenewableLease(resp *secretsv1alpha1.Va
...
+func (r *VaultDynamicSecretReconciler) syncSecret(
+	ctx context.Context, c vault.ClientBase, o *secretsv1alpha1.VaultDynamicSecret,
+) (*secretsv1alpha1.VaultSecretLease, error) {
 	path := fmt.Sprintf("%s/%s", o.Spec.Mount, o.Spec.Path)
-	resp, err := vClient.Read(ctx, path)
-	if err != nil {
-		return nil, err
+	var err error
+	var resp *api.Secret
+	if len(o.Spec.Params) > 0 {
```
tomhjp, line 230: I like that params implies POST, as that will feel familiar from tools like curl. However, not all POST endpoints require parameters, e.g. AWS STS creds can be a POST with no params: https://developer.hashicorp.com/vault/api-docs/secret/aws#generate-credentials

And the Kubernetes secrets engine will also soon allow generating creds without any params: https://github.com/hashicorp/vault-plugin-secrets-kubernetes/pull/27

benashz: Woah, that could make things tricky. I wonder if we want to add an option to make the method explicit.

tomhjp: Yeah that was my thought too, so people get the current behaviour as the default but they can also override the method.

chart/crds/vaultdynamicsecret-crd.yaml
benashz: Fixed in https://github.com/hashicorp/vault-secrets-operator/commit/500f1e0f926cac8106c76751f8316a083cf65d90

config/crd/bases/secrets.hashicorp.com_vaultdynamicsecrets.yaml
benashz: Fixed in https://github.com/hashicorp/vault-secrets-operator/commit/500f1e0f926cac8106c76751f8316a083cf65d90

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -18,10 +18,25 @@ type VaultDynamicSecretSpec struct {
 	Namespace string `json:"namespace,omitempty"`
 	// Mount path of the secret's engine in Vault.
 	Mount string `json:"mount"`
+	// RequestHTTPMethod to use when syncing Secrets from Vault.
+	// Setting a value here is not typically required.
+	// If left unset the Operator will make requests using the GET method.
+	// In the case where Params are specified the Operator will use the PUT method.
+	// Please consult https://developer.hashicorp.com/vault/docs/secrets if you are
+	// uncertain about what method to use.
+	// +kubebuilder:validation:Enum={GET,PUT}
```
tomhjp, line 27: PUT will work as well, but in all the API docs I've checked I've only found POST for generating creds (e.g. AWS STS and k8s). Are there examples of PUT as well? I'd say POST should definitely be a valid option, and optionally include PUT as a valid option if you have any counter examples.

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -18,10 +18,25 @@ type VaultDynamicSecretSpec struct {
...
+	RequestHTTPMethod string `json:"requestHTTPMethod,omitempty"`
+	// uncertain about what 'path' should be set to.
```
tomhjp, line 29:
```suggestion
```

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -18,10 +18,25 @@ type VaultDynamicSecretSpec struct {
...
 	// Path in Vault to get the credentials for, and is relative to Mount.
 	// Please consult https://developer.hashicorp.com/vault/docs/secrets if one is
```
tomhjp, line 31:
```suggestion
	// Please consult https://developer.hashicorp.com/vault/docs/secrets if you are
```

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -18,10 +18,25 @@ type VaultDynamicSecretSpec struct {
...
 	// uncertain about what 'path' should be set to.
 	Path string `json:"path"`
+	// Params that can be passed when requesting credentials/secrets.
+	// When Params is set the configured RequestHTTPMethod will be
+	// ignored. See RequestHTTPMethod for more details.
```
tomhjp, line 36: This is counter intuitive to me. `RequestHTTPMethod` is a more detailed and explicit setting, so it should take precedence. Even if they set up something weird like a GET with a body, it's not very restful but IIRC I think Vault still accepts that and treats the body parameters as though they're query parameters.

chart/crds/vaultdynamicsecret-crd.yaml
```diff
@@ -74,6 +75,20 @@ spec:
...
+                description: Params that can be passed when requesting credentials/secrets.
+                  When Params is set the configured RequestHTTPMethod will be ignored.
+                  See RequestHTTPMethod for more details. Please consult https://developer.hashicorp.com/vault/docs/secrets
+                  if you are uncertain about what 'params' should/can be set to.
+                type: object
+              path:
+                description: uncertain about what 'path' should be set to. Path in
+                  Vault to get the credentials for, and is relative to Mount. Please
+                  consult https://developer.hashicorp.com/vault/docs/secrets if one
+                  is uncertain about what 'path' should be set to.
```
tomhjp, line 90:
```suggestion
                  consult https://developer.hashicorp.com/vault/docs/secrets if you
                  are uncertain about what 'path' should be set to.
```

config/crd/bases/secrets.hashicorp.com_vaultdynamicsecrets.yaml
```diff
@@ -75,10 +75,19 @@ spec:
...
+                  if you are uncertain about what 'params' should/can be set to.
+                type: object
               path:
-                description: Path in Vault to get the credentials for, and is relative
-                  to Mount. Please consult https://developer.hashicorp.com/vault/docs/secrets
-                  if one is uncertain about what 'path' should be set to.
+                description: uncertain about what 'path' should be set to. Path in
+                  Vault to get the credentials for, and is relative to Mount. Please
+                  consult https://developer.hashicorp.com/vault/docs/secrets if one
+                  is uncertain about what 'path' should be set to.
```
tomhjp, line 90:
```suggestion
                  consult https://developer.hashicorp.com/vault/docs/secrets if you
                  are uncertain about what 'path' should be set to.
```

tomhjp (commented):
Looking good, I think we should allow specifying POST to match up with our documentation, but the rest is nits.

api/v1alpha1/vaultdynamicsecret_types.go
benashz: Funny the client API uses `PUT` 🤔

tomhjp: lol 🤦 both PUT and POST just get translated into `logical.WriteOperation` in Vault, so it shouldn't make any difference, it's purely a documentation exercise AFAICT

api/v1alpha1/vaultdynamicsecret_types.go
tomhjp: Ah I see where this came from, leave it as it is then IMO

config/crd/bases/secrets.hashicorp.com_vaultdynamicsecrets.yaml
benashz: Fixed in 53b3b367ef1a78cb9d978916f545fc943cd081e6

chart/crds/vaultdynamicsecret-crd.yaml
benashz: Fixed in 53b3b367ef1a78cb9d978916f545fc943cd081e6

api/v1alpha1/vaultdynamicsecret_types.go
benashz: Fixed in 53b3b367ef1a78cb9d978916f545fc943cd081e6

api/v1alpha1/vaultdynamicsecret_types.go
benashz: Fixed in 53b3b367ef1a78cb9d978916f545fc943cd081e6

tomhjp (approved):
LGTM! 👍

benashz:
Failing tests were introduced in #184 , working on a fix for those in #191.

Going to merge this one!

PR #188, opened 2023-05-18
Sync up chart CRDs

benashz:
The `config/crd/bases` directory is the canonical source for all CRDs, so we need to ensure that the CRDs deployed from the Helm chart are always in sync with it.

This PR adds the `sync-crds` make target that ensures that CRDs are always in sync. In order to make it simpler, we simply copy the files directly. Helm does not treat them as templates per: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

Please note, I had to do the initial sync on in the first commit so that Git would detect the rename properly. The VaultAuth from `chart/crds` had some strange whitespace differences compared to the others.

Please review each commit independently.


kschoche (approved):
I have been manually removing the - - - in the past. Very happy to leave them in now! Thanks for this change!

PR #200, opened 2023-05-22
Support setting kv-v2 version for VaultStaticSecret

tomhjp:
Adds support for specifying the [version](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#version) of a kv-v2 secret, so that when there are different personas managing the secret contents and the consuming application, they can more easily coordinate rollout and rollback.


tvoran (approved):

kschoche (approved):
Looks great!

PR #202, opened 2023-05-23
optionally revoke and purge all cached vault clients upon operator deployment deletion

thyton:
The revocation of tokens that are stored in K8s secrets will be done by the `pre-delete` hook as it's less likely to fail due to the existing `hook-succeeded` policy of the hook. The revocation of in-memory cached tokens will be done by the operator's leader pod.
 
* Use the operator deployment OwnerReference to allow auto cleanup upon the deployment deletion
* Add the operator deployment OwnerReference to HMAC key Secret
* Add two new annotations to the operator and the pre-delete hook to "communicate" when the pre-delete hook started and the operator pod finished revoking in-memory cached tokens
    ```
        vso.secrets.hashicorp.com/pre-delete-hook-started: "false" with a downward API pre-delete-hook-started under podinfo
        vso.secrets.hashicorp.com/in-memory-vault-tokens-revoked: "false"
    ```
* Add a command line flag `vault-token-revocation-required` to toggle the option to revoke and purge all cached Vault clients
* If the flag is on
  * Add the operator deployment OwnerReference to all token Secrets    
  * Add to the operator a watcher for pre-delete-hook-started file
  * Add to the pre-delete hook a watcher for `in-memory-vault-tokens-revoked`
  * When in-memory-vault-tokens-revoked is "true", the operator will
    * First, disable ClientFactory interface
    * Then, revoke all client tokens that are cached in-memory
    * Finally, annotate the operator pods in-memory-vault-tokens-revoked: "true" 
  * When vault-token-revocation-required is "true", the pre-delete hook will revoke all client tokens that are stored in K8s secrets


chart/templates/deployment.yaml
```diff
@@ -105,7 +110,7 @@ spec:
       securityContext:
         runAsNonRoot: true
       serviceAccountName: {{ include "chart.fullname" . }}-controller-manager
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds : {{ .Values.controller.terminationGracePeriodSeconds }}
```
tomhjp, line 113: Given the pre-delete hook and owner references handle all the Kubernetes object deletions, is it possible to get rid of the config map by using this grace period and a `preStop` hook for the in-memory-only leases?

The tricky part is the pod needs to know the difference between the deployment getting deleted (revoke everything) and any other reason a pod could be deleted (do nothing). One signal we could send that doesn't require any RBAC permissions (and thus avoids a race condition with the RBAC being deleted) is an annotation that's projected into the pod via the volume downward API. The pre-delete hook could update that annotation on all the operator pods to say they should revoke everything on shutdown. WDYT?

internal/helpers/deployment.go
```diff
@@ -0,0 +1,139 @@
...
+		key       = getDeploymentStatusConfigMapKey()
+	)
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Error(ctx.Err(), "failed to await manager acked")
+			return
+		default:
+			if err := c.Get(ctx, key, &configMap); err != nil {
```
tomhjp, line 68: It requires a slightly expanded client interface, but the k8s client package has a `WithWatch` interface that wraps the `client.Client` currently available which could be used to more efficiently wait for updates to the config map: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#WithWatch. Happy to pair up on helping you get going with that, as the documentation isn't super detailed.

tomhjp (commented):
Sorry for the slow review, I've been reflecting a lot on the high level strategy here so I haven't gone too detailed on the review yet. I think the config map can work but I'm just doing a bit more searching for simplifications before we commit. LMK what you think - it also might be easier to have a quick sync chat about it.

chart/templates/deployment.yaml
thyton: > a preStop hook for the in-memory-only leases

https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/. Looks like the `preStop` is a command. I'm not sure how to use the `preStop` hook to revoke in-memory-only leases of the running process/command `/vault-secrets-operator` though they are all in the same container. I wonder if you meant using it to signal `/vault-secrets-operator` to revoke the leases.

> One signal we could send that doesn't require any RBAC permissions (and thus avoids a race condition with the RBAC being deleted) is an annotation that's projected into the pod via the volume downward API. The pre-delete hook could update that annotation on all the operator pods to say they should revoke everything on shutdown

Using the annotation is a good idea. I wonder if there is a concept of watchers for annotations. I will research a bit on that.

tomhjp: For the `preStop` command, we could do something like `kill -s HUP 1`. Then the operator can listen for that HUP signal and know that it should check the annotation value and start revoking if the deployment is being deleted: https://pkg.go.dev/os/signal#example-Notify

For the annotation, the downward API can project a specific annotation into a file within the pod: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-fieldRef. From there, you just need to know when to check the file. [fsnotify](https://github.com/fsnotify/fsnotify) is a good library for that generally, but I don't think it would be required in this case, because we can just listen for that HUP signal from the `preStop` hook.

tomhjp: I just want to stress as well I'm not totally sure this will work! And I'm not totally confident it would turn out better/simpler. Just exploring ideas :-)

benashz: > For the `preStop` command, we could do something like `kill -s HUP 1`. Then the operator can listen for that HUP signal and know that it should check the annotation value and start revoking if the deployment is being deleted: https://pkg.go.dev/os/signal#example-Notify

Would that require enabling shared process namespace?

> 
> For the annotation, the downward API can project a specific annotation into a file within the pod: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-fieldRef. From there, you just need to know when to check the file. [fsnotify](https://github.com/fsnotify/fsnotify) is a good library for that generally, but I don't think it would be required in this case, because we can just listen for that HUP signal from the `preStop` hook.

I like the idea of using inotify here, but I would hesitate to add it to the Operator especially when it comes to cross platform compatibility.

benashz: In this case, I think best effort should be good enough. The token revocation is already best effort.

thyton: @tomhjp. I found "updating the pod annotations will replace the pod" info on a random source. Sorry I should have tested it prior. Thanks for the example.

internal/vault/cache.go
```diff
@@ -133,6 +134,14 @@ func (c *clientCache) Prune(filterFunc ClientCachePruneFilterFunc) []ClientCache
 	return pruned
 }
 
+func (c *clientCache) Keys() []ClientCacheKey {
+	var keys []ClientCacheKey
```
tomhjp, line 138: nit: May as well allocate all the slice memory upfront for efficiency.

```suggestion
	keys := make([]ClientCacheKey, len(c.cache.Keys(), 0)
```

tomhjp (commented):
I haven't finished reviewing, but wanted to leave an in-progress review to say I think the overall approach looks good, just some minor comments so far.

main.go
```diff
@@ -207,6 +223,40 @@ func main() {
...
+			helpers.AwaitInMemoryVaultTokensRevoked(preDeleteDeadlineCtx, cleanupLog, defaultClient)
+			clientFactory.RevokeAllInStorage(preDeleteDeadlineCtx, defaultClient)
+			return
+		}
+
+		preDeleteStartedCtx, preDeleteHandler := context.WithCancel(context.Background())
+		go helpers.AwaitPreDeleteStarted(ctx, preDeleteHandler, setupLog)
+
+		go func() {
+			for {
```
tomhjp, line 241: I think the `for` loop here is unnecessary.

internal/vault/client_factory.go
```diff
@@ -242,6 +337,11 @@ func (m *cachingClientFactory) Get(ctx context.Context, client ctrlclient.Client
 		)
 	}()
 
+	if m.isDisabled() {
```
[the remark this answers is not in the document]
thyton: Yes. The disable mechanism is to "block" future calls of ClientFactory interface. I added this note to `func (m *cachingClientFactory) Disable() `
```
// NOTE: If a consumer of the client cache factory already has a reference to a client in hand,
// they may continue using it and generating new tokens/leases.
```

main.go
```diff
@@ -207,6 +223,40 @@ func main() {
...
+				cleanupLog.Error(err, fmt.Sprintf("failed to annotate %s", helpers.AnnotationPreDeleteHookStarted))
+				return
+			}
+			helpers.AwaitInMemoryVaultTokensRevoked(preDeleteDeadlineCtx, cleanupLog, defaultClient)
+			clientFactory.RevokeAllInStorage(preDeleteDeadlineCtx, defaultClient)
+			return
+		}
+
+		preDeleteStartedCtx, preDeleteHandler := context.WithCancel(context.Background())
+		go helpers.AwaitPreDeleteStarted(ctx, preDeleteHandler, setupLog)
```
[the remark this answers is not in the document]
thyton: I was thinking of one goroutine designated watching the pre-delete hook started annotation and triggering a context handler, and we can have multiple goroutines later on for doing different things based on that context. It's probably too mechanic.

I switched to one goroutine for simplicity. Thanks! :)

chart/templates/manager-rbac.yaml
```diff
@@ -74,6 +74,14 @@ rules:
...
   - patch
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - patch
```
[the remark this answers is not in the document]
thyton: [My current approach ](https://github.com/hashicorp/vault-secrets-operator/blob/d96e5c1596a4ea2c8245e8311c2a3718519dd735/internal/helpers/deployment.go#L106-L123)for annotating pods is to get the list of pods that have the control plane label and annotate all pods. The reason for annotating all pods is in the case the leader pod dies for some reason, other pods can be used. I think it's a partial solution so it doesn't cover all failure cases. I wonder if the permission should be added to a new role not the existing cluster role, which is more than what's needed.

internal/helpers/deployment.go
```diff
@@ -0,0 +1,119 @@
...
+					}
+				}
+			}
+			time.Sleep(300 * time.Millisecond)
+		}
+	}
+}
+
+func AwaitPreDeleteStarted(ctx context.Context, handler context.CancelFunc, logger logr.Logger) {
+	for {
```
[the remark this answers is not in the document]
thyton: Per our DM, we decided to stick with the loop polling file instead of the `preStop` command for now due to the following reasons 
1. We use a distroless build for the manager container. `kill` won’t be available on distroless https://github.com/hashicorp/vault-secrets-operator/blob/b65708ca82157dd4102dea341aa6647f3cd43b5e/Dockerfile#L38
2. The only thing we can do is run the operator binary, and that could implement sending the HUP signal (for example, using `client-go` to remote exec a command like `vault-secrets-operator --start-shutdown`) to the original operator process. However, it's an anti-pattern to remote exec a command in the operator pod from a pre-delete hook pod.
3.  `preStop` is not applicable to the current design. The operator pod waits for the pre-delete hook to start before revoking the tokens, and the pre-delete hook waits for the operator pod's in-memory token revocation to finish.  `preStop` runs as the operator pod is getting shutdown. The “stopping” process only starts when the pod receives a deletion timestamp, which is after the pre-delete hook finishes.

internal/vault/utils.go
```diff
@@ -3,9 +3,51 @@
...
 	return strings.Join(parts, "/")
 }
+
+func getOperatorDeployment(c client.Client) (*appsv1.Deployment, error) {
+	operatorDeploymentName := ""
+	if operatorDeploymentName = os.Getenv("OPERATOR_DEPLOYMENT_NAME"); operatorDeploymentName == "" {
+		operatorDeploymentName = operatorDeploymentDefaultName
+	}
+	deployment := appsv1.Deployment{}
+	err := c.Get(context.Background(), types.NamespacedName{
```
benashz, line 35: any chance we could update the signature to the context?

internal/vault/utils.go
```diff
@@ -3,9 +3,51 @@
...
+		Namespace: common.OperatorNamespace,
+		Name:      operatorDeploymentName,
```
benashz, line 37: What if `operatorDeploymentName` is not known?

internal/vault/utils.go
```diff
@@ -3,9 +3,51 @@
...
+)
 
 // JoinPath for Vault requests.
 func JoinPath(parts ...string) string {
 	return strings.Join(parts, "/")
 }
+
+func getOperatorDeployment(c client.Client) (*appsv1.Deployment, error) {
+	operatorDeploymentName := ""
+	if operatorDeploymentName = os.Getenv("OPERATOR_DEPLOYMENT_NAME"); operatorDeploymentName == "" {
```
benashz, line 31: I wonder if there might be a better way of getting the deployment name. This approach feels a bit brittle. Maybe we could find the Deployment based on some known label?

main.go
```diff
@@ -77,13 +84,22 @@ func main() {
 				"choices=%v", []string{persistenceModelDirectUnencrypted, persistenceModelDirectEncrypted, persistenceModelNone}))
 	flag.IntVar(&vdsOptions.MaxConcurrentReconciles, "max-concurrent-reconciles-vds", 100,
 		"Maximum number of concurrent reconciles for the VaultDynamicSecrets controller.")
-	flag.BoolVar(&finalizerCleanup, "finalizer-cleanup", false, "Remove finalizers from all CRs in preparation for shutdown.")
+	flag.BoolVar(&preDeleteHook, "pre-delete-hook", false, "Run as helm pre-delete hook")
+	flag.BoolVar(&revokeVaultTokensOnUninstall, "revoke-vault-tokens-on-uninstall", false,
+		"Revoke all cached Vault client tokens on Helm uninstall.")
+	flag.BoolVar(&pruneVaultTokensOnUninstall, "prune-vault-tokens-on-uninstall", false,
+		"Prune all Vault client tokens in storage on Helm uninstall.")
```
benashz, line 91: I wonder if we collapse these two flags into one? Similar to the way that we do for configuring the storage cache.

main.go
```diff
@@ -118,27 +134,39 @@ func main() {
...
+		Scheme: scheme,
+	})
+	if err != nil {
+		setupLog.Error(err, "Failed to instantiate a default Client")
+		os.Exit(1)
+	}
 
-		d := time.Now().Add(time.Second * 60)
-		shutdownCtx, cancel := context.WithDeadline(context.Background(), d)
+	if preDeleteHook {
```
benashz, line 145: I wonder if we could maybe rename `preDeleteHook` to `shutdown` ?

main.go
```diff
@@ -207,6 +221,49 @@ func main() {
 		}
 	}
 
+	if revokeVaultTokensOnUninstall {
```
benashz, line 224: This bit is a little bit hard to parse. We are checking `preDeleteHook` twice, starting a watcher that may not be necessary. I wonder if we could move all of the shutdown handling higher up, maybe to its own function, then have something like:
```
if shutdown:
  shutdown()
else:
   do the normal stuff
```

internal/helpers/deployment.go
```diff
@@ -0,0 +1,110 @@
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func IsDeploymentShutdown(m *corev1.ConfigMap) bool {
+	val, ok := m.Data[deploymentShutdown]
+	return ok && val == StringTrue
```
benashz, line 74: Could we maybe use https://pkg.go.dev/strconv#ParseBool instead?

internal/helpers/deployment.go
```diff
@@ -0,0 +1,110 @@
...
+	watcher, err := c.Watch(ctx, &list)
+	if err != nil {
+		return nil, fmt.Errorf("failed to watch the manager configmap err=%s", err)
+	}
+	return watcher, nil
+}
+
+type OnConfigMapChange func(context.Context, *corev1.ConfigMap, client.Client) error
+
+func WaitForManagerConfigMapModified(ctx context.Context, logger logr.Logger, watcher watch.Interface, c client.Client, evaluator func(*corev1.ConfigMap) bool, onChanges ...OnConfigMapChange) {
```
benashz, line 51: nit: this line is quite long anyway we could shorten it a bit.

internal/vault/utils.go
```diff
@@ -3,9 +3,51 @@
...
+}
+
+func GetOperatorDeploymentOwnerReference(c client.Client) (*metav1.OwnerReference, error) {
+	deployment, err := getOperatorDeployment(c)
+	if err != nil {
+		return nil, err
+	}
+	return &metav1.OwnerReference{
+		APIVersion: operatorDeploymentAPIVersion,
+		Kind:       operatorDeploymentKind,
```
benashz, line 49: These should probably come from the `Deployment` instance. I think you can probably replace this with https://github.com/hashicorp/vault-secrets-operator/blob/acf29a5148b525fab4246d3eb15a9b1ffa3d188a/internal/helpers/secrets.go#L89 - you could make that function public instead.

internal/vault/utils.go
benashz, line 1: I think this code actually should go in `internal/helpers`

internal/helpers/deployment.go
benashz, line 1: nit: filename does not really match up with its contents, perhaps it is left over from the previous approach.

internal/vault/cache_storage.go
```diff
@@ -585,8 +590,9 @@ func (c *defaultClientCacheStorage) incrementOperationCounter(operation string,
 type ClientCacheStorageConfig struct {
 	// EnforceEncryption for persisting Clients i.e. the controller must have VaultTransitRef
 	// configured before it will persist the Client to storage. This option requires Persist to be true.
-	EnforceEncryption bool
-	HMACSecretObjKey  ctrlclient.ObjectKey
+	EnforceEncryption           bool
+	PruneVaultTokensOnUninstall bool
```
benashz, line 594: I fell like this variable name could be confusing, what does uninstall mean for ClientCacheStorage?. I wonder if it should be something related to how the secrets are created - it could even just be pointer to some `OwnerReferences`, so if those are set just include the owner refs for any stored Secret?

internal/vault/cache_storage.go
```diff
@@ -625,10 +642,11 @@ func NewDefaultClientCacheStorage(ctx context.Context, client ctrlclient.Client,
 	}
 
 	cacheStorage := &defaultClientCacheStorage{
-		hmacSecretObjKey:  config.HMACSecretObjKey,
-		hmacKey:           s.Data[hmacKeyName],
-		enforceEncryption: config.EnforceEncryption,
-		logger:            zap.New().WithName("ClientCacheStorage"),
+		enforceEncryption:           config.EnforceEncryption,
+		pruneVaultTokensOnUninstall: config.PruneVaultTokensOnUninstall,
+		deploymentOwnerRef:          deploymentOwnerRef,
```
benashz, line 647: this should probably be `ownerRefs` then you can drop the extra variable above.

benashz (commented):
Getting there!

internal/helpers/deployment.go
```diff
@@ -0,0 +1,130 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	deploymentShutdown         = "deploymentShutdown"
```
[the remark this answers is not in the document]
thyton: This is for Deployment deletion or Helm uninstall (which will trigger Deployment deletion).

internal/vault/utils.go
thyton: Nice. Thanks!

internal/helpers/deployment.go
```diff
@@ -0,0 +1,130 @@
...
+	inMemoryVaultTokensRevoked = "inMemoryVaultTokensRevoked"
+	StringTrue                 = "true"
+	managerConfigMapNameEnv    = "MANAGER_CONFIGMAP_NAME"
```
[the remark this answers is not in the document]
thyton: Thanks!

internal/helpers/deployment.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	deploymentShutdown         = "deploymentShutdown"
+	inMemoryVaultTokensRevoked = "inMemoryVaultTokensRevoked"
+	StringTrue                 = "true"
```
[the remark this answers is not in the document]
thyton: We do. It's a field in the manager configmap for the operator to notify the pre-delete hook once it finishes revoking in-memory tokens. The operator does the in-memory revocation first, then the pre-delete does the in-storage one.

internal/vault/cache_storage.go
thyton: I agree. I will move the logic to populate `OwnerReferences` during the setup and drop other fields.

main.go
thyton: Do you mean something like `VaultTokensCleanUpModelOnUninstall` with values `revoke`, `prune`, `revoke-and-prune`?

main.go
thyton: Both the operator and the pre delete hook need the watcher, but they use different contexts. I thought adding a `preDeleteHook` check before starting the watcher would help get away with 2 different watcher setup code blocks.

https://github.com/hashicorp/vault-secrets-operator/blob/5c487a942bbb2b0faaa98faa845b62ef203f1cac/main.go#L230-L233

thyton: I will look into it but am unsure that we can move much higher up in a neater way since the pre-delete hook needs the client factory.

config/manager/kustomization.yaml
```diff
@@ -6,6 +6,8 @@ resources:
 
 generatorOptions:
   disableNameSuffixHash: true
+  labels:
```
kschoche, line 9: I believe we've chatted about why this label is added: because we're starting the watcher in the operator even when the helm chart has the revocation logic explicitly disabled. I think that's a separate discussion and we can iterate on that approach later on. But, adding this one label to the manager and none of the other kustomize resources seems a bit odd. For reference, helm adds it to all resources not just the deployment.
nit: you probably don't need to go and add the label to every resource in the config directory, but if there is an easy one-line approach to adding a label to all resources in Kustomize I'd like that instead :)

internal/helpers/secrets.go
```diff
@@ -38,21 +37,46 @@ var OwnerLabels = map[string]string{
...
+}
 
-	gvk, err := apiutil.GVKForObject(owner, scheme)
-	if err != nil {
-		return ownerRef, err
+// NewSyncableSecretMetaData returns SyncableSecretMetaData if obj is a supported type.
+// An error will be returned of obj is not a supported type.
+//
+// Supported types for obj are: VaultDynamicSecret, VaultStaticSecret. VaultPKISecret
+func NewSyncableSecretMetaData(obj ctrlclient.Object) (*SyncableSecretMetaData, error) {
```
kschoche, line 57: I think this is a merge issue, we recently moved `NewSyncableSecretMetadata` to `internal/common/common.go` and modified it to return `AuthRef` as well, see here: https://github.com/hashicorp/vault-secrets-operator/blob/8a710f0feb6e4e1bd6f1aba816de43a1cd29f8eb/internal/common/common.go#L322

This change will probably break namespaces?

internal/vault/cache.go
```diff
@@ -134,7 +157,6 @@ func (c *clientCache) Prune(filterFunc ClientCachePruneFilterFunc) []ClientCache
 }
 
 func (c *clientCache) remove(key ClientCacheKey, client Client) bool {
-	client.Close()
```
kschoche, line 137: I don't recall why this was changed, why is `client.Close()` removed here?

internal/vault/cache_storage.go
```diff
@@ -183,6 +186,9 @@ func (c *defaultClientCacheStorage) Store(ctx context.Context, client ctrlclient
 		return nil, err
 	}
 
+	// TODO we currently don't pass in owner references
```
kschoche, line 189: nit: Something like this?
```suggestion
	// TODO: we currently don't pass in owner references to avoid deleting the cache on restart.
```

internal/vault/client.go
```diff
@@ -361,14 +361,23 @@ func (c *defaultClient) GetTokenSecret() *api.Secret {
 
 // Close un-initializes this Client, stopping its LifetimeWatcher in the process.
 // It is safe to be called multiple times.
```
kschoche, line 363:
```suggestion
// Close un-initializes this Client, stopping its LifetimeWatcher in the process and optionally revoking the token.
// It is safe to be called multiple times.
```

internal/vault/client_factory.go
```diff
@@ -58,11 +65,14 @@ type cachingClientFactory struct {
 	recorder           record.EventRecorder
 	persist            bool
 	encryptionRequired bool
+	shutDown           bool
 	// clientCacheKeyEncrypt is a member of the ClientCache, it is instantiated whenever the ClientCacheStorage has enforceEncryption enabled.
 	clientCacheKeyEncrypt  ClientCacheKey
 	logger                 logr.Logger
 	requestCounterVec      *prometheus.CounterVec
 	requestErrorCounterVec *prometheus.CounterVec
+	revokeOnEvict          bool
```
kschoche, line 74: great work on the naming conventions here :)

internal/vault/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+
+type OnConfigMapChange func(context.Context, client.Client, *corev1.ConfigMap)
+
+func WaitForManagerConfigMapModified(ctx context.Context, watcher watch.Interface, c client.Client, onChanges ...OnConfigMapChange) {
+	defer watcher.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case event := <-watcher.ResultChan():
```
kschoche, line 69: I feel like we should return from the for loop after this triggers because any update to our configmap while we are processing shutdown will cause us to enter this path again. What do you think?

internal/vault/configmap.go
```diff
@@ -0,0 +1,211 @@
...
+	var list corev1.ConfigMapList
+	if err := c.List(ctx, &list, opts...); err != nil {
+		return nil, err
+	}
+
+	if len(list.Items) == 0 {
+		return nil, fmt.Errorf("no configmaps matching labels=%v found in the operator namespace", labels)
+	}
+
+	if len(list.Items) > 1 {
```
kschoche, line 99: I'm wondering if it make sense to not return error in this case, and just log that we found more than one and return the first element so that we can proceed still? It looks like we're only processing element `0` anyways in the calling function. Thoughts?

main.go
```diff
@@ -270,10 +296,43 @@ func main() {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			if ctx.Err() != nil {
+				return fmt.Errorf("shutdown context canceled err=%s", ctx.Err())
+			}
+			return nil
+		default:
+			time.Sleep(500 * time.Millisecond)
```
kschoche, line 324: can we please add a comment about why we are sleeping here?

main.go
```diff
@@ -12,6 +12,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/go-logr/logr"
```
kschoche, line 15: I'm trying to see why this change occurred,  but I cant spot anything, did we bring in the wrong logger somewhere?

test/integration/revocation/terraform/main.tf
```diff
@@ -0,0 +1,171 @@
+# Copyright (c) HashiCorp, Inc.
```
kschoche, line 1: Reminder to re-run the copywrite tool so we have the correct license (BSL) on all of these files.

test/integration/revocation/terraform/main.tf
```diff
@@ -0,0 +1,171 @@
...
+  backend                          = vault_kubernetes_auth_backend_config.default.backend
+  role_name                        = var.auth_role
+  bound_service_account_names      = [kubernetes_default_service_account.default.metadata[0].name]
+  bound_service_account_namespaces = [kubernetes_namespace.tenant-1.metadata[0].name]
+  token_ttl                        = 3600
+  token_policies                   = [vault_policy.default.name]
+  audience                         = "vault"
+}
+
+# jwt auth config
```
kschoche, line 116: I believe all of the jwt auth stuff is unnecessary.

test/integration/revocation/terraform/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+
+variable "k8s_config_path" {
+  default = "~/.kube/config"
+}
+
+variable "k8s_host" {
+  default = "https://kubernetes.default.svc"
+}
+
+variable "k8s_ca_pem" {
```
kschoche, line 26: I think the oidc related stuff, along with the k8s_ca_pem is unnecessary for the test.

test/integration/revocation_integration_test.go
```diff
@@ -0,0 +1,308 @@
...
+	b, err := json.Marshal(terraform.OutputAll(t, tfOptions))
+	require.Nil(t, err)
+
+	var outputs revocationK8sOutputs
+	require.Nil(t, json.Unmarshal(b, &outputs))
+
+	// Set the secrets in vault to be synced to kubernetes
+	vClient := getVaultClient(t, testVaultNamespace)
+
+	auths := []*secretsv1beta1.VaultAuth{
```
kschoche, line 102: Okay - now I see why the tf code had the JWT/oidc setup resources in it.
nit: I'm pretty sure that the test is still going to be valid (and doesnt really change?) using only the k8s auth method which can simplify the test a bit? Not going to block on this, but would be nice to trim it up and use only k8s auth.

test/unit/deployment.bats
```diff
@@ -243,6 +243,32 @@ load _helpers
...
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/deployment.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].args | select(documentIndex == 2)' | tee /dev/stderr)
+
+  local actual=$(echo "$object" | yq 'contains(["--pre-delete-hook-timeout-seconds=120"])' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
kschoche, line 271: Can we add a test for the revocation command arg that is added in this part of the template:
```
        {{- if .Values.controller.manager.clientCache.revokeClientCacheOnUninstall }}
        - --revoke-client-cache
        {{- end }}
```

kschoche (commented):
Great work @thyton !!! 
This is looking really close, I left a few comments here and there but I like the approach and happy to give a 👍🏽 once you get a chance to go through them!

main.go
thyton: I used https://pkg.go.dev/github.com/go-logr/logr#NewContext in main.go to get a Context from a logger. But I revisited and learned it's not necessary.

internal/vault/cache.go
thyton: [Our onEvict function calls `client.Close()`](https://github.com/hashicorp/vault-secrets-operator/blob/main/internal/vault/client_factory.go#L123C37-L134), and the lru cache remove function calls this onEvict function `github.com/hashicorp/golang-lru@v0.5.4/simplelru/lru.go:175`.

test/integration/revocation_integration_test.go
thyton: I will trim and simplify the test. Thank you!

config/manager/kustomization.yaml
thyton: Thank you! I've switch to this one-line approach https://github.com/hashicorp/vault-secrets-operator/blob/371ef5d9042e7739dfca1356e7be642fdda0ba70/config/default/kustomization.yaml#L15-L16

internal/vault/configmap.go
tvoran, line 1: Have we considered adding some unit testing here, using the fake controller-runtime k8s client? (Just a thought, not sure it's worth the trouble.)

test/integration/revocation_integration_test.go
```diff
@@ -0,0 +1,308 @@
...
+	})
+
+	// Check if all test tokens were revoked.
+	for _, accessor := range testTokenAccessors {
+		// expect to receive an error that contains "invalid accessor" when looking up the token using its accessor
+		// as an indication that the token was successfully revoked
+		_, err = getTokenData(t, accessor)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "invalid accessor")
+	}
```
tvoran, line 298: I love that we can test all this for real. Nice job!

thyton: Thank you!

internal/vault/client_factory.go
```diff
@@ -223,6 +245,48 @@ func (m *cachingClientFactory) RestoreAll(ctx context.Context, client ctrlclient
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.logger.Info("Shutting down ClientFactory")
+	// This will shut down client cache, "blocking" future ClientFactory interface calls
+	// NOTE: If a consumer of the client cache factory already has a reference to a client in hand,
+	// they may continue using it and generating new tokens/leases.
+	m.shutDown = true
+
+	var errs error
```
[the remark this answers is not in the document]
thyton: Good catch! Will remove it.

internal/vault/configmap.go
thyton: I agree with you. I will refactor the on change function a bit to avoid the case.

internal/vault/configmap.go
thyton: I removed it the len > 1 error. Not sure if the case there are more than 1 manager ConfigMap existing in the operator. I can put a note/TODO for now. We can reiterate in the next PR?

internal/vault/configmap.go
thyton: I will look into it

internal/vault/configmap_test.go
```diff
@@ -0,0 +1,79 @@
...
+			true,
+		},
+	}
+
+	for _, tt := range tests {
+		onConfigMapChange := OnShutDown(&clientFactory)
+		actual := onConfigMapChange(ctx, client, tt.cm)
+		assert.Equal(t, tt.expected, actual)
+
+		if tt.expected {
```
kschoche, line 71: nit: I think we could use an assertion when `tt.expected=false` so that we are asserting correct behaviour on both paths.

PR #203, opened 2023-05-23
do not delete the vaultAuthMethod before checking for lease revocation status

kschoche:
Previously we were checking for lease revocation state for VDS secrets which have the `Revoke` field set, but AFTER we'd deleted the VaultAuthMethod/Connection CRDs, this leads to a race on the `default` AuthMethod being deleted and the VDS controller being able to use a Client in order to issue the revocation, causing the VDS integration tests to flake with un-revoked leases.


tvoran (approved):

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -363,10 +357,19 @@ func TestVaultDynamicSecret(t *testing.T) {
...
-			return "", fmt.Errorf("Leases still found: %d", len(keys))
+			// Print out the lease ids that are still found to make debugging easier.
+			return "", fmt.Errorf("leases still found: %v", keys)
 		}
 		return "", nil
 	})
+	// Delete custom resources which were created before tf destroy removes the k8s namespace.
+	// This includes the VaultAuthMethods/Connections which are required for revocation to complete,
+	// so do it after we test for revocation.
+	for _, c := range created {
```
benashz, line 368: So the revocation is being triggered a reason other than the CR being deleted?

benashz, line 368: Should this be done in a `t.Cleanup()`

kschoche: It turns out that only the auth+conn objects end up in the `created` list, and we have a secondary list called `objsCreated` which holds the VDS objects which gets cleaned up as well in a `t.Cleanup()` function inside of each test: https://github.com/hashicorp/vault-secrets-operator/blob/a3a2d18a03caebd4de4688e532a0c3fb2b3884fd/test/integration/vaultdynamicsecret_integration_test.go#L224.

My understanding is that the `t.Cleanup()` [inside](https://github.com/hashicorp/vault-secrets-operator/blob/a3a2d18a03caebd4de4688e532a0c3fb2b3884fd/test/integration/vaultdynamicsecret_integration_test.go#L224) of the `t.Run()` will trigger at the end of `t.Run()`, and the `t.Cleanup()` [which holds](https://github.com/hashicorp/vault-secrets-operator/blob/a3a2d18a03caebd4de4688e532a0c3fb2b3884fd/test/integration/vaultdynamicsecret_integration_test.go#L98) the `tf destroy` will trigger at the end of the test??

kschoche: Moved this up into the main `t.Cleanup()`  in 179e8ef

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -99,6 +99,13 @@ func TestVaultDynamicSecret(t *testing.T) {
 		if !skipCleanup {
 			exportKindLogs(t)
 
+			// Deletes the VaultAuthMethods/Connections.
```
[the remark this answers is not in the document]
kschoche: Thanks! I was on the fence on this, but saw it used this way in the PKI test so that's where I grabbed it, will move it.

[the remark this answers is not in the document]
benashz: we should probably fix it in the PKI tests as well.

PR #208, opened 2023-05-24
create our own serviceaccount in auth methods test to ensure it exists

kschoche:
The auth method integration tests are experiencing some race conditions on the asynchronous creation of the default service-account via the creation of the test namespace. This causes TF to fail the creation of the `default-sa-secret` resource because it references the `default` serviceaccount via annotation and Kubernetes blocks creation of the secret if the serviceaccount it references does not exist yet.

This is probably a bug similar to this: https://github.com/hashicorp/terraform-provider-kubernetes/issues/1943

* Create our own serviceaccount for the namespace in the vault-auth-method test.


test/integration/vaultauthmethods_integration_test.go
```diff
@@ -138,7 +136,7 @@ func TestVaultAuthMethods(t *testing.T) {
 				Mount:     "jwt",
 				JWT: &secretsv1alpha1.VaultAuthConfigJWT{
 					Role:           outputs.AuthRole,
-					ServiceAccount: testServiceAccountName,
+					ServiceAccount: "default",
```
benashz: similar to the above, about using the constant.

Makefile
```diff
@@ -275,7 +275,7 @@ integration-test: set-image setup-vault ## Run integration tests for Vault OSS
 	SUPPRESS_TF_OUTPUT=$(SUPPRESS_TF_OUTPUT) SKIP_CLEANUP=$(SKIP_CLEANUP) OPERATOR_NAMESPACE=$(OPERATOR_NAMESPACE) \
 	OPERATOR_IMAGE_REPO=$(IMAGE_TAG_BASE) OPERATOR_IMAGE_TAG=$(VERSION) \
     INTEGRATION_TESTS=true KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) KIND_CLUSTER_CONTEXT=$(KIND_CLUSTER_CONTEXT) CGO_ENABLED=0 \
-	go test github.com/hashicorp/vault-secrets-operator/test/integration/... $(TESTARGS) -timeout=45m
+	go test github.com/hashicorp/vault-secrets-operator/test/integration/... $(TESTARGS) -timeout=30m
```
benashz: 👍

benashz (approved):
A few comments/suggestions, otherwise 👍

PR #209, opened 2023-05-24
VDS: only revoke on delete

tvoran:
Instead of also revoking a previous lease after renewal

As discussed in https://github.com/hashicorp/vault-secrets-operator/pull/170/files#r1184851528


tomhjp (approved):
LGTM!

benashz (approved):
LGTM!

PR #211, opened 2023-05-24
Extend Client.Init() tests

benashz:
The Client.Init() tests now provide more coverage and cover the original regression.


kschoche (approved):
LGTM!
super-minor nit that you DONT need to do, but `"default"` everywhere could be a const (throughout the whole codebase tbh)

benashz:
> LGTM! super-minor nit that you DONT need to do, but `"default"` everywhere could be a const (throughout the whole codebase tbh)

Fixed with https://github.com/hashicorp/vault-secrets-operator/pull/211/commits/624b0c76316584b64f1107edebeef2977a3ac475

PR #217, opened 2023-05-26
Use the kubernetes_default_service_account resource for vaultauthmethod test

kschoche:
This is a follow-on change to #208 which introduced it's own serviceaccount to ensure that we could get a valid JWT token for authentication in the auth method tests.
Instead of creating our own serviceaccount we will use `kubernetes_default_service_account` resources which provide a guarantee that the default serviceaccount for a namespace does exist.

### DIFF
--- a/test/integration/vaultauthmethods/terraform/main.tf
+++ b/test/integration/vaultauthmethods/terraform/main.tf
@@ -36,19 +36,18 @@ resource "kubernetes_namespace" "tenant-1" {
   }
 }
 
-resource "kubernetes_service_account" "test-sa" {
+resource "kubernetes_default_service_account" "default-sa" {
   metadata {
-    name      = var.test_service_account
     namespace = kubernetes_namespace.tenant-1.metadata[0].name
   }
 }
 
-resource "kubernetes_secret" "test-sa-secret" {
+resource "kubernetes_secret" "default-sa-secret" {
   metadata {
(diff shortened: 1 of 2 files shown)


test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -36,19 +36,18 @@ resource "kubernetes_namespace" "tenant-1" {
   }
 }
 
-resource "kubernetes_service_account" "test-sa" {
+resource "kubernetes_default_service_account" "default-sa" {
```
benashz, line 39: nit: the resource name also includes the resource type, suggest just calling it `default`

test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -36,19 +36,18 @@ resource "kubernetes_namespace" "tenant-1" {
...
   metadata {
-    name      = var.test_service_account
     namespace = kubernetes_namespace.tenant-1.metadata[0].name
   }
 }
 
-resource "kubernetes_secret" "test-sa-secret" {
+resource "kubernetes_secret" "default-sa-secret" {
```
benashz, line 45: nit: similar comment as for `kubernetes_default_service_account` above.

benashz (approved):
a couple of minor nits, otherwise 👍

PR #226, opened 2023-05-30
Updates transit auth methods to support additional auth methods

kschoche:
* Breaking Change: Adds support for additional Auth Methods for the transit auth method template. In order to support this the existing transit auth method configuration in `values.yaml` is uses the VaultAuthMethod spec more closely in structure.
* Create a helper function to unify vault auth method templating in helm chart.
* To migrate, set kubernetes specific auth method configuration under `controller.manager.clientCache.storageEncryption` via the new stanza `controller.manager.clientCache.storageEncryption.kubernetes`.
* Adds new sections to `values.yaml` for the above, fixes a typo in values.yaml and aligns a chunk of values.yaml on 100 characters.

### DIFF
--- a/chart/templates/default-transit-auth-method.yaml
+++ b/chart/templates/default-transit-auth-method.yaml
@@ -20,38 +20,8 @@ spec:
   namespace: {{ .Values.controller.manager.clientCache.storageEncryption.namespace }}
   method: {{ .Values.controller.manager.clientCache.storageEncryption.method }}
   mount: {{ .Values.controller.manager.clientCache.storageEncryption.mount }}
-  {{- if eq .Values.controller.manager.clientCache.storageEncryption.method "kubernetes" }}
-  kubernetes:
-    role: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.role }}
-    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.serviceAccount | default (printf "%s-controller-manager" (include "chart.fullname"
-     . )) }}
-    {{- if .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences }}
-    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences | toJson }}
-    {{- end }}
-  {{- else if eq .Values.controller.manager.clientCache.storageEncryption.method "jwt" }}
-  jwt:
-    role: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.role }}
-    {{- if .Values.controller.manager.clientCache.storageEncryption.jwt.secretRef }}
-    secretRef: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.secretRef }}
-    {{- else if .Values.controller.manager.clientCache.storageEncryption.jwt.serviceAccount }}
-    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.serviceAccount }}
-    {{- if .Values.controller.manager.clientCache.storageEncryption.jwt.tokenAudiences }}
-    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.tokenAudiences | toJson }}
-    {{- end }}
-    {{- end }}
-  {{- else if eq .Values.controller.manager.clientCache.storageEncryption.method "appRole" }}
-  appRole:
-    roleId: {{ .Values.controller.manager.clientCache.storageEncryption.appRole.roleid }}
-    secretRef: {{ .Values.controller.manager.clientCache.storageEncryption.appRole.secretRef }}
-  {{- end }}
-  {{- if .Values.controller.manager.clientCache.storageEncryption.headers }}
-  headers:
-    {{ tpl .Values.controller.manager.clientCache.storageEncryption.headers . | trim }}
-  {{- end }}
-  {{- if .Values.controller.manager.clientCache.storageEncryption.params }}
-  params:
-    {{ tpl .Values.controller.manager.clientCache.storageEncryption.params . | trim }}
-  {{- end }}
+  {{- $kubeServiceAccount := .Values.controller.manager.clientCache.storageEncryption.kubernetes.serviceAccount | default (printf "%s-controller-manager" (include "chart.fullname" .)) -}}
+  {{- include "operator.vaultAuthMethod" (list .Values.controller.manager.clientCache.storageEncryption $kubeServiceAccount . ) }}
   storageEncryption:
     keyName: {{ .Values.controller.manager.clientCache.storageEncryption.keyName }}
     mount: {{ .Values.controller.manager.clientCache.storageEncryption.transitMount }}
--- a/chart/templates/default-transit-auth-method.yaml
+++ b/chart/templates/default-transit-auth-method.yaml
@@ -18,14 +18,40 @@ metadata:
 spec:
   vaultConnectionRef: {{ .Values.controller.manager.clientCache.storageEncryption.vaultConnectionRef }}
   namespace: {{ .Values.controller.manager.clientCache.storageEncryption.namespace }}
-  method: "kubernetes"
+  method: {{ .Values.controller.manager.clientCache.storageEncryption.method }}
   mount: {{ .Values.controller.manager.clientCache.storageEncryption.mount }}
+  {{- if eq .Values.controller.manager.clientCache.storageEncryption.method "kubernetes" }}
   kubernetes:
-    role: {{ .Values.controller.manager.clientCache.storageEncryption.role }}
-    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.serviceAccount | default (printf "%s-controller-manager" (include "chart.fullname" . )) }}
-    {{- if .Values.controller.manager.clientCache.storageEncryption.tokenAudiences }}
-    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.tokenAudiences | toJson }}
+    role: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.role }}
+    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.serviceAccount | default (printf "%s-controller-manager" (include "chart.fullname"
+     . )) }}
+    {{- if .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences }}
+    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences | toJson }}
     {{- end }}
+  {{- else if eq .Values.controller.manager.clientCache.storageEncryption.method "jwt" }}
+  jwt:
+    role: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.role }}
+    {{- if .Values.controller.manager.clientCache.storageEncryption.jwt.secretRef }}
+    secretRef: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.secretRef }}
+    {{- else if .Values.controller.manager.clientCache.storageEncryption.jwt.serviceAccount }}
+    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.serviceAccount }}
+    {{- if .Values.controller.manager.clientCache.storageEncryption.jwt.tokenAudiences }}
+    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.jwt.tokenAudiences | toJson }}
+    {{- end }}
+    {{- end }}
+  {{- else if eq .Values.controller.manager.clientCache.storageEncryption.method "appRole" }}
+  appRole:
+    roleId: {{ .Values.controller.manager.clientCache.storageEncryption.appRole.roleid }}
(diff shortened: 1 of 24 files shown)


chart/templates/default-transit-auth-method.yaml
```diff
@@ -18,14 +18,40 @@ metadata:
...
-    {{- if .Values.controller.manager.clientCache.storageEncryption.tokenAudiences }}
-    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.tokenAudiences | toJson }}
+    role: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.role }}
+    serviceAccount: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.serviceAccount | default (printf "%s-controller-manager" (include "chart.fullname"
+     . )) }}
+    {{- if .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences }}
+    audiences: {{ .Values.controller.manager.clientCache.storageEncryption.kubernetes.tokenAudiences | toJson }}
     {{- end }}
+  {{- else if eq .Values.controller.manager.clientCache.storageEncryption.method "jwt" }}
+  jwt:
```
benashz, line 32: At first glance this configuration seems pretty similar to what we have in `default-vault-auth-methods.yaml`. I wonder if there could be a way to share the rendering tasks between the two files?

chart/values.yaml
```diff
@@ -69,46 +69,117 @@ controller:
...
+        appRole:
+          # AppRole Role's RoleID to use for authenticating to Vault.
+          # This is a required field when using appRole and must be setup in Vault prior to deploying the
+          # helm chart.
+          # @type: string
+          roleId: ""
+          # Name of Kubernetes Secret that has the AppRole Role's SecretID used to authenticate with Vault.
+          # The Kubernetes Secret must contain a key named `id` which references the AppRole Role's
+          # SecretID, and must exist in the namespace of any consuming VaultSecret CR.
+          # This is a required field when using appRole and must be setup in Vault prior to deploying the
```
benashz, line 163: nit: mind folding long lines around 100 columns.

chart/values.yaml
```diff
@@ -69,46 +69,117 @@ controller:
...
+          # @type: string
+          secretName: ""
+
+        # Params to use when authenticating to Vault using this auth method.
+        # params: |
+        #   "vault-something1": "foo"
+        #   "vault-something2": "bar"
+        #   "vault-something3": "baz"
+        # @type: string
+        params: ""
```
benashz, line 174: would it be possible to make these a map instead of a multiline string? Ideally we could update all others of this type.

chart/values.yaml
```diff
@@ -69,46 +69,117 @@ controller:
...
 
-        # Mount path for the Transit Method.
+        # Headers to be included in all Vault requests.
+        # headers: |
+        #   "vault-something1": "foo"
+        #   "vault-something2": "bar"
+        #   "vault-something3": "baz"
         # @type: string
-        transitMount: ""
+        headers: ""
```
benashz, line 182: would it be possible to make these a map instead of a multiline string?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
-        --set 'controller.manager.clientCache.storageEncryption.role=foo-role' \
-        --set 'controller.manager.clientCache.storageEncryption.tokenAudiences={vault,foo}' \
-        --set 'controller.manager.clientCache.storageEncryption.keyName=foo-keyName' \
-        --set 'controller.manager.clientCache.storageEncryption.transitMount=foo-transit-mount' \
-        --set 'controller.manager.clientCache.storageEncryption.vaultConnectionRef=foo' \
+        --set 'controller.manager.clientCache.storageEncryption.enabled=true' \
+        . | tee /dev/stderr)
+
+    local actual=$(echo "$object" | yq '.metadata.name' | tee /dev/stderr)
+     [ "${actual}" = "release-name-vault-secrets-operator-default-transit-auth" ]
```
benashz, line 61: super nit: seems like we have some extra whitespace here, would it be possible to reformat it to line up with the line above? Typically shell uses 2 space indentation.

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
-     [ "${actual}" = "foo-sa" ]
+     [ "${actual}" = "tenant-1" ]
     actual=$(echo "$object" | yq '.spec.kubernetes.audiences' | tee /dev/stderr)
      [ "${actual}" = '["vault", "foo"]' ]
-    actual=$(echo "$object" | yq '.spec.storageEncryption.keyName' | tee /dev/stderr)
-     [ "${actual}" = "foo-keyName" ]
-    actual=$(echo "$object" | yq '.spec.storageEncryption.mount' | tee /dev/stderr)
-     [ "${actual}" = "foo-transit-mount" ]
-    actual=$(echo "$object" | yq '.spec.vaultConnectionRef' | tee /dev/stderr)
+    actual=$(echo "$object" | yq '.spec.headers.foo' | tee /dev/stderr)
```
benashz, line 103: should we check the length of `spec.headers` first?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
+     [ "${actual}" = "bar" ]
+    actual=$(echo "$object" | yq '.spec.params.foo' | tee /dev/stderr)
```
benashz, line 105: should we check the length of `spec.params` first?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
+     [ "${actual}" = "jwt" ]
+    actual=$(echo "$object" | yq '.spec.mount' | tee /dev/stderr)
+     [ "${actual}" = "foo" ]
+    actual=$(echo "$object" | yq '.spec.jwt.role' | tee /dev/stderr)
+     [ "${actual}" = "role-1" ]
+    actual=$(echo "$object" | yq '.spec.jwt.serviceAccount' | tee /dev/stderr)
+     [ "${actual}" = "tenant-1" ]
+    actual=$(echo "$object" | yq '.spec.jwt.audiences' | tee /dev/stderr)
+     [ "${actual}" = '["vault", "foo"]' ]
+    actual=$(echo "$object" | yq '.spec.headers.foo' | tee /dev/stderr)
```
benashz, line 169: should we check the length of `spec.headers` first?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
+     [ "${actual}" = "bar" ]
+    actual=$(echo "$object" | yq '.spec.params.foo' | tee /dev/stderr)
```
benashz, line 171: should we check the length of `spec.params` first?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
+     [ "${actual}" = "jwt" ]
+    actual=$(echo "$object" | yq '.spec.mount' | tee /dev/stderr)
+     [ "${actual}" = "foo" ]
+    actual=$(echo "$object" | yq '.spec.jwt.role' | tee /dev/stderr)
+     [ "${actual}" = "role-1" ]
+    actual=$(echo "$object" | yq '.spec.jwt.secretRef' | tee /dev/stderr)
+     [ "${actual}" = "secret-1" ]
+    actual=$(echo "$object" | yq '.spec.headers.foo' | tee /dev/stderr)
+     [ "${actual}" = "bar" ]
+    actual=$(echo "$object" | yq '.spec.params.foo' | tee /dev/stderr)
```
benashz, line 209: should we check the length of `spec.params` first?

test/unit/default-transit-auth-method.bats
```diff
@@ -49,40 +49,197 @@ load _helpers
...
+
+    actual=$(echo "$object" | yq '.spec.method' | tee /dev/stderr)
+     [ "${actual}" = "jwt" ]
+    actual=$(echo "$object" | yq '.spec.mount' | tee /dev/stderr)
+     [ "${actual}" = "foo" ]
+    actual=$(echo "$object" | yq '.spec.jwt.role' | tee /dev/stderr)
+     [ "${actual}" = "role-1" ]
+    actual=$(echo "$object" | yq '.spec.jwt.secretRef' | tee /dev/stderr)
+     [ "${actual}" = "secret-1" ]
+    actual=$(echo "$object" | yq '.spec.headers.foo' | tee /dev/stderr)
```
benashz, line 207: should we check the length of `spec.headers` first?

benashz (commented):
I had few comment/nits, otherwise looking good!

chart/templates/default-transit-auth-method.yaml
kschoche: Good point, I was planning on creating a function for it, but got stuck on the templating logic, fixed in 18a15e7

test/unit/default-transit-auth-method.bats
kschoche: I'm not 100% sure it's necessary or what it would test, aside from a builtin helm function `tpl`, in the current form.
 
For each of these, if `spec.params/headers` is undefined we simply get `null` from yq and the check fails without crashing anything, here is an example:
```
@test "defaultTransitAuthMethod/CR: service account and token audiences settings can be modified for jwt auth method" {
    cd `chart_dir`
    local object=$(helm template \
        -s templates/default-transit-auth-method.yaml  \
        --set 'controller.manager.clientCache.storageEncryption.enabled=true' \
        --set 'controller.manager.clientCache.persistenceModel=direct-encrypted' \
        --set 'controller.manager.clientCache.storageEncryption.params=foo: baz' \  # <-- note no header was provided
        . | tee /dev/stderr)

     actual=$(echo "$object" | yq '.spec.headers.foo' | tee /dev/stderr)
      [ "${actual}" = "bar" ]
    actual=$(echo "$object" | yq '.spec.params.foo' | tee /dev/stderr)
      [ "${actual}" = "baz" ]
}

 ✗ defaultTransitAuthMethod/CR: test with nil h auth method
   (in test file test/unit/default-transit-auth-method.bats, line 148)
     `[ "${actual}" = "bar" ]' failed
   ---
   # Source: vault-secrets-operator/templates/default-transit-auth-method.yaml
   kind: VaultAuth
   metadata:
     name: release-name-vault-secrets-operator-default-transit-auth
     namespace: default
     labels:
       control-plane: controller-manager
       component: controller-manager
       cacheStorageEncryption: "true"
       helm.sh/chart: vault-secrets-operator-0.1.0-beta.1
       app.kubernetes.io/name: vault-secrets-operator
       app.kubernetes.io/instance: release-name
       app.kubernetes.io/version: "0.1.0-beta.1"
   spec:
     vaultConnectionRef: default
     namespace:
     method: kubernetes
     mount: kubernetes
     kubernetes:
       role:
       serviceAccount: release-name-vault-secrets-operator-controller-manager
     params:
       foo: baz
     storageEncryption:
       keyName:
       mount:
       null  <-------- the test prints out the actual value for headers test on `spec.headers.foo` which failed here
```

chart/values.yaml
kschoche: Sure thing.. the tests still pass though so now I'm worried :)

chart/values.yaml
```diff
@@ -69,46 +69,119 @@ controller:
...
+          # Name of Kubernetes Secret that has the AppRole Role's SecretID used to authenticate with
+          # Vault. The Kubernetes Secret must contain a key named `id` which references the AppRole
+          # Role's SecretID, and must exist in the namespace of any consuming VaultSecret CR.
+          # This is a required field when using appRole and must be setup in Vault prior to
+          # deploying the helm chart.
+          # @type: string
+          secretName: ""
+
+        # Params to use when authenticating to Vault using this auth method.
+        # params: |
```
benashz, line 171: The example seems to suggest multi-line string.

chart/values.yaml
```diff
@@ -69,46 +69,119 @@ controller:
...
+        #   "vault-something1": "foo"
+        #   "vault-something2": "bar"
+        #   "vault-something3": "baz"
+        # @type: map
+        params: {}
+
+        # Headers to be included in all Vault requests.
+        # headers: |
```
benashz, line 179: https://hashicorp.atlassian.net/browse/VAULT-11606

test/unit/default-transit-auth-method.bats
```diff
@@ -27,15 +27,15 @@ load _helpers
...
+@test "defaultTransitAuthMethod/CR: kubernetes default serviceaccount uses operator sa as a default" {
     cd `chart_dir`
     local object=$(helm template \
         -s templates/default-transit-auth-method.yaml  \
         --set 'controller.manager.clientCache.persistenceModel=direct-encrypted' \
         . | tee /dev/stderr)
 
     actual=$(echo "$object" | yq '.spec.kubernetes.serviceAccount' | tee /dev/stderr)
-     [ "${actual}" = "release-name-vault-secrets-operator-controller-manager" ]
+      [ "${actual}" = "release-name-vault-secrets-operator-controller-manager" ]
```
benashz, line 38: nit: would it be possible to shift this over to line up with the line above? I think there are few more below like that as well.

benashz (approved):
Only a few minor nits, then :+1:

PR #227, opened 2023-05-30
add prometheus ServiceMonitor chart template

thyton:


chart/values.yaml
```diff
@@ -303,6 +303,41 @@ defaultAuthMethod:
...
+    #  release: prometheus
+    selectors: {}
+
+    # Scheme of the service Prometheus scrape metrics from. This must match the scheme of the metrics service of VSO
+    scheme: https
+
+    # Port at which Prometheus scrapes metrics. This must match the port of the metrics service of VSO
+    port: 8443
+
+    # Path at which Prometheus scraps metrics
```
benashz, line 332:
```suggestion
    # Path at which Prometheus scrapes metrics
```

chart/templates/prometheus-servicemonitor.yaml
```diff
@@ -0,0 +1,42 @@
+{{/*
```
benashz, line 1: I wonder if we could base the contents of this off of https://github.com/hashicorp/vault-secrets-operator/blob/ab5766f4ebbde57e9b2cbae518fe8f7985ece94d/config/prometheus/monitor.yaml#L1-L0 - that would align more closely with a kustomize based deployment

chart/values.yaml
thyton: Thanks!

test/unit/prometheus-servicemonitor.bats
```diff
@@ -0,0 +1,118 @@
...
+  cd `chart_dir`
+  local output=$( (helm template \
+      --show-only templates/prometheus-servicemonitor.yaml \
+      --set 'telemetry.serviceMonitor.enabled=true' \
+      --set 'telemetry.serviceMonitor.selectors.baz=qux' \
+      --set 'telemetry.serviceMonitor.selectors.bar=foo' \
+      . ) | tee /dev/stderr)
+
+  [ "$(echo "$output" | yq -r '.metadata.labels | has("app")')" = "false" ]
+  [ "$(echo "$output" | yq -r '.metadata.labels.baz')" = "qux" ]
```
benashz, line 87: not sure if it would be helpful, but you might want to consider checking the length of `.metadata.labels`

benashz (approved):
Looks good to me! Only one minor suggestion.

Issue #187, opened 2023-05-18
cannot use static ldap secrets, only dynamic


ninjadude333:
**Describe the bug**
trying to fetch a static ldap secret, a rotated password for an existing ldap user, this is not working, dynamic ldap secrets do work.

**To Reproduce**
```
vault write ldap/static-role/hashicorp \
    dn='uid=hashicorp,ou=People,dc=ninjadude,dc=com' \
    username='hashicorp' \
    rotation_period="6h"


---
apiVersion: secrets.hashicorp.com/v1alpha1
kind: VaultDynamicSecret
metadata:
  namespace: default
  name: example
spec:
  vaultAuthRef: example
  mount: ldap   **(Also tried ldap/static-role)**
  role: hashicorp
  destination:
    create: true
    name: dynamic1


**Expected behavior**
the ldap credential should be fetched.

**Environment**
* EKS 1.24
vault ent 
vault-secrets-operator version: latest

benashz:
@ninjadude333 I believe we have fixed this in https://github.com/hashicorp/vault-secrets-operator/pull/172 (unreleased). We have replace `role` with `path` so that you can specify any `path` relative to the secret engine's mount.

ninjadude333:
great thanks. is there an eta for release?

benashz:
We don't have an exact date yet. It should be just after [v0.1.0-beta.1](https://github.com/hashicorp/vault-secrets-operator/milestone/1) is closed.

benashz:
Fixed in #172

PR #229, opened 2023-05-30
VDS: ensure that any update results in a complete sync.

benashz:
Previously, the VDS controller would ignore changes to the CR being reconciled, this would happen if the associated secret lease ID was renewable. This PR extends the VDS Status to support tracking the generation of the last synced CR. If the CR that is currently being reconciled has a generation other the last synced CR then a full sync will occur.


api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -58,8 +58,10 @@ type VaultDynamicSecretSpec struct {
 
 // VaultDynamicSecretStatus defines the observed state of VaultDynamicSecret
 type VaultDynamicSecretStatus struct {
-	// LastRenewalTime of the last, successful, secret lease renewal,
+	// LastRenewalTime of the last successful secret lease renewal,
```
kschoche, line 61: Looks like you missed one extra comma.. Reminder to run `make generate manifests` after this fix, I'll skip comments related to this during the review.

```suggestion
	// LastRenewalTime of the last successful secret lease renewal.
```

controllers/vaultdynamicsecret_controller.go
```diff
@@ -101,9 +101,13 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 	}
 
-	var doRolloutRestart bool
+	// forceSync indicates that the controller should perform the secret sync,
```
kschoche, line 104: minor nit:  `forceSync` seems to imply that we're doing this when we otherwise wouldn't, now that that is the expected behaviour, I'm wondering if `doSync` is a bit more descriptive?

controllers/vaultdynamicsecret_controller.go
```diff
@@ -183,15 +187,17 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	o.Status.LastGeneration = o.GetGeneration()
 	if err := r.updateStatus(ctx, o); err != nil {
 		return ctrl.Result{}, err
 	}
 
 	reason := consts.ReasonSecretSynced
+	var horizon time.Duration
 	leaseDuration := time.Duration(secretLease.LeaseDuration) * time.Second
-	horizon := computeDynamicHorizonWithJitter(leaseDuration, o.Spec.RenewalPercent)
+	horizon = computeDynamicHorizonWithJitter(leaseDuration, o.Spec.RenewalPercent)
```
kschoche, line 198: super minor nit: I don't think this change is necessary?

controllers/vaultdynamicsecret_controller.go
```diff
@@ -201,16 +207,18 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}
 
 	if !r.isRenewableLease(secretLease, o) {
+		logger.Info("Secret lease is not renewable, will not requeue for renewal",
```
kschoche, line 210: Great job adding this logging!

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -339,14 +342,19 @@ func TestVaultDynamicSecret(t *testing.T) {
 					))
 					require.Equal(t, 1, len(pods.Items))
 					assert.Equal(t, pods.Items[0].UID, vdsObjFinal.Status.LastRuntimePodUID)
+					assert.Equal(t, vdsObjFinal.GetGeneration(), vdsObjFinal.Status.LastGeneration)
```
kschoche, line 345: I think this indirectly is a duplicate of line 331 since we are not doing any `Get()` of the `vdsObjFinal` between line 331 and here, so these will always be true if line 331 is true.

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -371,6 +379,32 @@ func TestVaultDynamicSecret(t *testing.T) {
...
+// assertDynamicSecretNewGeneration tests that an update to vdsObjOrig results in a full secret rotation.
+func assertDynamicSecretNewGeneration(t *testing.T, ctx context.Context, client ctrlclient.Client, vdsObjOrig *secretsv1alpha1.VaultDynamicSecret) {
+	t.Helper()
+
+	objKey := ctrlclient.ObjectKeyFromObject(vdsObjOrig)
+	vdsObjLatest := &secretsv1alpha1.VaultDynamicSecret{}
+	if assert.NoError(t, client.Get(ctx, objKey, vdsObjLatest)) {
+		vdsObjLatest.Spec.Destination.Name += "-new"
+		if assert.NoError(t, client.Update(ctx, vdsObjLatest)) {
+			assertDynamicSecretRotation(t, ctx, client, vdsObjLatest)
```
kschoche, line 391: I'm just trying to follow this, we issue a `client.Update()` on the vdsObj, changing it's destination secret, which should trigger a reconciliation + creation of the new k8s secret.
But the revocation of the lease as part of `assertDynamicSecretRotation` is actually what triggers the rotation right?

It seems to me that either a lease revocation or the changing of the destination.name should trigger this (and a rollout restart)?
Am I confusing myself?

kschoche (commented):
I left a couple small nits and had a question about the testing at the end, but otherwise I think this is looking great!

controllers/vaultdynamicsecret_controller.go
benashz: Think that was leftover from original patch for #187

kschoche (approved):
Excellent, this looks great!

PR #233, opened 2023-05-31
rename name to role in VaulPKISpec

thyton:


api/v1alpha1/vaultpkisecret_types.go
```diff
@@ -25,8 +25,8 @@ type VaultPKISecretSpec struct {
 	// Mount for the secret in Vault
 	Mount string `json:"mount"`
 
-	// Name of the secret in Vault
-	Name string `json:"name"`
+	// Role in Vault to use for issuing credentials
```
benashz, line 28:
```suggestion
	// Role in Vault to use when issuing TLS certificates.
```

benashz (commented):
Minor nit, otherwise it's looks good. Please update the CHANGELOG as well, especially since this is a breaking change.

Also, I think we need to the stuff under `config/samples`, there may be some other places in the repo where we are still using the old value.

benashz (approved):
LGTM!

PR #234, opened 2023-06-01
Ensure any previously created k8s Secrets are cleaned up after sync.

benashz:
Prior to this fix, whenever `Spec.Destination.Name` changed the previous k8s Secret would remain after the sync. This PR ensures that any previously created k8s secrets that are owned by the Syncable Secret object will be deleted, and only the newly synced secret remains.

### DIFF
--- a/internal/helpers/secrets.go
+++ b/internal/helpers/secrets.go
@@ -22,6 +22,11 @@ import (
 	"github.com/hashicorp/vault-secrets-operator/internal/consts"
 )
 
+// labelOwnerRefUID is used as the primary key when listing the Secrets owned by
+// a specific VSO object. It should be included in every Secret that is created
+// by VSO.
+var labelOwnerRefUID = fmt.Sprintf("%s/vso/ownerRefUID", secretsv1alpha1.GroupVersion.Group)
+
 // OwnerLabels will be applied to any k8s secret we create. They are used in Secret ownership checks.
 // There are similar labels in the vault package. It's important that component secret's value never
 // intersects with that of other components of the system, since this could lead to data loss.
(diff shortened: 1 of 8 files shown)


internal/helpers/secrets.go
```diff
@@ -169,11 +262,38 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 
 	if exists {
 		logger.V(consts.LogLevelDebug).Info("Updating secret")
-		return client.Update(ctx, &dest)
+		if err := client.Update(ctx, &dest); err != nil {
```
kschoche, line 265: One thing that is really powerful about the Kubernetes API's is our ability to use Label filtering and push that computation out to the kube api server.

In `FindSecretsOwnedByObj` we're doing a `client.List()` with `OwnerLabels` but that will still return every Kubernetes secret (in this namespace) that is managed by VSO.
If we add an extra "Owner label" on line 252 that directly references the VSO Secret we can dramatically reduce the work needed in `FindSecretsOwnedByObj` and have `client.List()` return *only* the Kube Secrets that we want to interact with. I imagine that there could be thousands of Kube secrets managed in a single ns and we'd be iterating through that list of 1000's of secrets every time this code triggers otherwise. (Assuming I'm reading the code correctly?)

 
I'm thinking something like this (again Line 252):
```
	// always add the "owner" labels last to guard against intersections with meta.Destination.Labels
	for k, v := range OwnerLabels {
		labels[k] = v
	}
        labels["secrets.hashicorp.com/ownerRef"] = obj.GetName()
```
..and then add it to the OwnerLables again in `FindSecretsOwnedByObj` when you're about to issue the `client.List()`?

Thoughts?

benashz: > One thing that is really powerful about the Kubernetes API's is our ability to use Label filtering and push that computation out to the kube api server.

Indeed. I think we are already benefiting from that feature with our use of `OwnerLabels`

> 
> In `FindSecretsOwnedByObj` we're doing a `client.List()` with `OwnerLabels` but that will still return every Kubernetes secret (in this namespace) that is managed by VSO. If we add an extra "Owner label" on line 252 that directly references the VSO Secret we can dramatically reduce the work needed in `FindSecretsOwnedByObj` and have `client.List()` return _only_ the Kube Secrets that we want to interact with. I imagine that there could be thousands of Kube secrets managed in a single ns and we'd be iterating through that list of 1000's of secrets every time this code triggers otherwise. (Assuming I'm reading the code correctly?)
> 
> I'm thinking something like this (again Line 252):
> 
> ```
> 	// always add the "owner" labels last to guard against intersections with meta.Destination.Labels
> 	for k, v := range OwnerLabels {
> 		labels[k] = v
> 	}
>         labels["secrets.hashicorp.com/ownerRef"] = obj.GetName()
> ```
> 
> ..and then add it to the OwnerLables again in `FindSecretsOwnedByObj` when you're about to issue the `client.List()`?
> 
> Thoughts?

Sounds like a pretty nice optimization, although that means we would miss the secrets that do not have that label. Presumably the next time they are sync'd it would be added.

One issue I see with the proposal to use the object's name is that we would end up with orphans whenever the CR's name was updated. We could probably mitigate that issue if we use the object's UID instead of its Name.

benashz: added in https://github.com/hashicorp/vault-secrets-operator/pull/234/commits/fcc25d8cb3765c9d5e1e9165d5f724d79e16caea

internal/helpers/secrets.go
```diff
@@ -89,6 +200,10 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 		Name:      meta.Destination.Name,
 	}
 
+	if err := common.ValidateObjectKey(key); err != nil {
+		return fmt.Errorf("empty name for destination Secret, err=%w", err)
```
kschoche, line 204: Ni: I believe `ValidateObjectKey` will return `err` if either name/namespaces are empty string, suggest changing the wording a bit.
```suggestion
		return fmt.Errorf("empty name or namespace for destination Secret, err=%w", err)
```

Right now this will output something like:
`empty name (or namespace) for destination Secret, err=invalid objectKey` which maybe seems a bit redundant, but change it how you see fit, if you see fit :)

internal/helpers/secrets.go
```diff
@@ -112,7 +240,13 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 		// syncable-secret's Status, but...
 		dest.Data = data
 		logger.V(consts.LogLevelDebug).Info("Updating secret")
-		return client.Update(ctx, &dest)
+		if err := client.Update(ctx, &dest); err != nil {
+			return err
```
kschoche, line 244: Nit: Does it make sense to add some logging here around SyncSecret returning error? I looked at the places where `SyncSecret()` is called and it looks like the error gets returned directly to Reconcile, so it's not entirely lost. 🤷🏽

kschoche (approved):
Awesome work, I *really* like the unit tests, great job!

kschoche (approved):
Awesome work, I *really* like the unit tests, great job!

kschoche (approved):
Awesome work, I *really* like the unit tests, great job!

internal/helpers/secrets.go
benashz: Hmm, we probably want to make the message more generic, and ensure that the wrapped error provides the relevant bits?

benashz: Fixed with d9e218884470d5ff0394f56b84657325ed443131

internal/helpers/secrets.go
benashz: Yeah, I think it might be okay to let the caller log it. We can add it later if it becomes an issue.

PR #235, opened 2023-06-02
VaultAuth: AWS auth support

tvoran:
Adds AWS auth support for [`iam` type authentication](https://developer.hashicorp.com/vault/docs/auth/aws#iam-auth-method). AWS credentials can be specified as static, inferred from the EKS node role, inferred from the EKS instance profile, or from an IRSA-enabled k8s service account.

To use an IRSA service account, the name must be specified in the `VaultAuth.Spec.AWS.IRSAServiceAccount` field, and that service account must have an `eks.amazonaws.com/role-arn` annotation with an IRSA assumable role. A token requested from k8s for that service account (with the default audience of `sts.amazonaws.com`) can then be used to assume the specified AWS role and auth to Vault.

Depends on https://github.com/hashicorp/go-secure-stdlib/pull/78

Includes VaultAuth integration tests to run in EKS for various configurations of AWS auth. Example AWS auth test run on EKS:

```shell
$ make -f aws.mk integration-test-eks TESTARGS="-count 1 -v -run TestVaultAuth" SKIP_AWS_STATIC_CREDS_TEST=false TF_VAR_aws_access_key_id=... TF_VAR_aws_secret_access_key="..." TF_VAR_aws_static_creds_role="arn:aws:iam::..."

...
--- PASS: TestVaultAuthMethods (79.92s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-kubernetes (6.96s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-jwt-serviceaccount (1.40s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-jwt-secret (0.33s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-approle (0.32s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-aws-irsa (1.41s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-aws-node (1.41s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-aws-instance-profile (1.44s)
    --- PASS: TestVaultAuthMethods/vaultauth-test-aws-static (1.45s)
PASS
ok  	github.com/hashicorp/vault-secrets-operator/test/integration	80.431s
```

Note that the static creds test is skipped by default in CI when on EKS, and all the AWS auth tests are skipped when not in EKS.


internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+
+var _ CredentialProvider = (*AWSCredentialProvider)(nil)
+
+type AWSCredentialProvider struct {
+	authObj           *secretsv1alpha1.VaultAuth
+	providerNamespace string
+	uid               types.UID
+}
+
+func (l *AWSCredentialProvider) GetNamespace() string {
```
tomhjp, line 37: Doesn't matter, but OOI why `l`?

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+		// there, use the secret UID as l.uid
+		key := ctrlclient.ObjectKey{
+			Namespace: l.providerNamespace,
+			Name:      l.authObj.Spec.AWS.AWSCredsRef,
+		}
+		credsSecret, err := getSecret(ctx, client, key)
+		if err != nil {
+			return err
+		}
+		l.uid = credsSecret.UID
```
tomhjp, line 60: It's maybe not best practice, but I think there might be some scenarios where people have multiple VaultAuths pointing at the same static secret or service account, but using different Vault roles. In that case, the UID of those auth providers won't be unique. Should we generate our own UUID to avoid that?

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+			Namespace: l.providerNamespace,
+			Name:      l.authObj.Spec.AWS.AWSCredsRef,
+		}
+		credsSecret, err = getSecret(ctx, client, key)
+		if err != nil {
+			logger.Error(err, "Failed to get secret", "secret_name", l.authObj.Spec.AWS.AWSCredsRef)
+			return nil, err
+		}
+	}
+	if l.authObj.Spec.AWS.IRSAServiceAccount != "" {
```
tomhjp, line 100: nit: To match the logic in `Init` and avoid any potential errors, which I think would be irrelevant?

```suggestion
	} else if l.authObj.Spec.AWS.IRSAServiceAccount != "" {
```

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+		key := ctrlclient.ObjectKey{
+			Namespace: l.providerNamespace,
+			Name:      l.authObj.Spec.AWS.IRSAServiceAccount,
+		}
+		irsaServiceAccount, err := getSA(ctx, client, key)
+		if err != nil {
+			logger.Error(err, "Failed to get IRSA service account", "service_account", l.authObj.Spec.AWS.IRSAServiceAccount)
```
tomhjp, line 107: How come the error is both logged and returned in a few places? It looks like `err` gets logged a few levels up so I wonder if we should skip logging and instead imbue it with the additional context?

tomhjp (approved):
Nice tests! I left a few comments but none are major or blocking

internal/vault/credentials/aws.go
benashz: > It's maybe not best practice, but I think there might be some scenarios where people have multiple VaultAuths pointing at the same static secret or service account, but using different Vault roles. 

I think that should be fine, although as you stated maybe a little odd.

> In that case, the UID of those auth providers won't be unique. Should we generate our own UUID to avoid that?
I am not seeing the issue with that since in the end the Vault Client will be unique for each VaultAuth. Could you maybe unpack that a bit more.  What issues do you foresee, etc?

For the generation bit, the contract is that all credential providers should provide their own UUID (UID) that persists for the lifetime of that provider. We haven't encountered a situation where VSO has ever had to generate one on behalf of a credential provider.

README.md
```diff
@@ -143,6 +143,14 @@ make -f aws.mk build-push integration-test-eks
 make -f aws.mk build-push integration-test-eks VAULT_ENTERPRISE=true ENT_TESTS=true
 ```
 
+#### AWS auth test with static credentials
+
+```shell
+# Set SKIP_AWS_STATIC_CREDS_TEST=false and specify the terraform variables aws_access_key_id,
+# aws_secret_access_key, and aws_static_creds_role ARN for a user in AWS.
+make -f aws.mk integration-test-eks TESTARGS="-run TestVaultAuth" SKIP_AWS_STATIC_CREDS_TEST=false TF_VAR_aws_access_key_id=... TF_VAR_aws_secret_access_key="..." TF_VAR_aws_static_creds_role="arn:aws:iam::..."
```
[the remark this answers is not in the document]
benashz: Setting the vars on the command line is good option. I don't think we should have to set TF_ prefixed env vars at all.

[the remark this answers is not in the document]
benashz: Personally I like writing the env vars, especially ones like this to a temporary tfvars file, that way they do not leak. I think we are doing something like that for the vault license.

test/integration/vaultauthmethods/terraform/variables.tf
```diff
@@ -83,3 +83,41 @@ variable "operator_image_tag" {
...
+  default     = ""
+}
+
+variable "aws_secret_access_key" {
+  description = "AWS_SECRET_ACCESS_KEY for testing static creds with AWS auth"
+  default     = ""
+  sensitive   = true
+}
+
+variable "aws_static_creds_role" {
```
gautambaghel, line 110: Is there any way we could create the required variables here in the infra/eks folder or perhaps create-eks make cmd?

Trying to reduce some of the pre-requisites before running the integration-test-eks

README.md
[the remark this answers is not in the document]
gautambaghel: true, although some tests have aws cli cmds + terraform so "export vars" might be easier to work with.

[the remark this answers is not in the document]
benashz: true, better not to create our own special variables then.

internal/vault/credentials/aws.go
tomhjp: I'm not sure if it's a problem. But generally I would expect each distinct object to have a unique UID, so if there are multiple providers with the same UID, it feels like it could lead to surprising bugs. For example, UIDs are used to generate cache keys so UID collisions could in theory cause cache key collisions. And to be clear, I don't think that specific example _is_ a problem right now, my question is more generally "it looks like provider UID collisions can happen, are we ok with that?"

wrt "Should we generate our own UUID to avoid that?" - I don't really mind what solution we pick, but that was a possible suggestion we could copy what we do just below in the case that we don't have static or IRSA creds: https://github.com/hashicorp/vault-secrets-operator/blob/be159c35ceb97e477e3c244876688a960062e31c/internal/vault/credentials/aws.go#L73-L77

benashz: Given the current design, we should definitely not be generating new credential provider UIDs.

test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -169,6 +169,81 @@ path "auth/${vault_auth_backend.approle.path}/login" {
 EOT
 }
 
+# aws auth config
+resource "kubernetes_service_account" "irsa_assumable" {
+  metadata {
```
kschoche, line 174: Super nit, but I think you can put a `count` on this resource too?
```suggestion
    count = var.run_aws_tests || run_aws_static_creds_test? 1 : 0
    metadata {
```

api/v1alpha1/vaultauth_types.go
```diff
@@ -56,6 +56,37 @@ type VaultAuthConfigAppRole struct {
...
+	// The STS endpoint to use; if not set will use the default
+	STSEndpoint string `json:"stsEndpoint,omitempty"`
+
+	// The IAM endpoint to use; if not set will use the default
+	IAMEndpoint string `json:"iamEndpoint,omitempty"`
+
+	// AWSCredsRef is the name of a Kubernetes Secret which holds credentials
+	// for AWS. Expected keys include `access_key_id`, `secret_access_key`,
+	// `session_token`
+	AWSCredsRef string `json:"awsCredsRef,omitempty"`
```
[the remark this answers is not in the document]
benashz: :+1:

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		irsaServiceAccount, err := getSA(ctx, client, key)
+		if err != nil {
+			return err
+		}
+		l.uid = irsaServiceAccount.UID
+	} else {
+		// At this point either the node role or the instance profile will be
+		// used for credentials, so just generate a new UID
+		l.uid = uuid.NewUUID()
```
benashz, line 76: The only issue I see with this is that the UUID change when VSO is restarted, that will make it impossible for the client to ever restored from the storage cache. See https://github.com/hashicorp/vault-secrets-operator/pull/155#discussion_r1183097733 for a bit more  context.

Do either the node role or instance profile provide any unique ID we could use here?

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, err
+	}
+
+	headerValue := l.authObj.Spec.AWS.HeaderValue
+
+	loginData, err := awsutil.GenerateLoginData(creds, headerValue, config.Region, config.Logger)
+	if err != nil {
+		return nil, err
+	}
+	loginData["role"] = l.authObj.Spec.AWS.Role
```
benashz, line 140: May want to blow up early if `Spec.AWS.Role` is empty, adding some suspenders to the API validation belt...

internal/vault/credentials/helpers.go
```diff
@@ -47,3 +47,15 @@ func getSecret(ctx context.Context, client ctrlclient.Client, key ctrlclient.Obj
 
 	return secret, nil
 }
+
+func getSA(ctx context.Context, client ctrlclient.Client, saKey ctrlclient.ObjectKey) (*corev1.ServiceAccount, error) {
```
benashz, line 51: nit: can we rename `saKey` to `key` like in `getSecret()` or  `objKey` is also an option for both functions.

test/integration/vaultauthmethods/terraform/variables.tf
```diff
@@ -83,3 +83,41 @@ variable "operator_image_tag" {
...
+  description = "AWS role ARN for the static creds"
+  default     = ""
+}
```
benashz, line 113: Seems like this  should be required, unless we had a `datasource` that could fetch the value for us.

benashz, line 113: Should probably be required.

benashz, line 113: Should probably required as well.

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -190,22 +302,30 @@ func TestVaultAuthMethods(t *testing.T) {
 
 	// Apply all the Auth Methods
 	for _, a := range auths {
-		require.Nil(t, crdClient.Create(ctx, a))
-		created = append(created, a)
+		if run, why := a.shouldRun(); !run {
```
benashz, line 305: cute :+1:

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -256,3 +379,34 @@ func TestVaultAuthMethods(t *testing.T) {
...
+	}
+	if v := os.Getenv(envSkipAWSStaticCreds); v == "true" {
+		return false, envSkipAWSStaticCreds + " is set to 'true'"
+	}
+	tfVars := []string{
+		"TF_VAR_aws_access_key_id",
+		"TF_VAR_aws_secret_access_key",
+		"TF_VAR_aws_static_creds_role",
+	}
+	for _, tfv := range tfVars {
```
benashz, line 406: I wonder if we should fail the test since the required env vars are not set.

README.md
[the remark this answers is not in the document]
tvoran: The idea here was to use static credentials that are different (and hopefully lower privilege) than the creds used to provision the cluster and the tests. So I'd rather not rely on the usual AWS_* env vars that would be picked up by the aws sdk.

internal/vault/credentials/aws.go
tvoran: Ok, I think that should work.

internal/vault/credentials/aws.go
tvoran: I don't really know; it's that way in the other providers too.

internal/vault/credentials/aws.go
tvoran: Hmm, I don't think so, since we wouldn't know which of the node role or instance profile to use here since that's configured on the Vault role side. Plus those would be the same for any VaultAuth in the cluster. 

Would a UUID constructed from a hash of the VaultAuth spec plus the k8s name and namespace suffice?

api/v1alpha1/vaultauth_types.go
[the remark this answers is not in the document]
tvoran: Sure, changed in b2680f6

api/v1alpha1/vaultauth_types.go
```diff
@@ -56,6 +56,37 @@ type VaultAuthConfigAppRole struct {
 	SecretRef string `json:"secretRef"`
 }
 
+// VaultAuthConfigAWS provides VaultAuth configuration options needed for
+// authenticating to Vault via an AWS AuthMethod
```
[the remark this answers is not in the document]
tvoran: Yeah that makes sense, added something similar in b2680f6

test/integration/vaultauthmethods/terraform/main.tf
tvoran: Yeah, I think run_aws_tests should suffice since the static creds test doesn't use irsa. Updated in 1383a3b

internal/vault/credentials/aws.go
tvoran: Gave that a shot in 348a5aa

benashz: Hmm, I don't think that will work in the case of the "default" VaultAuth. I think we would end up with the same hash for all tenants. I wonder if we really need to support this credential source, since it seems quite global? Unless I am missing something here?

internal/vault/credentials/aws.go
```diff
@@ -0,0 +1,220 @@
...
+		config.RoleARN = v
+	}
+
+	// Override defaults from any annotations set
+	if v, ok := annotations[AWSAnnotationAudience]; ok {
+		config.Audience = v
+	}
+	if v, ok := annotations[AWSAnnotationTokenExpiration]; ok {
+		check, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
```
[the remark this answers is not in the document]
tvoran: Ok, changed in 378cf98

internal/vault/credentials/aws.go
tvoran: For the support question, it's just with the aim of supporting as many of the aws auth options as we can.

test/integration/vaultauthmethods_integration_test.go
tvoran: Sure, that makes sense.

README.md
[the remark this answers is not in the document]
tvoran: I reworked this a bit in 20ab235 to set the static creds through environment variables (instead of just tf_var's).

internal/vault/credentials/aws.go
tvoran: Since this case is cluster-wide, would setting a static UID work here? i.e. `l.uid = types.UID("01234567-89ab-cdef-01234567")`

Otherwise I guess we just drop this case for now? Though it seems like it may be useful in the default VaultAuth case (i.e. the default operator auth uses the cluster's node role, then users configure their own for IRSA, etc.)

tvoran: Or instead of hard-coding a UID, we could use the `kube-root-ca.crt` configmap UID from the operator's namespace as something "cluster-wide". Or there's an `aws-auth` configmap in the `kube-system` namespace that is specific to EKS clusters, though I'm a little hesitant to read something from kube-system.

aws.mk
```diff
@@ -27,18 +27,20 @@ create-eks: ## Create a new EKS cluster
...
 -include $(TF_EKS_STATE_DIR)/outputs.env
 
-# Currently only supports amd64
 .PHONY: build-push
 build-push: import-aws-vars ci-build ci-docker-build ## Build the operator image and push it to the GAR repository
 	aws ecr get-login-password --region $(AWS_REGION) | docker login --username AWS --password-stdin $(ECR_URL)
 	docker push $(IMG)
 
 .PHONY: integration-test-eks
+integration-test-eks: export SKIP_AWS_TESTS=false
```
benashz, line 36: nit: any reason we couldn't pass these to the `$(MAKE) integration-test` below?

benashz (approved):
Looks good. I only had few nits/suggestions then :+1:

kschoche (approved):
Awesome work!

gautambaghel (approved):
Thanks for your work on this! 🙏🏾

aws.mk
tvoran: Yeah it's setup this way to essentially override the "defaults" set in `Makefile`; I had trouble getting that to work when passing them directly to the make command.

benashz: ok, no worries then.

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -240,17 +367,73 @@ func TestVaultAuthMethods(t *testing.T) {
...
+		return false, envSkipAWSStaticCreds + " is set to 'true'"
+	}
+	tfVars := []string{
+		"TEST_AWS_ACCESS_KEY_ID",
+		"TEST_AWS_SECRET_ACCESS_KEY",
+		"AWS_STATIC_CREDS_ROLE",
+	}
+	for _, tfv := range tfVars {
+		if v := os.Getenv(tfv); v == "" {
+			t.Fail()
```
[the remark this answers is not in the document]
tvoran: Indeed I think you're right, though it was failing the parent `t`, while the subtest `t`'s were still reflecting the correct PASS/FAIL status.

I split up the checks into `shouldRun()` and `canRun()` to be a little more accurate there in d761a91.

test/integration/vaultauthmethods_integration_test.go
```diff
@@ -240,17 +367,73 @@ func TestVaultAuthMethods(t *testing.T) {
...
 
+	logEvents := func(t *testing.T, vss *secretsv1alpha1.VaultStaticSecret) {
+		eventList := &corev1.EventList{}
+		listOptions := &client.ListOptions{
+			Namespace:     vss.Namespace,
+			FieldSelector: fields.OneTermEqualSelector("involvedObject.name", vss.Name),
+		}
+		if err := crdClient.List(ctx, eventList, listOptions); err != nil {
+			t.Logf("event list error: %q", err)
+			t.Fail()
```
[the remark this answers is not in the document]
tvoran: `assert.Fail()` would probably make sense, though since it's just collecting events for debugging purposes, leaving the `.Fail()` out might make sense here too.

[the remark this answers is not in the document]
benashz: Yeah, I wasn't sure what the intent was here. Maybe something like. If failure to list events should fail the tests then, maybe do `assert.NoError(err)` after the log call, that way the test can continue on its merry way.

tvoran:
Thanks all!

PR #239, opened 2023-06-05
VDS: add support for static-creds

### LINKED ISSUE #187: cannot use static ldap secrets, only dynamic

benashz:
This PR extends VDS to support synchronizing "static-creds" from a Vault dynamic secrets engine. The two engines known to support static-reds are [database](https://developer.hashicorp.com/vault/api-docs/secret/databases#get-static-credentials), and [ldap](https://developer.hashicorp.com/vault/api-docs/secret/ldap#static-role-passwords).

The VDS CRD was extended to include the following new configuration option:
```
	// AllowStaticCreds should be set when syncing credentials that are periodically
	// rotated by the Vault server, rather than created upon request. These secrets
	// are sometimes referred to as "static roles", or "static credentials", with a
	// request path that contains "static-creds" or "static-cred", or ....
```

When `StaticCreds` is set to to true, VSO will ignore not follow the lease renewal flow. Instead, VSO will honour the automatic rotation period of the static creds.  The TTL is tracked in the Status' new `VaultStaticCredsMetaData` structure.

```
type VaultStaticCredsMetaData struct {
	// LastVaultRotation represents the last time Vault rotated the password
	LastVaultRotation int64 `json:"lastVaultRotation"`
	// RotationPeriod is number in seconds between each rotation, effectively a
	// "time to live". This value is compared to the LastVaultRotation to
	// determine if a password needs to be rotated
	RotationPeriod int64 `json:"rotationPeriod"`
	// TTL is the seconds remaining before the next rotation.
	TTL int64 `json:"ttl"`
}
```

Secret drift detection will be handled in subsequent PR.

Closes #187


api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -46,6 +46,11 @@ type VaultDynamicSecretSpec struct {
 	RenewalPercent int `json:"renewalPercent,omitempty"`
 	// Revoke the existing lease on VDS resource deletion.
 	Revoke bool `json:"revoke,omitempty"`
+	// StaticCreds should be set when syncing credentials that are periodically
+	// rotated by the Vault server, rather than created upon request. These secrets
+	// are sometimes referred to as "static roles", or "static credentials", with a
+	// request path that contains "static-creds".
+	StaticCreds bool `json:"staticCreds,omitempty"`
```
tvoran, line 53: Would it be possible to determine if the creds are static just by inspecting the Vault response?

kschoche: I was thinking this as well.
It looks like for DB there isnt anything in the (sample) response in the api-docs that screams out as "static-cred". We can parse the path `/database/static-creds/`, but that doesn't seem future-proof.

benashz: Yeah, that might be possible. I hesitated on that originally, since I was not sure that we could guarantee that response is really for "static-creds". It's possible we could come up with some heuristic that would work most of the time, but we would have to provide an escape hatch for the user in case VSO gets it wrong.

We could do something like:
```
isStaticCreds = leaseNotRenewable && some-static-creds-keys in resp.Data
```

I sort of prefer the more declarative approach that we have now, but I could definitely be persuaded otherwise.

kschoche: I think there is definitely enough in the response to figure this out (at the time of getting a credential, which may be too late in the logic of Reconcile):

Example response for dynamic:
```
{
  "data": {
    "username": "root-1430158508-126",
    "password": "132ae3ef-5a64-7499-351e-bfe59f3a2a21"
  }
}
```

and static:
```
{
  "data": {
    "username": "static-user",
    "password": "132ae3ef-5a64-7499-351e-bfe59f3a2a21",
    "last_vault_rotation": "2019-05-06T15:26:42.525302-05:00",
    "rotation_period": 30,
    "ttl": 28
  }
}
```

But I do agree that it feels hacky to rely on the response alone, and a more declarative approach feels cleaner and might make the code easier to read.

benashz: Updated with 41e9143257679ed973a6d9ecf9f7fbe68bc41494

api/v1alpha1/vaultdynamicsecret_types.go
```diff
@@ -64,6 +69,8 @@ type VaultDynamicSecretStatus struct {
 	LastGeneration int64 `json:"lastGeneration"`
 	// SecretLease for the Vault secret.
 	SecretLease VaultSecretLease `json:"secretLease"`
+	// StaticCredsMetaData contains the static creds response meta-data
+	StaticCredsMetaData VaultStaticCredsMetaData `json:"staticMeta,omitempty"`
```
kschoche, line 73: Nit: maybe makes sense to have the json be closer to the name to be descriptive? `staticCredsMeta`?

```suggestion
	StaticCredsMetaData VaultStaticCredsMetaData `json:"staticCredsMeta,omitempty"`
```

benashz: Fixed with 5d7596d16fb034a28f70aca377d804331a0bd301

controllers/vaultdynamicsecret_controller.go
```diff
@@ -205,9 +232,8 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 		_ = helpers.HandleRolloutRestarts(ctx, r.Client, o, r.Recorder)
 	}
 
-	if !r.isRenewableLease(secretLease, o) {
-		logger.Info("Secret lease is not renewable, will not requeue for renewal",
-			"lease_id", secretLease.ID)
+	if (!r.isRenewableLease(secretLease, o) && !o.Spec.AllowStaticCreds) || horizon.Seconds() == 0 {
+		// no need to requeue
```
kschoche, line 236: I believe this will result in an Event of type Warning from `isRenewableLease()` being emitted even though we are not requeueing or returning an error, that seems a little off to me?

test/integration/vaultdynamicsecret/terraform/postgres.tf
```diff
@@ -69,13 +92,29 @@ resource "vault_database_secret_backend_role" "postgres" {
...
+resource "vault_database_secret_backend_static_role" "postgres" {
+  namespace           = local.namespace
+  backend             = vault_database_secrets_mount.db.path
+  name                = local.db_role_static
+  db_name             = vault_database_secrets_mount.db.postgresql[0].name
+  username            = local.db_role_static_user
+  rotation_statements = ["ALTER USER \"{{name}}\" WITH PASSWORD '{{password}}';"]
+  rotation_period     = 30
+  depends_on = [
+    null_resource.create-pg-user,
```
kschoche, line 104: this looks odd, I guess I'm learning something new today :)
Is this primarily due to a lack of being able to create a static role in tf for the db engine?

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -327,14 +386,19 @@ func TestVaultDynamicSecret(t *testing.T) {
...
+					if vdsObjFinal.Spec.AllowStaticCreds {
+						assert.Empty(t, vdsObjFinal.Status.SecretLease.ID)
+					} else {
+						assert.NotEmpty(t, vdsObjFinal.Status.SecretLease.ID)
+					}
+
 					assertLastRuntimePodUID(t, ctx, crdClient, operatorNS, vdsObjFinal)
-					assertDynamicSecretRotation(t, ctx, crdClient, vdsObjFinal)
+					if !vdsObjFinal.Spec.AllowStaticCreds {
+						assertDynamicSecretRotation(t, ctx, crdClient, vdsObjFinal)
```
kschoche, line 397: Silly question, maybe my brain is off today: I don't see a test for Rotation for static creds?
I see that we are checking for the existence of TTL/etc, but maybe it makes sense to test a scenario where the `lastRotated' value has changed?

kschoche (approved):
Looks great, I had a few smallish comments around testing and some logging, but otherwise it looks good!

test/integration/vaultdynamicsecret/terraform/postgres.tf
benashz: A pre-existing PG "user" must be configured before we can provision the vault static role. The use of a `depends_on` is needed since a `null_resource` has no result that could be used by the dependent resource.

test/integration/vaultdynamicsecret_integration_test.go
benashz: Yeah, makes sense. Adding it.

benashz: Updated in e530736f78ad9e7a296cc53d2b0b347a2bfd1dce

controllers/vaultdynamicsecret_controller.go
benashz: Updated in e530736f78ad9e7a296cc53d2b0b347a2bfd1dce

kschoche (approved):
⭐

tvoran (approved):

ninjadude333:
"""When StaticCreds is set to to true, VSO will ignore not follow the lease renewal flow. Instead, VSO will honor the automatic rotation period of the static creds. """

hi, great job, just a quick question,
what happens if i manually trigger a password rotation for the ldap-static role, i think you should add some kind of manually defined check interval, that if configured will override the rotation period lookup.
something like: refreshAfter: 5m , that is already available in the static secret, or if you have any other way to listen for changes, this use case may also be needed, currently i just need it for demo purposes. so I can change the secret and see the change.

10x

PR #240, opened 2023-06-06
replace Name with Path in VaultStaticSecretSpec

maithyton:


api/v1alpha1/vaultstaticsecret_types.go
```diff
@@ -20,8 +20,8 @@ type VaultStaticSecretSpec struct {
 	Namespace string `json:"namespace,omitempty"`
 	// Mount for the secret in Vault
 	Mount string `json:"mount"`
-	// Name of the secret in Vault
-	Name string `json:"name"`
+	// Path of the secret in Vault
```
tomhjp, line 23: There might be reasons against I'm unaware of, but I would love us to be more explicit about what exactly this corresponds to, to avoid confusion.

```suggestion
	// Path of the secret in Vault, corresponds to the `path` parameter for,
	// kv-v1: https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v1#read-secret
	// kv-v2: https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#read-secret-version
```

tomhjp (approved):
LGTM, just a documentation suggestion

api/v1alpha1/vaultstaticsecret_types.go
thyton: Thank you!

kschoche (approved):
Awesome!

PR #247, opened 2023-06-07
chart: adding aws to defaultAuth and storageEncryption auth

tvoran:
Adding support for aws auth in the `defaultAuth` and `storageEncryption` (transit) sections of the chart, along with tests.


benashz (approved):
LGTM!

PR #248, opened 2023-06-07
update to go 1.20.5

tvoran:
go 1.20.4 => 1.20.5

Go 1.20.5 contains security fixes for the runtime and compiler. Details are here: https://github.com/golang/go/issues?q=milestone%3AGo1.20.5+label%3ACherryPickApproved

Not sure we need it for the RC but we'll probably want it for GA.


benashz:
> LGTM - one integration test failed which I am sure isn't related, so I re-ran it to check it's a flake.

Looks like it was the new static-creds test that failed. We'll look into the logs.

PR #249, opened 2023-06-07
chart: make defaultVaultConnection.headers a map

benashz:
Additional fixes:
- add support for filtering bats tests
- simplify some of the header/params examples


kschoche (approved):
Looks good!

PR #256, opened 2023-06-09
VaultPKISecret: include the CA chain (sans root) in 'tls.crt'

tvoran:
When the destination type is "kubernetes.io/tls", construct `tls.crt` as `certificate` + `ca_chain` from the Vault issue response. (`issuing_ca` is used when `ca_chain` is empty.) Uses the `remove_roots_from_chain=true` option with Vault to exclude the root CA from the Vault response. The `remove_roots_from_chain` option was adding in Vault 1.12.

Tested using the [Build Your Own Certificate Authority](https://developer.hashicorp.com/vault/tutorials/secrets-management/pki-engine) Vault tutorial and nginx ingress.


kschoche (approved):
This looks great!
Ran it locally and ✅

PR #266, opened 2023-06-13
Adds controller.imagePullSecrets to helm chart

### LINKED ISSUE #265: [Helm chart] Add global.imagePullSecrets parameter
At the moment, i must pull my images from a private registry, unfortunately within templates and values, no option is avalaible to set imagePullSecrets within deployment and jobs.

Within templates files, add the imagePullSecrets option to be set in the values file. (E.g. : make a global.imagePullSecrets that will set this parameter for any resources that needs it.

How i bypass that now is just patching the deployment, but if i try to uninstall the helm chart, i basically have issues with a job that can't pull the image.

How i do : kubectl patch deployment <deployment-name> -p '{"spec":{"template":{"spec":{"imagePullSecrets":[{"name":"<secret-name>"}]}}}}'

kschoche:
* Adds `controller.imagePullSecrets` to the helm chart which will apply imagePullSecrets to the controller serviceaccount so all containers in the controller's deployment will have access to private registry credentials.

From the Kube API docs for ServiceAccounts:
```
	// ImagePullSecrets is a list of references to secrets in the same namespace to use for pulling any images
	// in pods that reference this ServiceAccount. ImagePullSecrets are distinct from Secrets because Secrets
	// can be mounted in the pod, but ImagePullSecrets are only accessed by the kubelet.
	// More info: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
```
More info: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secret-to-service-account

Fixes #265


chart/templates/deployment.yaml
```diff
@@ -10,6 +10,12 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
   {{- include "chart.labels" . | nindent 4 }}
+{{- with .Values.controller.imagePullSecrets }}
+imagePullSecrets:
+{{- range . }}
+  - name: {{ .name }}
+{{- end }}
+{{- end }}
```
tvoran, line 18: We may want to consider also supporting just a list of just the secret names as well, similar to https://github.com/hashicorp/vault-helm/pull/576/files

i.e.
```yaml
controller:
  imagePullSecrets:
    - pull-secret-name
    - pull-secret-name-2
```

tvoran (approved):

chart/templates/deployment.yaml
kschoche: Good call, I did see that is how Vault-helm did things, but was on the fence about it, definitely can standardize on it!

kschoche: Fixed in 59822c9b352bacd25200fcbffc3f3620d462a3b8

tvoran (approved):

test/unit/deployment.bats
```diff
@@ -220,3 +220,52 @@ load _helpers
...
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/deployment.yaml  \
+      --set 'controller.imagePullSecrets[0].name=foo' \
+      --set 'controller.imagePullSecrets[1].name=bar' \
+      . | tee /dev/stderr |
+      yq '. | select(documentIndex == 0)' | tee /dev/stderr)
+
+  local actual=$(echo "$object" |
+     yq -r '.imagePullSecrets[0].name' | tee /dev/stderr)
```
benashz, line 266: We should probably check the length of `imagePullSecrets`, as well as the length of its elements.

benashz (approved):
Only one minor update to the tests, then 👍

PR #279, opened 2023-07-01
Sync rendered chart with deployed resource

freimer:
The chart specifies the skipTLSVerify value even if the value is the default of false.  This causes the rendered yaml to be different from the yaml reported when you read the resource from kubernetes (via say kubectl get vaultconnection/default -oyaml). This causes problems for orchestrators like Rancher as it sees this as a modification.  This simple chagne leaves out the value if it is the "default" - only including it if it is non-default value of true, so the rendered manifest and reported yaml match.


freimer:
What's the status of this?  Says tests are required, but this is already covered with https://github.com/hashicorp/vault-secrets-operator/blob/main/test/unit/default-vault-connection.bats

freimer:
@kschoche can you kick that off again if necessary?  I modified the default test to make sure the default value (false) is not in the spec as it is a default, and added another test to check to make sure it is also not in the spec if it is specified but set to the default value of false.  There's already a test to check if it is specified with the non-default value of true.  So that should cover all possibilities.  Thanks!

freimer:
OK @kschoche all tests pass locally now.  I recreated the "issue" with the test and then fixed it.  Sorry about that!

freimer:
Let me know if there is anything else I can help with to move this forward.  Thanks again!

benashz:
Closing in favor of #527

PR #280, opened 2023-07-03
set container resources on the cleanup job

### LINKED ISSUE #277: Be able to define resorce limits for all K8s resources.
In our K8s environment, we have a company-wide policy that enforces resource limits on K8s containers. When installing VSO in our environment, the installation fails because the `vault-secrets-operator-pre-delete-controller-cleanup` does not define resources limits. 

I'm already using the `controller.manager.resources` settings in the `values.yaml`, however those values are not being applied to the container above.

Either change the `controller.manager.resources` to also be applied to all other VSO resources, or expose new settings to define the limits per resource.

None so far. The only solution seems to be to ask internally to open an exception, which is undesirable. 

Thanks.

kschoche:
Sets container resource limits on the job for the deployment template to be the same as the VSO container.

Fixes #277

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -147,10 +147,11 @@ spec:
       containers:
       - name: pre-delete-controller-cleanup
         image: {{ .Values.controller.manager.image.repository }}:{{ .Values.controller.manager.image.tag }}
-        args:
-        - --finalizer-cleanup=true
         command:
         - /vault-secrets-operator
+        args:
+        - --finalizer-cleanup=true
+        resources: {{- toYaml .Values.controller.manager.resources | nindent 10 }}
       restartPolicy: Never
       {{- if .Values.controller.nodeSelector }}
       nodeSelector:
--- a/test/unit/deployment.bats
+++ b/test/unit/deployment.bats
@@ -82,6 +82,23 @@ load _helpers
     [ "${actual}" = "128Mi" ]
 }
 
+@test "controller/Deployment: default resources for job" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/deployment.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].resources | select(documentIndex == 2)' | tee /dev/stderr)
+
+   local actual=$(echo "$object" | yq '.requests.cpu' | tee /dev/stderr)
+    [ "${actual}" = "10m" ]
+   actual=$(echo "$object" | yq '.requests.memory' | tee /dev/stderr)
+    [ "${actual}" = "64Mi" ]
+   actual=$(echo "$object" | yq '.limits.cpu' | tee /dev/stderr)
+    [ "${actual}" = "500m" ]
+   actual=$(echo "$object" | yq '.limits.memory' | tee /dev/stderr)
+    [ "${actual}" = "128Mi" ]
+}
+
 @test "controller/Deployment: can set resources for controller" {
   cd `chart_dir`
   local object=$(helm template \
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 Improvements:
 * Helm: `controller.imagePullSecrets` stanza is added to provide imagePullSecrets to the controller's containers via the serviceAccount: [GH-266](https://github.com/hashicorp/vault-secrets-operator/pull/266)
+* Helm: `controller.manager.resources` values now also apply to the pre-delete-controller-cleanup-job. [GH-2R8066](https://github.com/hashicorp/vault-secrets-operator/pull/280)
```
tvoran, line 5: What does `GH-2R8066` mean?

chart/templates/deployment.yaml
```diff
@@ -147,10 +147,11 @@ spec:
       containers:
       - name: pre-delete-controller-cleanup
         image: {{ .Values.controller.manager.image.repository }}:{{ .Values.controller.manager.image.tag }}
-        args:
-        - --finalizer-cleanup=true
         command:
         - /vault-secrets-operator
+        args:
+        - --finalizer-cleanup=true
```
tvoran, line 153: Why isn't `--finalizer-cleanup=true` set in the kustomize version?

test/unit/deployment.bats
```diff
@@ -82,6 +82,23 @@ load _helpers
     [ "${actual}" = "128Mi" ]
 }
 
+@test "controller/Deployment: default resources for job" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/deployment.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].resources | select(documentIndex == 2)' | tee /dev/stderr)
```
tvoran, line 90: Just a thought to make this a bit more robust:

```suggestion
      yq '.spec.template.spec.containers[] | select(.name == "pre-delete-controller-cleanup") | .resources' | tee /dev/stderr)
```

test/unit/deployment.bats
```diff
@@ -91,15 +108,26 @@ load _helpers
       --set 'controller.manager.resources.limits.memory=200Mi' \
       --set 'controller.manager.resources.limits.cpu=200m' \
       . | tee /dev/stderr |
-      yq '.spec.template.spec.containers[1].resources | select(documentIndex == 1)' | tee /dev/stderr)
+      yq '.' | tee /dev/stderr)
 
-   local actual=$(echo "$object" | yq '.requests.cpu' | tee /dev/stderr)
+   local controller=$(echo "$object" | yq '.spec.template.spec.containers[1].resources | select(documentIndex == 1)' | tee /dev/stderr)
+   local job=$(echo "$object" | yq '.spec.template.spec.containers[0].resources | select(documentIndex == 2)' | tee /dev/stderr)
```
tvoran, line 114: Similar thought here:

```suggestion
   local controller=$(echo "$object" | yq '.spec.template.spec.containers[] | select(.name == "manager") | .resources' | tee /dev/stderr)
   local job=$(echo "$object" | yq '.spec.template.spec.containers[] | select(.name == "pre-delete-controller-cleanup") | .resources' | tee /dev/stderr)
```

test/unit/deployment.bats
```diff
@@ -82,6 +82,23 @@ load _helpers
     [ "${actual}" = "128Mi" ]
 }
 
+@test "controller/Deployment: default resources for job" {
```
tvoran, line 85: I wonder if this should be combined with the above "default resources for controller" test, like how the "can set resources for controller" checks both job and controller.

tvoran (approved):
Left a couple thoughts, but 👍

chart/templates/deployment.yaml
kschoche: This job is a helm hook which gets applied to the cluster on helm delete, there isn't a real concept of that for kustomize AFAIK. 
The original reason we created the job was because helm installs were not cleaning up cleanly in ci due to us deploying default auth and connection resources via helm. 
Kustomize doesn't do this so it doesn't need to run within CI.
That being said I think this hook is getting a slight refractor in the PR that Thy is working on.

CHANGELOG.md
kschoche: Typo good catch :) Should be GH-280

kschoche: Fixed in 21c5dc2

test/unit/deployment.bats
kschoche: Fixed in 21c5dc2

PR #288, opened 2023-07-20
Adding UBI image builds

tvoran:
Adds an image based on [ubi-micro](https://catalog.redhat.com/software/containers/ubi9-micro/61832b36dd607bfc82e66399) to the Dockerfile, build directives in the Makefile, and image build jobs in the CI `build` job for both dockerhub and redhat (quay.io).

~Also adds s390x and ppc64le to the arch list since they are supported by both the distroless and ubi-micro base images. Since we don't have a good way to test these I'm thinking they'd be supported similar to [etcd's Tier 3](https://etcd.io/docs/v3.5/op-guide/supported-platform/#supporting-a-new-platform) (i.e. Build Passes).~ Edit: removed s390x and ppc64le builds for now.

The integration tests passed locally with the ubi image.


.github/dependabot.yaml
```diff
@@ -21,4 +21,4 @@ updates:
   - package-ecosystem: "docker"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
tvoran, line 24: Bumped this to "daily" just so we catch any UBI updates as soon as we can, to minimize the risk of rejection at release time.

.github/workflows/build.yaml
```diff
@@ -145,6 +143,83 @@ jobs:
...
+      - name: Docker Build (Action)
+        uses: hashicorp/actions-docker-build@v1
+        env:
+          VERSION: ${{ needs.get-product-version.outputs.product-version }}
+          GO_VERSION: ${{ needs.build-pre-checks.outputs.go-version }}
+        with:
+          version: ${{env.version}}
+          target: release-ubi
+          arch: ${{matrix.arch}}
+          redhat_tag: quay.io/redhat-isv-containers/64b072322e2773c28d30d988:${{env.image_tag}}
```
benashz, line 209: Might be worth documenting where the hash/ID is coming from. I am assuming this is our repo on `quay.io`? Is the ID using during authentication, should it be considered confidential info?

benashz (approved):
LGTM!

Dockerfile
```diff
@@ -66,6 +66,35 @@ USER 65532:65532
...
+
+ENV BIN_NAME=vault-secrets-operator
+ARG PRODUCT_VERSION
+ARG PRODUCT_REVISION
+ARG PRODUCT_NAME=$BIN_NAME
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+LABEL name="Vault Secrets Operator" \
+      maintainer="Team Vault <vault@hashicorp.com>" \
```
tomhjp, line 81: I'm not sure if vault@ is a monitored inbox?

```suggestion
      maintainer="Team Vault <team-vault@hashicorp.com>" \
```

tomhjp (approved):
LGTM!

kschoche (approved):
LGTM!

.github/workflows/build.yaml
tvoran: Yeah the hash/ID corresponds to the image project ID in the redhat portal, and it's not considered to be sensitive.

This also follows the pattern of our other UBI images, ex:
* https://github.com/hashicorp/vault/blob/main/.github/workflows/build.yml#L213
* https://github.com/hashicorp/vault-k8s/blob/main/.github/workflows/build.yml#L125

Dockerfile
tvoran: Hmm, I don't know for sure either, but `vault@hashicorp.com` is what we've been using in our other Dockerfile's, e.g.
* https://github.com/hashicorp/vault/blob/main/Dockerfile#L19
* https://github.com/hashicorp/vault/blob/main/Dockerfile#L19

tomhjp: 👍 I'm fine with vault@ as well

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+Features:
+* UBI-based images now built and published with releases: [GH-288](https://github.com/hashicorp/vault-secrets-operator/pull/288)
+* `ppc64le` and `s390x` builds are now included in releases: [GH-288](https://github.com/hashicorp/vault-secrets-operator/pull/288)
```
tvoran, line 5: I couldn't decide between Changes or Features for these, so definitely open to suggestions.

Issue #274, opened 2023-06-21
Allow setting VaultAuth spec kubernetes mount as a default global setting


mmckane:
**Is your feature request related to a problem? Please describe.**
Developers currently need to define the vault mount auth endpoint for every namespace they deploy into. This adds overhead to our development teams as we have many clusters and they now need to know the correct vault auth mount point at deployment time for each cluster they deploy to. As we treat clusters and vault role to service account mappings like cattle, this is problematic as the developer now needs to know the mount point for the cluster each time they deploy.

**Describe the solution you'd like**
Configure a way to globally set the mount so that it can be omitted from VaultAuth spec when using Kubernetes auth. Possibly allow setting this via an ENV that is passed into the chart/deployment/pod that configures the setting for all namespaces, so that developers do not need to provide the specific vault auth mountpoint for the cluster they have been deployed to. 

Moving the mount point to the VaultConnection spec might also solve this as that can be globally defaulted, where the developer then only needs to worry about creating a VaultAuth spec that contains the role and service account mapping.

kschoche:
Hi @mmckane, thanks for filing this issue and sorry that you're running into this.

It sounds to me like what you're asking for might be the default auth method? You can configure it via the helm chart under `defaultAuthMethod` and then any CR that does not specify an auth method reference will make use of the default. 

I might be missing something though, if for example you're using different mounts per k8s ns/app?

Can you let me know if that would work for you or how we could extend it otherwise? 
Thanks!

mmckane:
@kschoche  Doesn't defaultAuthMethod also require a service account and vault role to be defined that would apply to each namespace using the default Auth Method? The problem is each namespace in my cluster uses different service account names per k8s best practices and each namespace/SA binds to a different vault role. This doesn't seem like the most secure setup even if I did use the same SA/role everywhere as they would all use the same vault role meaning a tenant in namespace A could create a resource and access namespace B secrets as they would both be bound to the same role.

xenochor:
I agree with @kschoche, please review the patterns afforded by the CSI secrets store driver and Vault provider regarding enabling system-wide defaults for URL, namespace and mount path while allowing mounts within namespaces to only be concerned with secret path, role and, indirectly in the case of CSI secrets store, service account. This is one of the key failings of the other options such as the Vault secrets operator and the external secrets operator.

mmckane:
[...]

PR #291, opened 2023-07-21
add ns to vault auth and vault connection

### LINKED ISSUE #136: Ability to manage secrets across multiple kubernetes namespaces
The output kubernetes secret from a sync operation is always stored in the namespace that contains the `VaultSecret*` custom resource. 

Theres two main problems with this:
1. I am unable to use the operator to manage secrets in scenarios where the secret is expected in a specific namespace.
2. a VaultConnection and VaultAuth resource needs to be created per kubernetes namespace that has vault managed secrets

It would be good if the [Destination spec](https://developer.hashicorp.com/vault/docs/platform/k8s/vso/api-reference#destination) could support specifying the desired kubernetes namespace for the secret to be created in.

Creating `VaultConnection`, `VaultAuth` and `VaultSecret*` custom resources all in the desired kubernetes namespace works as expected but this means that you end up with duplicate Connection/Auth custom resources spread across multiple kubernetes namespaces

kschoche:
Adds the ability to set a Namespace for the following objects:
* VDS/VSS/VPKI's VaultAuthRef
* VaultAuthRef's VaultConnectionRef
* While preserving existing behaviour.
* Adds `spec.AllowedNamespaces` to VaultAuthMethod CR so that it can allow-list namespaces

This allows all VaultAuthMethod and VaultConnections to be managed in their own respective namespaces instead of being a part of the Application's namespace.

Fixes #136 and #274


internal/common/common.go
```diff
@@ -261,3 +300,49 @@ func ValidateObjectKey(key ctrlclient.ObjectKey) error {
 	}
 	return nil
 }
+
+// SyncableSecretMetaData provides common data structure that extracts the bits pertinent
```
kschoche, line 304: Moved this from helpers because of circular dependences between `helpers` and `common`

api/v1beta1/vaultauth_types.go
```diff
@@ -93,11 +93,23 @@ type VaultAuthConfigAWS struct {
...
+	// If no namespace is included the Operator will default to the namespace of the VaultAuth CR.
 	// If no value is specified the Operator will default to the `default` VaultConnection,
-	// configured in its own Kubernetes namespace.
+	// configured in its own Kubernetes namespace unless prefixed.
 	VaultConnectionRef string `json:"vaultConnectionRef,omitempty"`
 	// Namespace to auth to in Vault
 	Namespace string `json:"namespace,omitempty"`
+	// AllowedNamespaces Kubernetes Namespaces which are allow-listed for use with this AuthMethod.
+	// This field allows administrators to customize which Kubernetes namespaces are authorized to
+	// act with this AuthMethod. While Vault will still enforce its own rules, this has the added
```
benashz, line 105:
```suggestion
	// use this AuthMethod. While Vault will still enforce its own rules, this has the added
```

api/v1beta1/vaultauth_types.go
```diff
@@ -93,11 +93,23 @@ type VaultAuthConfigAWS struct {
...
+	// configurability of restricting which VaultAuthMethods can be used by which namespaces.
+	// Accepted values:
+	// []{"*"} - wildcard, all namespaces.
+	// []{"a", "b"} - list of namespaces.
+	// []{} - empty list, no namespaces.
+	// nil - all namespaces, default behavior.
```
benashz, line 111: Does this mean they can assign the value `nil`. Maybe we could say `unset or empty then the VaultAuth can only be referenced from a Vault*Secret's namespace` ?  I think that nil slice and empty slice are the same.

I wonder if we could make it so that VSO will allow the namespace if it is in `AllowedNamespaces` or `== *` ?

api/v1beta1/vaultdynamicsecret_types.go
```diff
@@ -11,6 +11,8 @@ import (
 // VaultDynamicSecretSpec defines the desired state of VaultDynamicSecret
 type VaultDynamicSecretSpec struct {
 	// VaultAuthRef to the VaultAuth resource
+	// The VaultAuthRef can be prefixed with a namespace, eg: `namespaceA/vaultRefB`.
+	// If no namespace is specified the Operator will default to namespace of the VaultAuth CR.
```
benashz, line 15: nit: this is a bit confusing since it's not possible to explicitly set the vaultAuth's namespace, wonder if we could say something like if the `namespace is included...` or if the `value is absolute`, not sure.

internal/common/common.go
```diff
@@ -41,50 +42,77 @@ func init() {
...
+
+	m, err := NewSyncableSecretMetaData(obj)
+	if err != nil {
+		return types.NamespacedName{}, types.NamespacedName{}, fmt.Errorf("unsupported type %T", obj)
 	}
 
-	var authName types.NamespacedName
-	if authRef == "" {
+	if m.AuthRef != "" {
+		// todo: some sorta sanity checking?
```
benashz, line 55: Still required? What sort of check did you have in  mind?

internal/common/common.go
```diff
@@ -41,50 +42,77 @@ func init() {
...
+		names := strings.Split(m.AuthRef, "/")
+		if len(names) > 1 {
```
benashz, line 57: Probably want to return an error if `len(names) > 2`

internal/common/common.go
```diff
@@ -41,50 +42,77 @@ func init() {
...
+}
+
+// According to the spec for secretsv1beta1.VaultAuth, Spec.AllowedNamespaces behave as follows:
+// AllowedNamespaces:
+//
+//	 nil - all namespaces are allowed
+//	[]{}  - no namespaces are allowed
+//	[]{"*"} - with length of 1, all namespaces are allowed
+//	[]{"a","b"} - explicitly namespaces a, b are allowed
+func isValidTargetNamespace(auth *secretsv1beta1.VaultAuth, name types.NamespacedName) bool {
```
benashz, line 86: nit: this seems to check the namespace is allowed, maybe consider renaming to something `isNamespaceAllowed` ?

internal/common/common.go
```diff
@@ -41,50 +42,77 @@ func init() {
...
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+	return authRef, target, nil
+}
+
+// According to the spec for secretsv1beta1.VaultAuth, Spec.AllowedNamespaces behave as follows:
+// AllowedNamespaces:
+//
+//	 nil - all namespaces are allowed
```
benashz, line 82: I am not sure I get the reason for the `nil`  that should be the same as an empty slice/array. Also, wouldn't that force the user to always set `allowedNamespaces` to `[]` on all VaultAuth instances?

internal/common/common.go
```diff
@@ -160,10 +188,21 @@ func GetConnectionNamespacedName(a *secretsv1beta1.VaultAuth) (types.NamespacedN
 		}, nil
 	}
 
-	// the VaultConnection CR must be in the same namespace as its VaultAuth.
+	// Use the NS of the AuthRef, unless it's overridden in the AuthRef Spec.
+	var vaultConnectionRefNamespace string
+	var vaultConnectionRefName string
+	vcrNames := strings.Split(a.Spec.VaultConnectionRef, "/")
+	if len(vcrNames) == 1 {
```
benashz, line 195: similar comment as for the value parsing done for `Vault*Secrets*`, maybe consider factoring this parsing bit out to a common func.

internal/common/common.go
benashz, line 86: seems like we might be missing the case where `auth.Namespace == name.Namespace`, which should always be allowed.

internal/common/common.go
benashz: :+1:

internal/common/common.go
benashz, line 86: Another case would be `auth.Namespace == OperatorNamespace && auth.Name == default`, that should always be allowed

benashz (commented):
Looking good!

internal/common/common.go
kschoche: This was an old comment, no longer needed thx!

internal/common/common.go
kschoche: Fixed in a52da2d

internal/common/common.go
kschoche: Refactored this per our conversation, should be fixed in a52da2d

internal/common/common.go
kschoche: Renamed to `AllowedNamespace`
Fixed in a52da2d

kschoche: Fixed in a52da2d

kschoche: Fixed in a52da2d

api/v1beta1/vaultauth_types.go
kschoche: Fixed in a52da2d

api/v1beta1/vaultauth_types.go
kschoche: Refactored this per our conversation.
Fixed in a52da2d

api/v1beta1/vaultdynamicsecret_types.go
kschoche: Updated the wording for this in a52da2d

test/integration/vaultstaticsecret_integration_test.go
```diff
@@ -37,6 +37,7 @@ import (
 func TestVaultStaticSecret_kv(t *testing.T) {
 	testID := strings.ToLower(random.UniqueId())
 	testK8sNamespace := "k8s-tenant-" + testID
+	testK8sNamespace2 := "k8s-tenant-" + testID + "-test"
```
benashz, line 40:
```suggestion
	testK8sNamespace2 := testK8sNamespace + "-test"
```

kschoche: Fixed in c175ed93

internal/common/common.go
kschoche: Refactored this into a common function in c175ed93, thanks for the suggestion, it's much cleaner

internal/common/common.go
```diff
@@ -41,50 +42,86 @@ func init() {
...
+	// Allow if target ns is the same as auth
+	if name.Namespace == auth.ObjectMeta.Namespace {
+		return true
+	}
+	// Default Auth Method
+	if auth.ObjectMeta.Name == consts.NameDefault && auth.ObjectMeta.Namespace == OperatorNamespace {
+		return true
+	}
+	// Disallow by default
+	if auth.Spec.AllowedNamespaces == nil || len(auth.Spec.AllowedNamespaces) == 0 {
```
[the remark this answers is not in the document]
kschoche: Fixed in c175ed93

internal/common/common.go
```diff
@@ -41,50 +42,86 @@ func init() {
...
+	return authRef, target, nil
+}
+
+// According to the spec for secretsv1beta1.VaultAuth, Spec.AllowedNamespaces behaves as follows:
+// AllowedNamespaces:
+//
+//	unset - disallow all except the OperatorNamespace and the AuthMethod's ns, default behavior.
+//	[]{"*"} - with length of 1, all namespaces are allowed
+//	[]{"a","b"} - explicitly namespaces a, b are allowed
+func AllowedNamespace(auth *secretsv1beta1.VaultAuth, name types.NamespacedName) bool {
```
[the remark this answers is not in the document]
kschoche: Fixed in c175ed93

internal/common/common.go
```diff
@@ -41,50 +42,86 @@ func init() {
...
+			authRef.Name = names[1]
+		} else {
+			authRef.Namespace = obj.GetNamespace()
+			authRef.Name = names[0]
+		}
+	} else {
 		// if no authRef configured we try and grab the 'default' from the
 		// Operator's current namespace.
-		authName = types.NamespacedName{
+		authRef = types.NamespacedName{
```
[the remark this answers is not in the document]
kschoche: Fixed in c175ed93

internal/common/common.go
```diff
@@ -41,50 +42,86 @@ func init() {
...
+	if err != nil {
+		return types.NamespacedName{}, types.NamespacedName{}, fmt.Errorf("unsupported type %T", obj)
 	}
 
-	var authName types.NamespacedName
-	if authRef == "" {
+	if m.AuthRef != "" {
+		names := strings.Split(m.AuthRef, "/")
+		if len(names) > 2 {
+			return types.NamespacedName{}, types.NamespacedName{}, fmt.Errorf("invalid auth method name %s", m.AuthRef)
```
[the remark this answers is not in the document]
kschoche: fixed in c175ed93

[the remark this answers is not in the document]
kschoche: Fixed in c175ed93

benashz (approved):
LGTM!

umutkacar:
How can I define these **AllowedNamespaces** in the values.yaml on VSO chart?

PR #299, opened 2023-07-28
Ensure client cache keys are always lower cased

benashz:
Since the Client cache key is encoded in the K8s Secret name it needs to follow the K8s specification for resource names, i.e. it can only contain lowercase characters.

To reproduce the issue:
- VSO running with persistent client cache enable
- Deploy a VaultAuth with the `appRole` method specified

### DIFF
--- a/internal/vault/cache_key.go
+++ b/internal/vault/cache_key.go
@@ -128,7 +128,7 @@ func computeClientCacheKey(authObj *secretsv1beta1.VaultAuth, connObj *secretsv1
 		connObj.GetUID(), connObj.GetGeneration(), providerUID)
 
 	sum := sha256.Sum256([]byte(input))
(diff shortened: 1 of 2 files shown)


thyton (approved):

kschoche (approved):
Looks good, this seems like a healthy change.
IIUC, this will have the undesired effect of leaving all previous clients (that use capital letters) unused and stale going forward since we wont be able to address them anymore, should we put some sort of documentation in place about how to manually revoke all the old client tokens?

benashz:
> Looks good, this seems like a healthy change. IIUC, this will have the undesired effect of leaving all previous clients (that use capital letters) unused and stale going forward since we wont be able to address them anymore, should we put some sort of documentation in place about how to manually revoke all the old client tokens?

Yeah, I had considered that but this bug prevented any `appRole` clients from ever being stored, so they were only ever in memory.

PR #301, opened 2023-07-31
VSS: fix issue where the response error was not being set

benashz:
While working on HCP issue I noticed that the error returned from Vault was never being set. The error was being assigned in the wrong scope. This PR ensures that the response error is properly scoped for later evaluation.

### DIFF
--- a/controllers/vaultstaticsecret_controller.go
+++ b/controllers/vaultstaticsecret_controller.go
@@ -80,31 +80,32 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	var resp *api.KVSecret
+	var respErr error
 	switch o.Spec.Type {
 	case consts.KVSecretTypeV1:
-		w, err := c.KVv1(o.Spec.Mount)
-		if err != nil {
+		if w, err := c.KVv1(o.Spec.Mount); err != nil {
 			return ctrl.Result{}, err
-		}
-		resp, err = w.Get(ctx, o.Spec.Path)
-	case consts.KVSecretTypeV2:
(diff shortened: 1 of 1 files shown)


controllers/vaultstaticsecret_controller.go
```diff
@@ -80,31 +80,32 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-			resp, err = w.GetVersion(ctx, o.Spec.Path, o.Spec.Version)
+			if o.Spec.Version == 0 {
+				resp, respErr = w.Get(ctx, o.Spec.Path)
+			} else {
+				resp, respErr = w.GetVersion(ctx, o.Spec.Path, o.Spec.Version)
+			}
 		}
 	default:
-		err = fmt.Errorf("unsupported secret type %q", o.Spec.Type)
+		err := fmt.Errorf("unsupported secret type %q", o.Spec.Type)
```
kschoche, line 102: I'm pretty sure we cannot hit the `default` code path anyways due to [this check](https://github.com/hashicorp/vault-secrets-operator/blob/3d45188560f240671652109390b1b8d55a6685ac/api/v1beta1/vaultstaticsecret_types.go#L32) in the spec, but good to have anyways I guess, unless I'm mistaken about how kubebuilder works here?

kschoche (approved):
Nice catch, looks good!

controllers/vaultstaticsecret_controller.go
benashz: True. Good to catch the issue if the marker's value ever gets out of sync with the controller.

PR #302, opened 2023-08-01
Simplify the way KV requests are handled

benashz:
Move away from the use of api.Client.Kv*, since it does not easily allow for Prometheus instrumentation and makes the overall ClientBase interface a bit more cumbersome.

### DIFF
--- a/controllers/vaultstaticsecret_controller.go
+++ b/controllers/vaultstaticsecret_controller.go
@@ -13,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -79,54 +78,24 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		requeueAfter = computeHorizonWithJitter(d)
 	}
 
-	var resp *api.KVSecret
-	var respErr error
-	switch o.Spec.Type {
-	case consts.KVSecretTypeV1:
-		if w, err := c.KVv1(o.Spec.Mount); err != nil {
-			return ctrl.Result{}, err
-		} else {
-			resp, respErr = w.Get(ctx, o.Spec.Path)
-		}
-	case consts.KVSecretTypeV2:
-		if w, err := c.KVv2(o.Spec.Mount); err != nil {
-			return ctrl.Result{}, err
-		} else {
-			if o.Spec.Version == 0 {
-				resp, respErr = w.Get(ctx, o.Spec.Path)
-			} else {
-				resp, respErr = w.GetVersion(ctx, o.Spec.Path, o.Spec.Version)
-			}
-		}
-	default:
-		err := fmt.Errorf("unsupported secret type %q", o.Spec.Type)
-		logger.Error(err, "")
+	kvReq, err := newKVRequest(o.Spec)
+	if err != nil {
 		r.Recorder.Event(o, corev1.EventTypeWarning, consts.ReasonVaultStaticSecret, err.Error())
 		return ctrl.Result{}, err
 	}
 
-	if respErr != nil {
-		logger.Error(err, "Failed to read Vault secret")
+	kvs, err := c.ReadKV(ctx, kvReq)
+	if err != nil {
 		r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonVaultClientError,
 			"Failed to read Vault secret: %s", err)
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
 	}
 
-	if resp == nil {
-		logger.Error(nil, "empty Vault secret", "mount", o.Spec.Mount, "path", o.Spec.Path)
-		r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonVaultClientError,
-			"Vault secret was empty, mount %s, path %s", o.Spec.Mount, o.Spec.Path)
-		return ctrl.Result{
-			RequeueAfter: requeueAfter,
-		}, nil
-	}
-
-	data, err := makeK8sSecret(resp)
+	data, err := makeK8sSecret(kvs)
 	if err != nil {
-		logger.Error(err, "Failed to construct k8s secret")
 		r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonVaultClientError,
-			"Failed to construct k8s secret: %s", err)
-		return ctrl.Result{}, err
+			"Invalid Vault Secret data: %s", err)
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
 	}
 
 	var doRolloutRestart bool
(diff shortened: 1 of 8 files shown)


internal/vault/vault.go
```diff
@@ -33,16 +34,41 @@ func UnmarshalPKIIssueResponse(resp *api.Secret) (*PKICertResponse, error) {
 	return result, nil
 }
 
+// MarshalSecretData returns Kubernetes Secret data from an api.Secret
+// TODO: move to internal/helpers
```
tvoran, line 38: Not sure I follow why this should be moved to `internal/helpers/`. It looks specific to vault?

controllers/vaultstaticsecret_controller.go
```diff
@@ -183,36 +152,10 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}, nil
 }
 
-// SetupWithManager sets up the controller with the Manager.
+// makeK8sSecret returns Kubernetes Secret data from an api.KVSecret
+// Deprecated: use vault.MarshalKVData instead
```
tvoran, line 156: Do we need to keep this around?

internal/vault/testutils_test.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, nil, err
+	}
+
+	server := &http.Server{
+		Handler:   handler,
+		TLSConfig: tlsConfig,
+	}
+
+	return server, ln, err
+}
```
tvoran, line 50: I'm curious why this function is split out from `NewTestHTTPServer()`, since I don't see it called independently.

internal/vault/testutils_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/hashicorp/vault/api"
+)
+
+// TestHTTPServer creates a test HTTP server that handles requests until
+// the listener returned is closed.
+// XXX: based off of github.com/hashicorp/vault/api/client_test.go
```
tvoran, line 21: What does `XXX` mean here? Is there a problem with this code?

internal/vault/responses.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (r *defaultResponse) Secret() *api.Secret {
+	return r.secret
+}
+
+func (r *defaultResponse) Data() map[string]any {
+	if r.secret == nil {
+		return nil
+		// return make(map[string]interface{}, 0)
```
tvoran, line 29: Still needed?

internal/vault/requests_test.go
```diff
@@ -0,0 +1,203 @@
...
+			r := &defaultWriteRequest{
+				path: tt.path,
+			}
+			assert.Equalf(t, tt.want, r.Path(), "Path()")
+		})
+	}
+}
+
+func Test_defaultWriteRequest_Params(t *testing.T) {
+	type fields struct{}
```
tvoran, line 174: How is this used?

internal/vault/requests_test.go
```diff
@@ -0,0 +1,203 @@
...
+			r := &defaultReadRequest{
+				path: tt.path,
+			}
+			assert.Equalf(t, tt.want, r.Path(), "Path()")
+		})
+	}
+}
+
+func Test_defaultReadRequest_Values(t *testing.T) {
+	type fields struct{}
```
tvoran, line 120: How is this used?

controllers/vaultstaticsecret_controller_test.go
```diff
@@ -91,7 +91,7 @@ func Test_makeK8sSecret(t *testing.T) {
 				},
 			},
 			expectedK8sSecret: nil,
-			expectedError:     fmt.Errorf(`failed to marshal key "password" from Vault secret: json: unsupported type: chan int`),
```
tvoran, line 94: The extra info in this error message seems like it could be useful, did we lose it for a reason?

controllers/vaultstaticsecret_controller.go
```diff
@@ -79,54 +78,24 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	}
 
-	if respErr != nil {
-		logger.Error(err, "Failed to read Vault secret")
+	resp, err := c.Read(ctx, kvReq)
+	if err != nil {
 		r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonVaultClientError,
 			"Failed to read Vault secret: %s", err)
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
```
tvoran, line 91: `requeueAfter` is the refresh interval here? It seems like this should requeue sooner than the next refresh interval if there's an error on read?

controllers/vaultstaticsecret_controller.go
```diff
@@ -79,54 +78,24 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-
-	data, err := makeK8sSecret(resp)
+	data, err := vault.MarshalData(resp.Data(), resp.Secret().Data)
 	if err != nil {
-		logger.Error(err, "Failed to construct k8s secret")
 		r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonVaultClientError,
-			"Failed to construct k8s secret: %s", err)
-		return ctrl.Result{}, err
+			"Invalid Vault Secret data: %s", err)
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
```
tvoran, line 98: Similar question here.

internal/vault/testutils_test.go
benashz: I can probably move it back to NewTestHTTPServer(), I left it here in case we want to setup an HTTPTestServer.

internal/vault/responses.go
benashz: Yes, we need this to satisfy the Response interface.

internal/vault/requests_test.go
benashz: Ah, good catch. This structure auto generated by the Goland. I usually move all members to the `test{}`- missed this clean up.

internal/vault/requests_test.go
benashz: Will drop it.

controllers/vaultstaticsecret_controller.go
benashz: Yeah, that seems like a good call.

internal/vault/responses.go
tvoran: Oh I was referring to the commented-out line here.

internal/vault/vault.go
benashz: Fixed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

controllers/vaultstaticsecret_controller.go
benashz: Dropped in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

internal/vault/testutils_test.go
benashz: Not sure, fixed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

internal/vault/responses.go
benashz: Fixed in 9b92290386d57e86cfa535b74cd0752ac6917ea3

internal/vault/requests_test.go
benashz: Fixed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

internal/vault/requests_test.go
benashz: Fixed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

controllers/vaultstaticsecret_controller.go
benashz: Addressed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

controllers/vaultstaticsecret_controller.go
benashz: Addressed in 7fa55d5ffeda41e8c2705cc32de3385a04e0ce1e

internal/vault/vault.go
tvoran: Ah, it's now vault-agnostic, nice!

internal/vault/client.go
```diff
@@ -537,14 +553,14 @@ func (c *defaultClient) renew(ctx context.Context) error {
 		return errs
 	}
 
-	resp, err := c.Write(ctx, "/auth/token/renew-self", nil)
+	resp, err := c.Write(ctx, nil)
```
tvoran, line 556: Does this still work without "/auth/token/renew-self" in the request?

internal/vault/responses.go
tvoran: Got one of them but missed this one.

internal/vault/vault_test.go
tvoran, line 1: Are there more tests to be added here? It looks like `MakeSecretK8sData()` is covered by the responses tests if I understand correctly.

internal/vault/client.go
benashz: Oh, good catch. Bad refactoring.

internal/vault/vault_test.go
benashz: Yes, but it would be good to add explicit tests for `MakeSecretK8sData()`

internal/vault/responses.go
benashz: Fixed in c359450fde4bb9167caf25b9b1664d2c038b8057

internal/vault/vault_test.go
benashz: Added in c359450fde4bb9167caf25b9b1664d2c038b8057

tvoran (approved):

internal/vault/testutils_test.go
benashz: Decided to leave it as-is.

PR #304, opened 2023-08-02
Add extra envs to Helmchart values.yaml

### LINKED ISSUE #287: Allow setting additional environment variables via Helmchart
We need to set additional environment variables (e.g. `HTTP_PROXY`) to our deployment of the controller-manager. As I don't think that we're the only ones with such use-cases, would it be possible to allow for such modifications to happen via the Helmchart?

Similar to many other Helmcharts, make a Value `controller.env` available which takes either a `list` or `dict` of environment variables to be injected into the `manager` container of the controller pod.

Right now we are patching the deployment once after deploying via Helm, which is error-prone and needs to be documented.

I'd happily contribute the minor changes needed to the Helmchart, if you want to go ahead and assign me to the issue.

mocdaniel:
This PR fixes #287 by introducing a new Helmchart value called `extraEnvs` to the `manager` container of the operator's deployment.

It allows users to define arbitrary environment variables they might need inside their operator, e.g. `HTTP_PROXY`/`NO_PROXY` settings.

I saw no use in adding the same configuration to the `kube-rbac-proxy` container, let me know if I should adapt this container's configuration, too.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -83,6 +83,10 @@ spec:
               fieldPath: metadata.uid
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ .Values.controller.kubernetesClusterDomain }}
+        {{- range .Values.controller.manager.extraEnvs }}
+        - name: {{ .name }}
+          value: {{ .value | quote }}
+        {{- end }}
         image: {{ .Values.controller.manager.image.repository }}:{{ .Values.controller.manager.image.tag }}
         imagePullPolicy: IfNotPresent
         livenessProbe:
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -257,6 +257,14 @@ controller:
     # @type: integer
     maxConcurrentReconciles:
 
+    # Defines additional environment variables to be added to the
+    # vault-secrets-operator container.
+    # extraEnvs:
+    #   - name: HTTP_PROXY
+    #     value: http://proxy.example.com
+    # @type: list
+    extraEnvs: []
+
     # Configures the default resources for the vault-secrets-operator container.
     # For more information on configuring resources, see the K8s documentation:
     # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/


chart/templates/deployment.yaml
```diff
@@ -83,6 +83,10 @@ spec:
               fieldPath: metadata.uid
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ .Values.controller.kubernetesClusterDomain }}
+        {{- range .Values.controller.manager.extraEnvs }}
+        - name: {{ .name }}
+          value: {{ .value | quote }}
```
benashz, line 88: Could this result in double quoting the value e.g.: `.value = "foo"`

benashz (commented):
Thanks for the PR! I've added a few comments/suggestions.  Also, it would be great if you could extend the `bats` tests here: https://github.com/hashicorp/vault-secrets-operator/tree/main/test/unit

chart/templates/deployment.yaml
mocdaniel: No, `quote` doesn't double quote already quoted strings. It's best practice to always pipe inputs expected to be strings or env values to `quote`, [according to the Helm documentation](https://helm.sh/docs/howto/charts_tips_and_tricks/#quote-strings-dont-quote-integers).

mocdaniel:
> Thanks for the PR! I've added a few comments/suggestions. Also, it would be great if you could extend the `bats` tests here: https://github.com/hashicorp/vault-secrets-operator/tree/main/test/unit

I added four new tests, one for the Chart's default configuration (no extraEnv), and one for each of the allowed env types (string, number, quoted string).

test/unit/deployment.bats
```diff
@@ -352,3 +352,67 @@ load _helpers
...
+      helm template -s templates/deployment.yaml --values /dev/stdin . |   \
+      tee /dev/stderr |  \
+      yq '.spec.template.spec.containers[1].env | select(documentIndex == 1)' |  \
+      tee /dev/stderr)
+
+    local actual=$(echo "$object" | yq '.[3].name' | tee /dev/stderr)
+    [ "${actual}" = 'QUOTED_ENV' ]
+    actual=$(echo "$object" | yq '.[3].value' | tee /dev/stderr)
+    [ "${actual}" = 'noquotesneeded' ]
+}
```
benashz, line 418: Looks like we lost the end of file newline.

test/unit/deployment.bats
```diff
@@ -352,3 +352,67 @@ load _helpers
...
+    cd `chart_dir`
+    local object=$(helm template  \
+      -s templates/deployment.yaml  \
+      --set 'controller.manager.extraEnv[0].name=RANDOM_PORT'  \
+      --set 'controller.manager.extraEnv[0].value=42'  \
+      . | tee /dev/stderr |  \
+      yq '.spec.template.spec.containers[1].env | select(documentIndex == 1)' |  \
+      tee /dev/stderr)
+
+    local actual=$(echo "$object" | yq '.[3].name' | tee /dev/stderr)
```
benashz, line 397: We should probably add a length check here to make sure we are not picking up any unexpected values.

test/unit/deployment.bats
```diff
@@ -352,3 +352,67 @@ load _helpers
...
+    cd `chart_dir`
+    local object=$(helm template  \
+      -s templates/deployment.yaml  \
+      --set 'controller.manager.extraEnv[0].name=HTTP_PROXY'  \
+      --set 'controller.manager.extraEnv[0].value=http://proxy.example.com/'  \
+      . | tee /dev/stderr |  \
+      yq '.spec.template.spec.containers[1].env | select(documentIndex == 1)' |  \
+      tee /dev/stderr)
+
+    local actual=$(echo "$object" | yq '.[3].name' | tee /dev/stderr)
```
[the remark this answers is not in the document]
mocdaniel: I will add this.

test/unit/deployment.bats
```diff
@@ -352,3 +352,67 @@ load _helpers
...
+    local actual=$(echo "$object" | yq '. | length' | tee /dev/stderr)
+    [ "${actual}" = '3' ]
+}
+
+@test "controller/Deployment: extra env string values can be set" {
+    cd `chart_dir`
+    local object=$(helm template  \
+      -s templates/deployment.yaml  \
+      --set 'controller.manager.extraEnv[0].name=HTTP_PROXY'  \
+      --set 'controller.manager.extraEnv[0].value=http://proxy.example.com/'  \
```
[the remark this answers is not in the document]
mocdaniel: I will add this, too.

test/unit/deployment.bats
mocdaniel: Could you elaborate on this? I don't get it, I'm afraid. I am already checking default behaviour (no `extraVar` set) in a dedicated check, am I missing something?

mocdaniel:
Regarding the quoting issue:

In your example, you *explicitly* quote the value (the equivalent to `value: \"42\"` in YAML if I'm not mistaken). Those quotes get preserved, as they are interpreted as intentional, since YAML doesn't require quoting of strings.

My test cases covers a *normal* quoted string in a `values.yaml`, as some users like to quote all their strings (the equivalent to `value: "some"`), as well as unquoted strings (`value: some`), and integers (`value: 42`).

However, I looked up some popular Helmcharts and how they handle envs, and the majority in fact doesn't quote the envs. So I would leave the final decision whether to quote envs or not to you.

Once this question is settled I will refactor the changes to the chart and tests to reflect our final decision.

benashz:
> Regarding the quoting issue:
> 
> In your example, you _explicitly_ quote the value (the equivalent to `value: \"42\"` in YAML if I'm not mistaken). Those quotes get preserved, as they are interpreted as intentional, since YAML doesn't require quoting of strings.

Indeed. Imagine the case where the value contains whitespace, it would need to be quoted on for the shell to properly handle it.

> 
> My test cases covers a _normal_ quoted string in a `values.yaml`, as some users like to quote all their strings (the equivalent to `value: "some"`), as well as unquoted strings (`value: some`), and integers (`value: 42`).
> 

I think it is best to treat all values as strings.

> However, I looked up some popular Helmcharts and how they handle envs, and the majority in fact doesn't quote the envs. So I would leave the final decision whether to quote envs or not to you
> 
> Once this question is settled I will refactor the changes to the chart and tests to reflect our final decision

Let's drop the  ` | quote`. Thanks!

mocdaniel:
I dropped `quote` and adjusted the tests. All comments should be resolved now.

benashz (approved):
LGTM!

Thank you for your contribution to HashiCorp!

PR #307, opened 2023-08-03
Collapse Client.ReadKV() into Client.Read()

benashz:
Unify client.ReadKV() with Client.Read() by implementing some common response/request interfaces. This means one less method on the Client interface, which should make overall Client interface simpler.


internal/vault/responses.go
```diff
@@ -0,0 +1,69 @@
...
+import "github.com/hashicorp/vault/api"
+
+var (
+	_ Response = (*defaultResponse)(nil)
+	_ Response = (*kvV2Response)(nil)
+)
+
+type Response interface {
+	Secret() *api.Secret
+	Data() map[string]interface{}
```
tvoran, line 15: Is there anything preventing this from returning `map[string]any` instead?

controllers/vaultdynamicsecret_controller.go
```diff
@@ -293,9 +293,9 @@ func (r *VaultDynamicSecretReconciler) syncSecret(ctx context.Context, c vault.C
 
 	switch method {
 	case http.MethodPut, http.MethodPost:
-		resp, err = c.Write(ctx, path, params)
+		resp, err = c.Write(ctx, vault.NewWriteRequest(path, params, method))
```
tvoran, line 296: I'm curious how the `method` param is used? I'm not seeing `Method()` called in the client, for example.

benashz: Oops, yeah it is only related to the VDS controller. All Write() requests are `PUT`. Fixing.

internal/vault/responses.go
benashz: Nope, should be fine to do so.

benashz: Updated with a74d5224b88c8929d26ee7380886c267a29d10aa

controllers/vaultdynamicsecret_controller.go
benashz: Fixed with a74d5224b88c8929d26ee7380886c267a29d10aa

tvoran (approved):

PR #312, opened 2023-08-04
Refactor credential packages for incoming HCP support

benashz:
This PR makes room for the incoming HCP authentication suppport being tracked in #313. The new layout allows for us to add some new HCP specific credential providers. Previously the addition would result in cyclical dependencies.

Some highlights:
- Move the `credentials` package from `internal/vault` to `internal/credentials`
- Move the hmac code from `internal/vault` to `internal/helpers`
- Move all vault credential provider code from `internal/vault/credentials` to `internal/credentials/vault`


tvoran:
Would you mind adding a bit more info to the description as to how the directory/package structure is changing in this PR?

tvoran (approved):
Thanks for the clarification! I think I get it now.

PR #313, opened 2023-08-04
Add new HCPAuth CRD and Controller

benashz:
Extend VSO to support HCP authentication by HCP Service Principal.

Sample HCPAuth Custom Resource
```yaml
kind: HCPAuth
metadata:
  name: default
spec:
  method: servicePrincipal
  organizationID: xxxxxxxx-xxxx-xxxx-xxxx-444444444444
  projectID: xxxxxxxx-xxxx-xxxx-xxxx-999999999999
  servicePrincipal:
    secretRef: vso-db-demo-hcp-vsa-sp
```


tvoran (commented):
Just a couple questions so far, but this looks pretty straightforward 👍

controllers/hcpauth_controller.go
```diff
@@ -0,0 +1,101 @@
...
+			return ctrl.Result{}, nil
+		}
+
+		logger.Error(err, "error getting resource from k8s", "secret", o)
+		return ctrl.Result{}, err
+	}
+
+	// perform a rudimentary health check on the HCP host on port 443.
+	conn, err := net.DialTimeout("tcp",
+		fmt.Sprintf("%s:443", hvsclient.DefaultHost), time.Second*5)
```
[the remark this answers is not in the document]
benashz: The user could check the clusters network configuration to troubleshoot why VSO cannot talk to the HCP service. In addition to that, a failed validation can be gleaned from the Prometheus  metrics. See `updateStatus()`. This could result in an alert being generated for the user/operator.

controllers/hcpauth_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	var requeueAfter time.Duration
+	if errs != nil {
+		requeueAfter = computeHorizonWithJitter(time.Second * 15)
+	}
+
+	return ctrl.Result{
+		RequeueAfter: requeueAfter,
+	}, nil
```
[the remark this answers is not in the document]
benashz: Ah, probably left over from testing. Good catch!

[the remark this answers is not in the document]
benashz: Actually, it will on be 15s if there was an error, by default it will be 0 (so no requeue)

internal/credentials/hcp/service_principal.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (l *ServicePrincipleCredentialProvider) Init(ctx context.Context, client ctrlclient.Client,
+	authObj *secretsv1beta1.HCPAuth, providerNamespace string,
+) error {
+	logger := log.FromContext(ctx)
+	l.authObj = authObj
+	l.providerNamespace = providerNamespace
+
+	// We use the UID of the secret which holds the AppRole Role's secret_id for the provider UID
```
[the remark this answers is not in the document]
benashz: fixed with 5f4551cfa63218a45ee89f35f924b58fc67402d9

controllers/hcpauth_controller.go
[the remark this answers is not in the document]
benashz: When/if we allow for configuring the host we can ensure that it is updated here as well.

controllers/hcpauth_controller.go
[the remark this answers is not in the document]
tvoran: Ah yep, I misread that, makes sense.

controllers/hcpauth_controller.go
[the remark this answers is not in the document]
tvoran: 👍

chart/templates/manager-rbac.yaml
```diff
@@ -212,6 +212,38 @@ rules:
...
+  verbs:
+    - get
+    - list
+    - watch
+- apiGroups:
+    - secrets.hashicorp.com
+  resources:
+    - hcpauths/status
+  verbs:
+    - get
```
tvoran, line 246: I forget if this is autogenerated, but it looks like these permissions (lines 233-246) are already included above.

chart/templates/manager-rbac.yaml
```diff
@@ -212,6 +212,38 @@ rules:
...
+    - list
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - secrets.hashicorp.com
+  resources:
+    - hcpauths/status
+  verbs:
+    - get
```
tvoran, line 232: Does this need `patch` and `update` as in config/rbac/role.yaml?

internal/credentials/hcp/service_principal_test.go
```diff
@@ -0,0 +1,163 @@
...
+				ProviderSecretClientID:  "client-id-1",
+				ProviderSecretClientKey: "client-key-1",
+			},
+			wantErr: assert.NoError,
+		},
+		{
+			name:              "secret-not-found",
+			authObj:           authObj,
+			providerNamespace: "tenant-ns",
+			client:            fake.NewClientBuilder().Build(),
```
tvoran, line 61: Looks like `client` is defined the same way in all the cases, so we might not need it as a test parameter?

kschoche (approved):
LGTM, had done a bit of review and played around with it a couple weeks back, great work.

internal/credentials/hcp/service_principal_test.go
benashz: Each test case needs its own client instance to ensure it does not pick up any cruft from the other tests.

chart/templates/manager-rbac.yaml
benashz: Interesting. They are not autogenerated, unfortunately.

chart/templates/manager-rbac.yaml
benashz: Good catch! I had to merge the hcpauths stuff config/rbac/role.yaml. We will need to look at how we can ensure that the RBAC config is always synced to the chart.

benashz: Fixed with fb82a562276ec679a7058512783b0156e7b06f9a

chart/templates/manager-rbac.yaml
```diff
@@ -212,6 +212,32 @@ rules:
...
+  resources:
+  - hcpauths/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - secrets.hashicorp.com
+  resources:
+  - vaultauths
```
tvoran, line 232: This probably needs to be hcpauths?

```suggestion
  - hcpauths
```

tvoran (approved):

internal/credentials/hcp/service_principal_test.go
tvoran: Sure, wasn't suggesting using the same client for all tests, just that the client could be defined at the start of the test run like the other local variables there.

chart/templates/manager-rbac.yaml
benashz: yup, bad merge. fixing..

benashz: Fixed with 510c96aebdc5590f24654c91c9e5caaecc0d8647

internal/credentials/hcp/service_principal_test.go
benashz: Updated with 865c24f050833a8cf29874f2d3003f254d2e6b75

PR #314, opened 2023-08-04
Add new HCPVaultSecretsApp CRD and Controller

benashz:
Extends VSO to support syncing HCP Vault Secrets to K8s.

Sample HCPVaultSecretsApp Custom Resource:
```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: HCPVaultSecretsApp
metadata:
  name: vso-db-demo-hcp-vsa
  namespace: demo-ns
spec:
  appName: vso
  destination:
    create: true
    name: vso-db-demo-hcp-vsa
```

PR #355, opened 2023-09-13
HVSA: ensure secrets are properly synced

benashz:
HVSA secrets were no longer be synced after merging #314. The logic used to determine whether or not to do the sync was invalid. We will be adding full acceptance tests soon.


tvoran (approved):

kschoche (approved):

benashz:
Integration tests added with #357

PR #376, opened 2023-09-25
Add validation for spec.refreshAfter and min constraints for HVSA

benashz:
Ensure that only support duration string values can be set for `spec.refreshAfter`. This change affects VaultStaticSecret and HCPVaultSecretsApp specs.

In addition to the change above, apply a minimum constraint on the configured HVSA refreshAfter, with the default value being 30s. This value can be set from the command line option `-min-refresh-after-hvsa`.

Additional enhancements:
- chart: support extending the controller.manager's args with the new extraArgs value.

### DIFF
--- a/api/v1beta1/hcpvaultsecretsapp_types.go
+++ b/api/v1beta1/hcpvaultsecretsapp_types.go
@@ -19,7 +19,7 @@ type HCPVaultSecretsAppSpec struct {
 	HCPAuthRef string `json:"hcpAuthRef,omitempty"`
 	// RefreshAfter a period of time, in duration notation e.g. 30s, 1m, 24h
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^(-?[0-9]+(\\.[0-9]+)?(s|m|h))+$"
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(s|m|h))$"
 	// +kubebuilder:default="600s"
 	RefreshAfter string `json:"refreshAfter,omitempty"`
 	// RolloutRestartTargets should be configured whenever the application(s)
--- a/api/v1beta1/hcpvaultsecretsapp_types.go
+++ b/api/v1beta1/hcpvaultsecretsapp_types.go
@@ -18,6 +18,8 @@ type HCPVaultSecretsAppSpec struct {
 	// Kubernetes namespace. HCPAuthRef string `json:"hcpAuthRef,omitempty"`
 	HCPAuthRef string `json:"hcpAuthRef,omitempty"`
 	// RefreshAfter a period of time, in duration notation e.g. 30s, 1m, 24h
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^(-?[0-9]+(\\.[0-9]+)?(s|m|h))+$"
 	// +kubebuilder:default="600s"
 	RefreshAfter string `json:"refreshAfter,omitempty"`
 	// RolloutRestartTargets should be configured whenever the application(s)
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -9,12 +9,12 @@ import (
 	"math/rand"
 	"time"
 
-	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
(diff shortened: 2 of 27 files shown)


api/v1beta1/hcpvaultsecretsapp_types.go
```diff
@@ -18,6 +18,8 @@ type HCPVaultSecretsAppSpec struct {
 	// Kubernetes namespace. HCPAuthRef string `json:"hcpAuthRef,omitempty"`
 	HCPAuthRef string `json:"hcpAuthRef,omitempty"`
 	// RefreshAfter a period of time, in duration notation e.g. 30s, 1m, 24h
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^(-?[0-9]+(\\.[0-9]+)?(s|m|h))+$"
```
tomhjp, line 22: I could be misreading, but a few minor suggestions on the regex:

* It seems like it would have accepted negative values before?
* The + on the end seems like it would allow the whole pattern to repeat, like 600s600s?
* `[.]` saves any headache around interpreting the level of escaping needed for backslashes.

```suggestion
	// +kubebuilder:validation:Pattern="^([0-9]+([.][0-9]+)?(s|m|h))$"
```

tomhjp (approved):
LGTM

api/v1beta1/hcpvaultsecretsapp_types.go
benashz: +1 for the catch on the duration string duplication. I think we can probably omit the negative duration for now. I believe that was to support the case where no refresh could be configured e.g: `skipRefresh := d < 0`, but I think we can probably omit support for that. It is possible to set the min-refresh to 0s and then `spec.refreshAfter` could be set to 0s.

test/unit/deployment.bats
```diff
@@ -700,3 +700,34 @@ load _helpers
...
+
+#--------------------------------------------------------------------
+# extraArgs
+
+@test "controller/Deployment: extraArgs not set by default" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -s templates/deployment.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[1].args | select(documentIndex == 1)' | tee /dev/stderr)
```
tvoran, line 712: Something like this might make these more robust:

```suggestion
      yq 'select(.kind == "Deployment") | .spec.template.spec.containers[] | select(.name == "manager") | .args' | tee /dev/stderr)
```

tvoran (approved):

api/v1beta1/hcpvaultsecretsapp_types.go
benashz: Think we still need to escape '.' in the character class. I am inclined to leave it as is since it's the only character allowed.

E.g:
 Invalid value: "^(?[0-9]+([.][0-9]+)?(s|m|h))$": must be a valid regular expression

test/unit/deployment.bats
benashz: +1, I think I was following what we have in the `extraLabels` tests.

api/v1beta1/hcpvaultsecretsapp_types.go
benashz: Updated in c6929b4eb6bd75e59a86f5e4e68b5a3962f7275f

kschoche (approved):
Looks good!

controllers/common.go
```diff
@@ -184,3 +184,22 @@ func removeFinalizers(ctx context.Context, c client.Client, log logr.Logger, obj
 	}
 	log.Info(fmt.Sprintf("Removed %d finalizers", cnt))
 }
+
+func parseDurationString(ds, path string, min time.Duration) (time.Duration, error) {
```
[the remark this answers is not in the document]
benashz: Updated in https://github.com/hashicorp/vault-secrets-operator/pull/376/commits/a6580774a37a1cd0178c88d9746c8dbf829ea4f6

test/unit/deployment.bats
benashz: Updated in https://github.com/hashicorp/vault-secrets-operator/pull/376/commits/a6580774a37a1cd0178c88d9746c8dbf829ea4f6

PR #377, opened 2023-09-25
Support running bats tests in parallel

benashz:
Add support for running all bats test in parallel. It requires that the `parallel` command be installed on the build host.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,7 +45,7 @@ jobs:
       - run: npm install -g bats@${BATS_VERSION}
         shell: bash
         env:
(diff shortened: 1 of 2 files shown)


kschoche (approved):
LGTM! I think the ci runners are only 1 or 2 cpu so it may not make substantial difference but shouldn't hopefully make it worse!

benashz:
> LGTM! I think the ci runners are only 1 or 2 cpu so it may not make substantial difference but shouldn't hopefully make it worse!

Seems to have cut the unit-test time in half. We should look into getting some larger runners, 2 cores is probably less than ideal.

PR #381, opened 2023-09-27
Update CRD manifests

benashz:
Missed a `make manifest` do to this being a docs PR. Will update the GH workflow in another PR to catch this next time.

### DIFF
--- a/chart/crds/secrets.hashicorp.com_hcpauths.yaml
+++ b/chart/crds/secrets.hashicorp.com_hcpauths.yaml
@@ -68,11 +68,11 @@ spec:
                   reasons, only project-level service principals should ever be used.
                 properties:
                   secretRef:
-                    description: "SecretRef is the name of a Kubernetes secret in
-                      the consumer's (VDS/VSS/PKI/HCP) namespace which provides the
-                      HCP ServicePrincipal clientID, and clientSecret. \n The secret
-                      data must have the following structure { \"clientID\": \"clientID\",
-                      \"clientSecret\": \"clientSecret\", }"
+                    description: 'SecretRef is the name of a Kubernetes secret in
+                      the consumer''s (VDS/VSS/PKI/HCP) namespace which provides the
+                      HCP ServicePrincipal clientID, and clientSecret. The secret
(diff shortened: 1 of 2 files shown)


chart/crds/secrets.hashicorp.com_hcpauths.yaml
```diff
@@ -68,11 +68,11 @@ spec:
                   reasons, only project-level service principals should ever be used.
                 properties:
                   secretRef:
-                    description: "SecretRef is the name of a Kubernetes secret in
-                      the consumer's (VDS/VSS/PKI/HCP) namespace which provides the
-                      HCP ServicePrincipal clientID, and clientSecret. \n The secret
-                      data must have the following structure { \"clientID\": \"clientID\",
-                      \"clientSecret\": \"clientSecret\", }"
+                    description: 'SecretRef is the name of a Kubernetes secret in
+                      the consumer''s (VDS/VSS/PKI/HCP) namespace which provides the
```
kschoche, line 72:
```quote
                      the consumer''s (VDS/VSS/PKI/HCP) namespace which provides the
```

I'm confused, does this work? I see two ticks here, wont that break the string, or remove the punctuation?
Same potential problem in the other file?

benashz: Not sure, this is generated from `make manifests`

benashz: The CRDs are always synced to the chart, as well.

benashz: Seems like it is just escaping the single quote:
```
$ kubectl -n vault-secrets-operator-system get crds hcpauths.secrets.hashicorp.com -o yaml|grep consumer
                      the consumer's (VDS/VSS/PKI/HCP) namespace which provides the
```

kschoche: Wild, alright then :D 👍🏽

kschoche (approved):

PR #382, opened 2023-09-27
HVSA: include User-Agent and requester HTTP request headers.

benashz:
These are being added to identify VSO as an HCP client.

### DIFF
--- a/controllers/hcpvaultsecretsapp_controller.go
+++ b/controllers/hcpvaultsecretsapp_controller.go
@@ -7,8 +7,10 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net/http"
 	"time"
 
+	httptransport "github.com/go-openapi/runtime/client"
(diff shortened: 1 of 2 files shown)


controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -28,8 +30,16 @@ import (
 	"github.com/hashicorp/vault-secrets-operator/internal/credentials"
 	"github.com/hashicorp/vault-secrets-operator/internal/credentials/hcp"
 	"github.com/hashicorp/vault-secrets-operator/internal/helpers"
+	"github.com/hashicorp/vault-secrets-operator/internal/version"
 )
 
+const (
+	headerHCPCLIRequester = "X-HCP-CLI-Requester"
```
VinnyHC, line 37:
```suggestion
	headerHCPCLIRequester = "x-hvs-requester"
```

benashz: Good catch! Fixed in 3342a4ad3a0fdb0f0dff6ababeb6eb5cc3c4541f . Since HTTP headers are case-insensitive I've preserved the capitalization.

kschoche (approved):

PR #383, opened 2023-09-27
Remove/disable revoke on uninstall

benashz:
The revocation Helm pre-delete job could sometimes fail forever, preventing the VSO release from being uninstalled. This PR temporarily disables this feature until VAULT-20196 is resolved


kschoche (approved):
👍🏽 - assuming everything still passes :)

PR #387, opened 2023-09-28
Check Helm and Kustomize versions for production builds.

benashz:
Adds a script that verifies that version tags match what the build thinks the `VERSION` should be. This fix  should help to ensure that the Helm/Kustomize/VSO versions are always in sync.

 have been updated in both the Helm chart and the Kustomize config. 
- [x] check Kustomize config
- [x] check Helm chart config

### DIFF
--- a/scripts/check-versions.sh
+++ b/scripts/check-versions.sh
@@ -31,7 +31,6 @@ function checkVersion {
  fi
 
  local doc="$(cat ${filename})"
- local queries=
  echo "* Checking version(s) in ${filename}"
  for query in "${@:2}"
  do
--- a/scripts/check-versions.sh
+++ b/scripts/check-versions.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+set -e
+
+if [ -z "${VERSION}" ]; then
+  echo "VERSION variable must be set" >&2
+  exit 1
+fi
+
+if [ "${VERSION}" == "0.0.0-dev" ]; then
+  echo "dev build, skipping version checks"
+  exit 0
+fi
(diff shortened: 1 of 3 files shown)


scripts/check-versions.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo "dev build, skipping version checks"
+  exit 0
+fi
+
+ROOT_DIR="${0%/*}"
+# update PATH to prefer scripts relative to this one e.g. yq
+export PATH="${ROOT_DIR}:${PATH}"
+
+CHART_ROOT="${CHART_ROOT-$(readlink -f ${0%/*}/../chart)}"
+KUSTOMIZE_ROOT="${KUSTOMIZE_ROOT-$(readlink -f ${0%/*}/../config)}"
```
tvoran, line 22: Any reason we wouldn't use ROOT_DIR here?

```suggestion
CHART_ROOT="${CHART_ROOT-$(readlink -f ${ROOT_DIR}/../chart)}"
KUSTOMIZE_ROOT="${KUSTOMIZE_ROOT-$(readlink -f ${ROOT_DIR}/../config)}"
```

tvoran (approved):

scripts/check-versions.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [ -z "${VERSION}" ]; then
+  echo "VERSION variable must be set" >&2
+  exit 1
+fi
+
+if [ "${VERSION}" == "0.0.0-dev" ]; then
+  echo "dev build, skipping version checks"
+  exit 0
+fi
```
tvoran, line 15: Just thinking out loud, someday we may want to relax this a little to allow any `-dev` builds, for testing images internally without doing a full release.

benashz: See #396

PR #389, opened 2023-09-28
Update remaining integration tests to use the VSO TF Helm module

benashz:
Updates the following integration tests to use the VSO TF Hem module:

- TestVaultAuthMethods
- TestVaultPKISecret
- TestVaultStaticSecret
- TestRevocation (tested locally, since it is disabled)

### DIFF
--- a/test/integration/revocation/terraform/main.tf
+++ b/test/integration/revocation/terraform/main.tf
@@ -104,39 +104,29 @@ resource "vault_kubernetes_auth_backend_role" "default" {
 }
 
 # VSO Helm chart
-resource "helm_release" "vault-secrets-operator" {
-  name             = "test"
-  namespace        = var.operator_namespace
-  create_namespace = true
-  wait             = true
-  chart            = var.operator_helm_chart_path
-
-  set {
-    name  = "controller.manager.image.repository"
-    value = var.operator_image_repo
-  }
-  set {
-    name  = "controller.manager.image.tag"
-    value = var.operator_image_tag
-  }
-
-  # Connection Configuration
-  set {
-    name  = "defaultVaultConnection.enabled"
-    value = "true"
-  }
-  set {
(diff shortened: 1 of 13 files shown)


test/integration/revocation/terraform/main.tf
```diff
@@ -104,39 +104,29 @@ resource "vault_kubernetes_auth_backend_role" "default" {
...
+      vault_connection_ref            = ""
+      namespace                       = ""
+      mount                           = ""
+      transit_mount                   = ""
+      key_name                        = ""
+      method                          = ""
+      kubernetes_auth_role            = ""
+      kubernetes_auth_service_account = ""
+      kubernetes_auth_token_audiences = ""
+    }
```
tvoran, line 130: Is this still needed? It looks like it matches the defaults in the module?

tvoran (approved):

test/integration/vaultauthmethods/terraform/main.tf
```diff
@@ -246,36 +246,19 @@ resource "kubernetes_secret" "static-creds" {
...
-    name  = "defaultVaultConnection.address"
-    value = var.k8s_vault_connection_address
-  }
+module "vso-helm" {
+  source                       = "../../modules/vso-helm"
+  operator_namespace           = var.operator_namespace
+  operator_image_repo          = var.operator_image_repo
+  operator_image_tag           = var.operator_image_tag
+  enable_default_connection    = true
+  enable_default_auth_method   = false
```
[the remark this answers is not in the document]
benashz: I think it's a Helm only test, so these are hard coded. We can add them variables if we test variant that needs rhythm them?

test/integration/revocation/terraform/main.tf
benashz: Unfortunately Terraform requires that they be set, despite there being defaults.

Issue #412, opened 2023-10-17
HCP Vault Secret with Vault Secret Operator failed to verify certificate


bizalu:
**Describe the bug**
I try to synchronise secret from HCP Vault Secret and my kubernetes cluster (openshift 4.13) with the Vault Secret Operator (v0.3.2).
I create my HCPAuth and HCPVaultSecretsApp and I’ve got error on my logs from vault-secrets-operator-controller-manager pod.

**To Reproduce**
1. Deploy VSO operator from Redhat certified operator catalog
2. Create a secret with my clienID and clienSecret (service principal with contributor role)
3. Create a HCPAuth with my organizationID and projectID
4. Create a HCPVaultSecretsApp 
5. See error ( vault-secrets-operator-controller-manager)

HCPAuth :
```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: HCPAuth
metadata:
  name: hcp-media
  namespace: openshift-vault-operator
spec:
  method: servicePrincipal
  organizationID: xxxx
  projectID: xxx
  servicePrincipal:
    secretRef: hcpvaultmedia
```

HCPVaultSecretsApp :
```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: HCPVaultSecretsApp
metadata:
  name: media-test
  namespace: openshift-vault-operator
spec:
  appName: media
  destination:
    create: true
    name: test
    type: Opaque
  hcpAuthRef: hcp-media
  refreshAfter: 600s
```

Log from vault-secrets-operator-controller-manager pod :
```log
2023-10-06T05:00:32Z ERROR Get App Secret {"controller": "hcpvaultsecretsapp", "controllerGroup": "secrets.hashicorp.com", "controllerKind": "HCPVaultSecretsApp", "HCPVaultSecretsApp": {"name":"xxx","namespace":"xxx"}, "namespace": "xxx", "name": "xxx", "reconcileID": "xxx", "appName": "xxx", "error": "Get \"https://api.cloud.hashicorp.com:443/secrets/2023-06-13/organizations/xxx/apps/media/open\": Post \"https://auth.idp.hashicorp.com/oauth2/token\": tls: failed to verify certificate: x509: certificate signed by unknown authority"}
```

**Environment**
* Kubernetes version: Openshift 4.13.14
* vault-secrets-operator version: 0.3.2

bizalu:
Thanks @benashz,

[...]

PR #415, opened 2023-10-19
Include the tls-ca-bundle.pem from ubi-minimal

benashz:
The ubi-micro base image does not include a trusted ca-bundle. This PR adds support for copying the latest bundle from `ubi-minimal:latest` into the target UBI VSO image.

Closes: #412

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,7 @@ ENTRYPOINT ["/vault-secrets-operator"]
 
 # ubi build image
 # -----------------------------------
-FROM registry.access.redhat.com/ubi9/ubi-minimal:latest as build-ubi
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2-750.1697625013 as build-ubi
 RUN microdnf update --nodocs && microdnf install ca-certificates --nodocs
 
 # ubi release image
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -66,6 +66,11 @@ USER 65532:65532
 
 ENTRYPOINT ["/vault-secrets-operator"]
 
+# ubi build image
+# -----------------------------------
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest as build-ubi
```
tvoran, line 71: We may want to explicitly version the ubi-minimal image as well:
```suggestion
FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2-750.1697625013 as build-ubi
```

tvoran (approved):

Dockerfile
benashz: Seems reasonable - I wonder if we should always be pulling latest for both minimal and micro, that approach would be in-line with our distroless builds.

tvoran: `latest` would certainly be convenient, though it makes it a little harder to tell which image was used for a release (when tracking down CVEs, etc.) `distroless` seems to be kind of an outlier since it doesn't have tagged versions.

benashz: Agreed. I prefer an explicit version over one derived at build time.

PR #430, opened 2023-10-27
PullPolicy in deployment.yaml ~ Add possibility to change pullPolicy for rbac & manager image

M-A-X-I-M:
I add to the **deployment.yaml** the possibility to change the **pullPolic**y in **values.yaml.**
I set the **pullPolicy** to **always** in **values.yaml** because it's the best practice in k8s. 
We use Kyverno in our k8s Clusters and is checking if the pullPolicy is setting to always, otherwise we can't deploy an image.
So it can be very useful for people with Kyverno.


chart/values.yaml
```diff
@@ -62,6 +62,7 @@ controller:
     image:
       repository: gcr.io/kubebuilder/kube-rbac-proxy
       tag: v0.14.4
+      pullPolicy: Always
```
benashz, line 65: This will introduce a behavioral change, as the default pull policy is `IfNotPresent`.  We can set the default of `IfNotPresent` here, or leave it empty and use default from a helper.

Another idea would be to add a global image pull policy, so all policies could be set from a single value. I think that would probably be our preferred approach, here.

chart/values.yaml
```diff
@@ -101,6 +102,7 @@ controller:
     image:
       repository: hashicorp/vault-secrets-operator
       tag: 0.3.4
+      pullPolicy: Always
```
benashz, line 105: See the comment related to the kubeRbacProxy container change above.

benashz (commented):
Thanks for submitting this PR. It would be good if you could extend the chart tests under `test/unit`.

We want to keep the default policy `IfNotPresent` and more than likely add a global image pull policy directive.

chart/values.yaml
M-A-X-I-M: ok, I will set pullPolicy in values.yaml to "IfNotPreset"

M-A-X-I-M: I set it to "IfNotPresent"

chart/values.yaml
M-A-X-I-M: I set it to ""IfNotPresent

M-A-X-I-M (commented):
I set the imagePullPolicy in values.yaml to "IfNotPresent"

M-A-X-I-M:
Now I think is ready to go, sorry that I was late for version v0.4.0, I hope it will be merged in version v0.5.0

benashz:
> Now I think is ready to go, sorry that I was late for version v0.4.0, I hope it will be merged in version v0.5.0

Thanks @M-A-X-I-M  , I wonder if you mind extending the bats tests for this new chart feature? After that we can move forward with merging this in for v0.5.0.

benashz:
Closing as a duplicate of #473

Issue #336, opened 2023-08-21
Namespaced Service Accounts


kcighon:
Off the back of #291 and #274, should the service account defined in VaultAuth also support namespaces? This will support global default Auth and Connections

If a VaultStaticSecret is located in a namespace which differs to the VaultAuth, the VaultStaticSecret will try to use the service account defined in the VaultAuth from within the VaultStaticSecret namespace.

Observed error in VaultStaticSecret:
```bash
Warning VaultClientConfigError 7m5s (x18 over 18m) VaultStaticSecret Failed to get Vault auth login: ServiceAccount "my-vso-sa" not found
```

Possible solution - as with #291, allow for namespaced values for spec.kubernetes.serviceAccount. e.g.
`spec.kubernetes.serviceAccount: vault-secret-operator-system/my-vso-sa`

TJM:
I am seeing this problem as well. Apparently when you fall back to the `default` VaultAuth in the "operators" namespace, it still attempts to find the service account in the local secrets namespace. I would think that it would use the service account that is in the same namespace as the VaultAuth, or as mentioned here, allow for a namespace prefix so that you can specify which one to use. ServiceAccounts do not have the concept of `allowedNamespaces`, so that might be tricky to support securely.

PR #431, opened 2023-10-27
internal/credentials: make kubernetes auth service account namespaced

inigohu:
This commit adds the ability to specify a namespace for the service account used to authenticate to Vault. This is useful when you want to use a service account in a different namespace than the one the secret (vaultStaticSecret, vaultDynamicSecret) is located. This change is backwards compatible, so if no namespace is specified, the service account will be looked up in the same namespace as the secret.

Example: Here the service account that will be used to authenticate to Vault is the service account `default` that it is in the namespace "vault-secrets-operator-system".
```yaml
kind: VaultAuth
metadata:
  name: static-auth
  namespace: app
spec:
  vaultConnectionRef: vault-connection
  allowedNamespaces:
    - "*"
  method: kubernetes
  mount: demo-auth-mount
  kubernetes:
    role: role1
    serviceAccount: vault-secrets-operator-system/default
```

Closes https://github.com/hashicorp/vault-secrets-operator/issues/336

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -402,3 +402,17 @@ func NewSyncableSecretMetaData(obj ctrlclient.Object) (*SyncableSecretMetaData, 
 		return nil, fmt.Errorf("unsupported type %T", t)
 	}
 }
+
+// GetKubernetesServiceAccountNamespacedName returns the NamespacedName for the Kubernetes VaultAuth's configured
+// serviceAccount.
+// If the serviceAccount is empty then defaults Namespace and Name will be returned.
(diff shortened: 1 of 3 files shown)


kylehodgetts:
This is great! I hope this gets merged soon

lingxiankong:
Keen to get this feature merged ASAP

benashz:
Unfortunately, extending VSO to access service accounts outside of the Secret CR's K8s namespace would violate VSO's security model. VSO requires that a service account (Vault credential provider) be configured in the Secret CR's K8s namespace.

The recommended approach is to create a service account for each tenant/application that will be used specifically for authenticating to Vault. Access to that service account should controlling using K8s RBAC.

PR #432, opened 2023-10-27
Fix AppRole secret reference

ownercoder:
Hi!

AppRole has `SecretRef` field, which contains name of the k8s secret. When i create AuthStatic in different namespace, i can't get secret, because operator try to get secret from current namespace.

For example:
VaultAuth with name `default` placed in vault-secrets-operator namespace, also i've created k8s secret in this namespace. When i create AuthStatic with `vault-secrets-operator/default` auth reference in different namespace, i've got an error `Failed to get secret ` because it try to get secret from namespace StaticAuth CR.


tvoran:
Hi @ownercoder, thanks for the patch here, but this is working as intended; credentials for auth should be coming from the VaultStaticSecret's namespace. We have some updated documentation about authentication in VSO up here now in case it's helpful: https://developer.hashicorp.com/vault/docs/platform/k8s/vso/sources/vault/auth

PR #474, opened 2023-11-21
Add allowedNamespaces to defaultAuthMethod

gillcaleb:
The actual change is relatively trivial. As written, it will only render the field if the value is set. The setting includes the " | quote" to account for the case when the value is a "*" wildcard. 

I'm still getting a handle on the unit and integration testing so for now my only touchpoint on that was adding the variables into the integration vso module. Apologies if this is insufficient as this is my first commit.


kschoche (changes requested):
Hi @gillcaleb -
Thanks for filing this PR, I think it looks really close.
As for testing, I think we can get by with just unit test changes, instead of the integration tests. Could you add a couple unit tests to `tests/unit/default-vault-auth-method.bats` to validate the behaviour?
Thanks!
~Kyle

stavvy-cgill:
Appreciate the review @kschoche!

I added various combinations of the setting throughout the existing units tests. I also change it to use the " | toJson" to better align with the way the way string arrays are handled currently (e.g. tokenAudiences). Let me know if there's anything else you'd like me to change.

benashz:
Closing in favor of #602. Thank you for your contribution to HashiCorp!

Issue #271, opened 2023-06-19
Artificial delay for eventually consistent secrets


adrianmoisey:
**Is your feature request related to a problem? Please describe.**
Some secrets (ie, AWS IAM) are eventually consistent and require a delay before they can be used. 

**Describe the solution you'd like**
A method to introduce a delay before VSO writes secrets to Kubernetes

**Describe alternatives you've considered**
It may be possible to get the pods that consume VSO secrets to have a delay before attempting to use their secrets, but then logic needs to be build into each application. It may make sense for VSO to handle this delay, as it a central service/tool.

**Additional context**
To quote:
https://developer.hashicorp.com/vault/docs/secrets/aws#usage
> Unfortunately, IAM credentials are eventually consistent with respect to other Amazon services. If you are planning on using these credential in a pipeline, you may need to add a delay of 5-10 seconds (or more) after fetching credentials before they can be used successfully.

adrianmoisey:
This bug is still hurting us. Is there any way it can be prioritised? I made a PR but it hasn't been looked at.

adrianmoisey:
Hey @benashz 
Sorry for the ping! 

Would it be possible for someone to take a look at this issue?
I did make a PR to handle it, but that went unreviewed.
I'm happy to fix up that PR if needed.

Thanks.

adrianmoisey:
Pinging this again. We still suffer from this issue.
I'm happy to update my PR to get this sorted.

PR #477, opened 2023-11-22
Add a configurable delay before writing a dynamic secret

adrianmoisey:
Fixes https://github.com/hashicorp/vault-secrets-operator/issues/271

Some services (such as AWS IAM) are eventually consistent and require some time between generating the secret, and using the secret.

Without this delay our services can't access AWS for a short while immediately after the secret rotation happens.

I'm not sure if my tests are sane. They could introduce some flakiness.


adrianmoisey:
Hey @benashz 
Would it be possible to get this change (or something similar) looked at?

This is a blocker for us from using VSO.

adrianmoisey:
Is it possible for this change to be merged?

I'm happy to do the work to get it updated, but I'm not certain it will actually get prioritised and merged.

PR #480, opened 2023-11-27
Remove Duplicate UBI Tags

emilymianeil:
This PR is intended to avoid a [production incident](https://github.com/hashicorp/releng-support/issues/123) where there are duplicate tags (auto-assigned) to UBI images. The entire issue is described [here](https://github.com/hashicorp/crt-workflows-common/pull/407#issue-1999804835), and can be avoided by building `UBI`s (intended for both `Dockerhub` and `Redhat`) in one step.

### PR Checklist

* [x] appropriate backport labels added
* [x] not a security concern

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -147,8 +147,8 @@ jobs:
             exit 1
           fi
 
-  build-docker-ubi-dockerhub:
-    name: UBI ${{ matrix.arch }} build for dockerhub
+  build-docker-ubi:
+    name: UBI ${{ matrix.arch }} build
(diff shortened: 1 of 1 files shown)


benashz:
Thanks for the PR @emilymianeil , but It looks like the UBI builds are failing with this change e.g: https://github.com/hashicorp/vault-secrets-operator/actions/runs/7007865159/job/19063019074?pr=480 - Please let us know if we can be of assistance.

benashz (approved):
SGTM! Thanks @emilymianeil!

PR #483, opened 2023-11-28
rename max-concurrent-reconciles-vds to max-concurrent-reconciles

kschoche:
Renames the controller manager flag `--max-concurrent-reconciles-vds` to `--max-concurrent-reconciles` and applies it to VPS, VSS, VHCPSA and VDS controllers.
This has the effect of enabling the all Syncable Secret controllers to process in parallel which we've tested at scale successfully and was causing scaling issues in the current architecture.

The original flag `--max-concurrent-reconciles-vds` is preserved for backward compatibility and will override the new flag for the VDS controller only.

The helm `values.yaml` is updated to reflect the new behaviour of `controller.manager.maxConcurrentReconciles`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -272,6 +272,7 @@ func main() {
 	// `--max-concurrent-reconciles`.
 	vdsOverrideOpts := controller.Options{}
 	if vdsOptions.MaxConcurrentReconciles != defaultVaultDynamicSecretsConcurrency {
+		setupLog.Info("The flag --max-concurrent-reconciles-vds has been deprecated, but will still be honored to set the VDS controller concurrency, please use --max-concurrent-reconciles.")
 		vdsOverrideOpts = vdsOptions
 	} else {
 		vdsOverrideOpts = controllerOptions
--- a/main.go
+++ b/main.go
@@ -55,7 +55,7 @@ func main() {
 	persistenceModelDirectUnencrypted := "direct-unencrypted"
 	persistenceModelDirectEncrypted := "direct-encrypted"
 	defaultPersistenceModel := persistenceModelNone
-	vdsOptions := controller.Options{}
+	controllerOptions := controller.Options{}
 	cfc := vclient.DefaultCachingClientFactoryConfig()
 	startTime := time.Now()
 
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -83,8 +83,8 @@ func main() {
 		fmt.Sprintf(
 			"The type of client cache persistence model that should be employed."+
 				"choices=%v", []string{persistenceModelDirectUnencrypted, persistenceModelDirectEncrypted, persistenceModelNone}))
-	flag.IntVar(&vdsOptions.MaxConcurrentReconciles, "max-concurrent-reconciles-vds", 100,
-		"Maximum number of concurrent reconciles for the VaultDynamicSecrets controller.")
+	flag.IntVar(&controllerOptions.MaxConcurrentReconciles, "max-concurrent-reconciles-global", 100,
```
benashz, line 86: I wonder if we could rename the option to `max-concurrent-reconciles`, since global seems to imply that we support per controller reconciles. Also, I wonder if we might be breaking anyone who has set `max-concurrent-reconciles-vds`, should we maybe leave that around as a deprecated option. We could assign its value to the option?

kschoche: Sounds good about renaming to `max-concurrent-reconciles` and applying to all controllers.

Keeping the old flag is maybe a bit hacky since we don't have a way of knowing if the flag was even passed due to the way the `flags.Int()` function takes a default value. I could set it to `0` by default and ignore it unless it's non-zero I suppose.
At any rate, this would only presumably be a problem for someone who's using Kustomize and had set the flag, and for those who are using the helm chart but pinning old versions of the operator binary. I think that second one is a bit non-standard and questionably supportable?

I guess I wonder if it's a trade-off of documenting it as a breaking change for Kustomize users, and/or making the code a little sloppy? Thoughts?

benashz: Yeah, keeping it around is not the best, but I think it is possible to set it from `controller.manager.extraArgs` here: https://github.com/hashicorp/vault-secrets-operator/blob/5845d8abe3d96fe9b9e3b6c27d1bce6aef78e020/chart/values.yaml#L330 - we can document this as a breaking change, or keep it as an alias to the new option.

kschoche: Okay - moved everything to `max-concurrent-reconciles`, preserving `max-concurrent-reconciles-vds` which will override `max-concurrent-reconciles` if supplied.

main.go
```diff
@@ -257,13 +267,22 @@ func main() {
 		setupLog.Error(err, "Unable to create controller", "controller", "VaultConnection")
 		os.Exit(1)
 	}
+	// This allows the user to customize VDS concurrency independently.
+	// It is mostly here to allow for backward compatibility from when we introduced the flag
+	// `--max-concurrent-reconciles`.
+	vdsOverrideOpts := controller.Options{}
+	if vdsOptions.MaxConcurrentReconciles != defaultVaultDynamicSecretsConcurrency {
```
benashz, line 274: Can we log a deprecation warning if `vdsOptions.MaxConcurrentReconciles > 0`?

main.go
```diff
@@ -83,8 +91,10 @@ func main() {
 		fmt.Sprintf(
 			"The type of client cache persistence model that should be employed."+
 				"choices=%v", []string{persistenceModelDirectUnencrypted, persistenceModelDirectEncrypted, persistenceModelNone}))
-	flag.IntVar(&vdsOptions.MaxConcurrentReconciles, "max-concurrent-reconciles-vds", 100,
+	flag.IntVar(&vdsOptions.MaxConcurrentReconciles, "max-concurrent-reconciles-vds", defaultVaultDynamicSecretsConcurrency,
```
benashz, line 94: Can we document the deprecation here, pointing the user at the option?

benashz (approved):
LGTM!

PR #502, opened 2023-12-06
Make bats run faster

benashz:
Drop use of the yq docker image, replacing it with a binary from the project repo.


kschoche (approved):
Nice!

PR #503, opened 2023-12-06
image/ubi: add separate target and build job for RedHat

tvoran:
Adds a separate ubi image build target and job for the RedHat container registry, that only builds amd64, until we get around our multi-arch issues there.


emilymianeil (approved):
🥳 LGTM

dekimsey (approved):
Looks like an accurate revert of #480 to me

tvoran:
Thanks!

PR #510, opened 2023-12-08
Drop metadata.replaces from clusterserviceversion

benashz:
This PR puts us in line with the currently published OLM. We believe the directive caused the OLM PR to be rejected by Red Hat.


kschoche (approved):
Not 100% what this does but I see you've been working with releng on it so go for it :D

PR #527, opened 2023-12-20
VC: provide default value for spec.skipTLSVerify

### LINKED ISSUE #278: Sync Chart and Go type for VaultConnection skipTLSVerify
**Describe the bug**
The type for a VaultConnection is [Here](https://github.com/hashicorp/vault-secrets-operator/blob/b65708ca82157dd4102dea341aa6647f3cd43b5e/api/v1beta1/vaultconnection_types.go#L24C48-L24C48)

The Helm chart defines the defaultVaultConnection and specifies the address and skipTLSVerify as "non-optional" items.  This means the chart, [here](https://github.com/hashicorp/vault-secrets-operator/blob/b65708ca82157dd4102dea341aa6647f3cd43b5e/chart/templates/default-vault-connection.yaml#L19), always includes the skptTLSVerify value, even if it is the "default" of false.  It only includes the remaining attributes if they are specified in the values for the chart.

So there is a discrepancy. The Helm Chart manifest for the defaultVaultConnection includes skipTLSVerify, whereas the Go type lists it as omitempty.  The resulting YAML when you get the created object from Kubernetes therefore does not have the skipTLSVerify.

This causes problems for orchestration engines like Rancher, which will compare the rendered YAML and that which was actually deployed, and flag this as a modified value (it is in the YAML for the chart, but is not shown in the YAML when doing a kubectl get vaultconnection/default -oyaml.

It's questionable whether a boolean value can actually be "empty" anyway.  Yes, the "default" or "empty" or "null" or "zero" value for a boolean in Go is false.  However, that does not translate to the Helm Chart, and YAML that should be returned by Kubernetes.

I _think_ we should change the Helm Chart so that the skipTLSVerify is not included in the YAML if the value is false.  That way the deployed YAML will match the expected YAML.

**To Reproduce**
1. Deploy Helm chart with Rancher
2. Note you will have to comment out KubeVersion in the main Chart.yaml, as there is a known bug in Rancher that does not detect the kubernetes version correctly.  May be fixed in latest version just released (2.7.5), but I'm on 2.7.4.
3. You will see in Rancher that there is a failure because the YAML was modified from what it expects to be deployed.  This shows up as: Modified(1) [Bundle vault-vault-vault-secrets-operator]; vaultconnection.secrets.hashicorp.com vault-secrets-operator/default modified {"spec":{"skipTLSVerify":false}}

**Expected behavior**
The spec in the YAML deployed by the Helm Chart reports back the same as was specified (after rendering).

**Environment**
* Kubernetes version: 
  * K3S 1.25.10
  * Rancher 2.7.4
* vault-secrets-operator version:
  * Chart version 0.1.0

**Additional context**
I'll try and submit a PR, but the change is like a line or 3 in the Chart.

benashz:
Some K8s distros like Rancher do their own Helm chart diff to the running state of a given K8s cluster. For the default VaultConnection provided by Helm chart, the value for spec.skipTLSVerify would always trigger a diff in the case where the corresponding chart value was false, since runtime configuration did not include spec.skipTLSVerify.

Interestingly enough, if the default VaultConnection.spec.skipTLSVerify was added/set to false using 'kubectl edit', the value would persist in the cluster.

Solution:

Update VaultConnectionSpec.SkipTLSVerify to use a default value for rather than setting its JSON flag to omitempty. This will ensure that comparisons between the Helm chart's default VaultConnection's spec.skipTLSVerify can be reliably made, even when the default VC is updated outside of Helm.

Closes #278

### DIFF
--- a/api/v1beta1/vaultconnection_types.go
+++ b/api/v1beta1/vaultconnection_types.go
@@ -21,7 +21,8 @@ type VaultConnectionSpec struct {
 	// CACertSecretRef is the name of a Kubernetes secret containing the trusted PEM encoded CA certificate chain as `ca.crt`.
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 	// SkipTLSVerify for TLS connections.
-	SkipTLSVerify bool `json:"skipTLSVerify,omitempty"`
+	// +kubebuilder:default=false
+	SkipTLSVerify bool `json:"skipTLSVerify"`
 }
 
 // VaultConnectionStatus defines the observed state of VaultConnection
--- a/chart/crds/secrets.hashicorp.com_vaultconnections.yaml
+++ b/chart/crds/secrets.hashicorp.com_vaultconnections.yaml
@@ -50,6 +50,7 @@ spec:
                 description: Headers to be included in all Vault requests.
                 type: object
               skipTLSVerify:
+                default: false
                 description: SkipTLSVerify for TLS connections.
                 type: boolean
               tlsServerName:
@@ -57,6 +58,7 @@ spec:
                 type: string
             required:
             - address
+            - skipTLSVerify
             type: object
           status:
             description: VaultConnectionStatus defines the observed state of VaultConnection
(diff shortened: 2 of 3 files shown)


kschoche (approved):
LGTM

PR #528, opened 2023-12-21
Update paused deployment error message

Jrc356:
Noticed this error was worded a bit oddly for the context


kschoche (approved):
Thanks for catching this!

PR #532, opened 2023-12-21
VDS: remediate deleted destination secret

### LINKED ISSUE #228: operator doesn't refresh a deleted result secret if lease still within window
2023-05-30T21:33:44Z	DEBUG	events	Not in renewal window after transitioning to a new leader/pod, lease_id=database/creds/keycloak2-keycloak-user/wdcXRo6BHgZ0g3naSyhLlxJT, horizon=499h31m52.92219023s	{"type": "Normal", "object": {"kind":"VaultDynamicSecret","namespace":"keycloak","name":"vault-db-creds-keycloak","uid":"1a6be57d-e3ab-401a-a127-f1ed5e06618b","apiVersion":"secrets.hashicorp.com/v1alpha1","resourceVersion":"39754840"}, "reason": "SecretLeaseRenewal"}

**Describe the bug**
secret was revoked and trying to replace but the operator seems to only care about age of existing entity even after deleting the result kube secret.

**To Reproduce**
1. obtain valid secret
2. attempt to get secret updated due to early revoke
3. ...
5. See error (vault-secrets-operator logs, application logs, etc.)

Application deployment:

**Expected behavior**
Refresh of invalid destination secret

**Environment**
* Kubernetes version: 1.24.4
* vault-secrets-operator version: beta.1

**Additional context**
trying to fix db cred generation, the operator seems to only grab locally known expired creds once it obtains one. needs to check if the destination generated secret still exists as a fallback at minimum.

benashz:
Previously, the VDS controller did not detect when the destination secret had been deleted after it had initially created it.

Fixes:
- ensure a deleted destination secret that is owned by the VDS controller results in a re-sync.
- add integration test to test this use-case
- update all other controller integration tests to test for destination deletion

Closes #228

### DIFF
--- a/controllers/vaultdynamicsecret_controller.go
+++ b/controllers/vaultdynamicsecret_controller.go
@@ -111,9 +111,16 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 	}
 
+	destExists, _ := helpers.CheckSecretExists(ctx, r.Client, o)
+	if !o.Spec.Destination.Create && !destExists {
+		logger.Info("Destination secret does not exist, either create it or "+
+			"set .spec.destination.create=true", "destination", o.Spec.Destination)
+		return ctrl.Result{RequeueAfter: requeueDurationOnError}, nil
+	}
+
 	// doSync indicates that the controller should perform the secret sync,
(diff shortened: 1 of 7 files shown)


kschoche (approved):
Looks good!

PR #534, opened 2023-12-22
tests: make assertDynamicSecretRotation() more robust.

benashz:
Previously, assertDynamicSecretRotation() would fail when updating the VDS object, due to the updated instance not being the latest. The update should be retried a few times before giving up.

### DIFF
--- a/test/integration/vaultdynamicsecret_integration_test.go
+++ b/test/integration/vaultdynamicsecret_integration_test.go
@@ -569,45 +569,75 @@ func assertDynamicSecretNewGeneration(t *testing.T,
 	t.Helper()
 
 	objKey := ctrlclient.ObjectKeyFromObject(vdsObjOrig)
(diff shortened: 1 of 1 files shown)


test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -569,45 +569,75 @@ func assertDynamicSecretNewGeneration(t *testing.T,
...
+			t.Logf(
+				"Retrying client.Update() of %s, err=%s, delay=%s", objKey, err, d)
+		},
+	)
+
+	if !assert.NoError(t, err) {
+		return
 	}
+
+	// wait for the object ot be reconciled
```
thyton, line 601:
```suggestion
	// wait for the object to be reconciled
```

thyton (approved):
LGTM!

PR #535, opened 2023-12-22
WIP: Parallel integration tests

benashz:


benashz:
> Great work bringing all of these together. There are quite a few commented out pieces of code / terraform, does it make sense to clean them up?

Ah, thought I had gotten them all. Good catch! FIxed in 905377be3da3d49526a4ea220d7743e7b1539638

> 
> A couple things that I ran into when running this locally. My normal workflow is something like this: `make setup-kind setup-integration-test load-docker-image integration-test` But now we end up with duplicate `kubernetes_namespace.vault` and `kubernetes_cluster_role_binding.oidc-reviewer` and the `make integration-test` fails presumably because `setup-integration-test` isnt necessary anymore?

I think it should still be necessary. Are you running that against a fresh kind cluster?

I ran the same command against a fresh cluster, and did not run into any issues.

> 
> If I just do `make setup-kind load-docker-image integration-test` I'm getting a lot of terraform state errors, even when I believe I'd cleaned up everything from past runs:
> 
> ```
> TestVaultAuthMethods 2024-02-02T16:22:29-06:00 retry.go:103: terraform [init -upgrade=false] returned an error: error while running command: exit status 1; ╷
> │ Error: Failed to query available provider packages
> │
> │ Could not retrieve the list of available versions for provider
> │ hashicorp/helm: locked provider registry.terraform.io/hashicorp/helm 2.8.0
> │ does not match configured version constraint 2.11.0; must use terraform
> │ init -upgrade to allow selection of new versions
> ╵
> . Sleeping for 500ms and will try again.
> ```
> 
> Also using `ctrl-c` to cancel the test doesn't seem to work, I have to `ctrl-c` several times for the test to run

Interesting. I will take a closer look.

> 
> Happy to walk through it next week with you, its probably something on my side but wanted to be sure.

👍

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -31,8 +30,37 @@ import (
...
+	TransitKeyName string `json:"transit_key_name"`
+	TransitRef     string `json:"transit_ref"`
+}
+
 func TestVaultDynamicSecret(t *testing.T) {
-	testID := fmt.Sprintf("vds")
+	if testInParallel {
+		//if testWithHelm {
+		//	t.Fatal("Parallel tests not support with Helm")
+		//}
```
kschoche, line 60:
```suggestion
```
I think we can get rid of this?

kschoche (approved):
Looks great!
I ran it locally and after cleaning up some stale tf state files in my repository everything works as expected and is down to ~5 minutes for me.
excellent work!

PR #545, opened 2024-01-10
Take ownership of an existing destination secret

### LINKED ISSUE #337: Cannot overwrite existing secrets
**Describe the bug**
After deploying vso to kubernetes cluster with kube auth, new static secrets are working just fine, however when dealing with existing secrets, and trying to use vso to manage the existing secret with the same name i am getting the following errors:

```
invalid owner label, key=app.kubernetes.io/name, present=false
invalid owner label, key=app.kubernetes.io/managed-by, present=false
secret default/oauth-client-secret has no ownerReferences
not the owner of the destination Secret default/oauth-client-secret     {"type": "Warning", "object": {"kind":"VaultStaticSecret","namespace":"default","name":"oauth-client-secret","uid":"f5ea7c7e-ff85-4274-acd4-4a1c7d53c77","apiVersion":"secrets.hashicorp.com/v1beta1","resourceVersion":"37293181"}, "reason": "SecretSyncError"}
```

**To Reproduce**
to deploy VSO (vso excerpt):

```
resource "helm_release" "vault-secrets-operator" {
  depends_on      = [kubernetes_namespace.ns] // Add this line
  wait            = true
  wait_for_jobs   = true
  name            = "vault-secrets-operator"
  repository      = "https://helm.releases.hashicorp.com"
  chart           = "vault-secrets-operator"
  namespace       = kubernetes_namespace.ns.metadata.0.name

  values = [templatefile("${path.module}/values.yaml.tpl", {
    vault_kube_auth_role = vault_kubernetes_auth_backend_role.vso.role_name
    namespace            = var.namespace
    vault_ip_address     = local.vault_ip_address
    auth_mount_path      = "${var.gcp_project}/${var.cluster_name}"
    gke_audience         = local.gke_audience
  })]
}

resource "vault_auth_backend" "kubernetes" {
  type = "kubernetes"
  path = "${var.gcp_project}/${var.cluster_name}" // mount path
}

resource "kubernetes_secret_v1" "sa" {
  depends_on                     = [helm_release.vault-secrets-operator]
  wait_for_service_account_token = true
  metadata {
    name      = "vault-sa"
    namespace = var.namespace
    annotations = {
      "kubernetes.io/service-account.name" = "vault-secrets-operator-controller-manager"
    }
  }
  type = "kubernetes.io/service-account-token"
}

data "kubernetes_config_map_v1" "ca" {
  metadata {
    name      = "kube-root-ca.crt"
    namespace = "kube-public"
  }
}

resource "vault_kubernetes_auth_backend_config" "kubernetes" {
  backend                = vault_auth_backend.kubernetes.path
  kubernetes_host        = "https://${data.google_container_cluster.primary.endpoint}"
  kubernetes_ca_cert     = data.kubernetes_config_map_v1.ca.data["ca.crt"]
  token_reviewer_jwt     = kubernetes_secret_v1.sa.data["token"]
  disable_iss_validation = true
  disable_local_ca_jwt   = true
  issuer                 = "https://container.googleapis.com/v1/projects/${var.gcp_project}/locations/${var.region}/clusters/${var.cluster_name}"
}

resource "vault_kubernetes_auth_backend_role" "vso" {
  backend                          = vault_auth_backend.kubernetes.path
  role_name                        = "vso"
  bound_service_account_names      = ["default"]
  bound_service_account_namespaces = ["*"]
  token_ttl                        = 3600
  token_policies                   = ["default", vault_policy.vso.name]
  audience                         = local.gke_audience
}

resource "vault_policy" "vso" {
  name   = "vso"
  policy = file("${path.module}/policies/vso.hcl") // read policy from file
}
```

Check the existing secret
λ › k get secrets                                                                                                                                   
NAME                       TYPE     DATA   AGE
oauth-client-secret        Opaque   2      19h

To write a secret with the same name so that it can be vault managed. 

```yaml
kind: VaultStaticSecret
metadata:
  namespace: test-vso2
  name: oauth-client-secret
spec:
  mount: secret #path of your secret engine, it could also be user/ or secret/
  type: kv-v2 #type of secret engine
  path: gcp-project/cluster-name/namespace/oauth-client-secret
  refreshAfter: 60s
  destination:
    create: true
    name: oauth-client-secret #what you want the display name to be in kubernetes secret.
    
   
   ```
   
   After applying that i am getting this error:
   ```
invalid owner label, key=app.kubernetes.io/name, present=false
invalid owner label, key=app.kubernetes.io/managed-by, present=false
secret default/oauth-client-secret has no ownerReferences
not the owner of the destination Secret default/oauth-client-secret     {"type": "Warning", "object": {"kind":"VaultStaticSecret","namespace":"default","name":"oauth-client-secret","uid":"f5ea7c7e-ff85-4274-acd4-4a1c7d53c77","apiVersion":"secrets.hashicorp.com/v1beta1","resourceVersion":"37293181"}, "reason": "SecretSyncError"}
```

**Expected behavior**
it should be overwritten by vso, and the secrets ttl should refresh and be now managed by vso. 

**Environment**
* Kubernetes version: 
  * GKE 1.24 
* vault-secrets-operator version: 0.1.0

**Additional context**
I am trying to do a secrets migration, and i believe vso need to be able to overwrite existing secrets with the same name. i am doing this via automation, and would not like to delete secrets then overwrite them. it could cause down time.

benashz:
Adds a new configuration option `spec.destination.overwrite` that when set to true VSO will replace an existing destination secret that it does not currently own. VSO will then take ownership of the destination secret's life-cycle.

Closes #337

### DIFF
--- a/internal/helpers/secrets.go
+++ b/internal/helpers/secrets.go
@@ -373,12 +373,7 @@ func checkSecretIsOwnedByObj(dest *corev1.Secret, references []metav1.OwnerRefer
 	// checking for Secret ownership relies on first checking the Secret's labels,
 	// then verifying that its OwnerReferences match the SyncableSecret.
 	errs := checkOwnerLabels(dest)
-	for k, v := range OwnerLabels {
-		if o, ok := dest.Labels[k]; o != v || !ok {
-			errs = errors.Join(errs, fmt.Errorf(
-				"invalid owner label, key=%s, present=%t", k, ok))
-		}
-	}
+
 	// check that obj is the Secret's true Owner
 	key := ctrlclient.ObjectKeyFromObject(dest)
 	if len(dest.OwnerReferences) > 0 {
--- a/internal/helpers/secrets.go
+++ b/internal/helpers/secrets.go
@@ -150,14 +150,9 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 		return fmt.Errorf("invalid Destination, err=%w", err)
 	}
 
-	var dest corev1.Secret
-	exists := true
-	if err := client.Get(ctx, key, &dest); err != nil {
-		if apierrors.IsNotFound(err) {
-			exists = false
-		} else {
-			return err
-		}
+	dest, exists, err := getSecretExists(ctx, client, key)
+	if err != nil {
+		return err
 	}
 
 	pruneOrphans := func() {
@@ -186,7 +181,7 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 		// syncable-secret's Status, but...
 		dest.Data = data
 		logger.V(consts.LogLevelDebug).Info("Updating secret")
-		if err := client.Update(ctx, &dest); err != nil {
+		if err := client.Update(ctx, dest); err != nil {
 			return err
 		}
 
@@ -213,21 +208,28 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
 	}
 	if exists {
 		logger.V(consts.LogLevelDebug).Info("Found pre-existing secret",
-			"secret", ctrlclient.ObjectKeyFromObject(&dest))
-		if err := checkSecretIsOwnedByObj(&dest, references); err != nil {
-			return err
+			"secret", ctrlclient.ObjectKeyFromObject(dest))
+
+		checkOwnerShip := true
+		if meta.Destination.Overwrite {
+			checkOwnerShip = hasOwnerLabels(dest)
 		}
 
+		if checkOwnerShip {
+			if err := checkSecretIsOwnedByObj(dest, references); err != nil {
+				return err
+			}
+		}
 	} else {
 		// secret does not exist, so we are going to create it.
-		dest = corev1.Secret{
+		dest = &corev1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      meta.Destination.Name,
 				Namespace: obj.GetNamespace(),
 			},
 		}
 		logger.V(consts.LogLevelDebug).Info("Creating new secret",
-			"secret", ctrlclient.ObjectKeyFromObject(&dest))
+			"secret", ctrlclient.ObjectKeyFromObject(dest))
 	}
 
 	// common setup/updates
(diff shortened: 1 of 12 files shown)


internal/helpers/secrets_test.go
```diff
@@ -393,6 +400,37 @@ func TestSyncSecret(t *testing.T) {
...
+			name:               "dest-exists-owned-overwrite-true",
+			client:             clientBuilder.Build(),
+			obj:                ownerWithDestClobber,
+			createDest:         true,
+			destLabels:         OwnerLabels,
+			expectSecretsCount: 1,
+			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
+				return assert.ErrorContains(t, err,
+					"not the owner of the destination Secret foo/baz")
+			},
```
tvoran, line 432: Does the operator check the type of object that owns a Secret? Just thinking it might be good to allow overwriting ownership if the owner is a non-VSO object, to help when migrating to VSO. Or I suppose it could allow multiple owner references, as long as only one is the VSO object being reconciled.

tvoran (approved):
Left a couple thoughts, but it looks great!

internal/helpers/secrets.go
```diff
@@ -337,22 +335,52 @@ func GetSecret(ctx context.Context, client ctrlclient.Client, objKey ctrlclient.
...
-	// check that all owner labels are present and valid, if not return an error
-	// this may cause issues if we ever add new "owner" labels, but for now this check should be good enough.
-	key := ctrlclient.ObjectKeyFromObject(dest)
+	errs := checkOwnerLabels(dest)
 	for k, v := range OwnerLabels {
 		if o, ok := dest.Labels[k]; o != v || !ok {
 			errs = errors.Join(errs, fmt.Errorf(
 				"invalid owner label, key=%s, present=%t", k, ok))
 		}
 	}
```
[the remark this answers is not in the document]
benashz: Oops, good catch! Fixed in 3cf8a7659a0aae13b99dfba820d6902a3236ba94

kschoche (approved):

internal/helpers/secrets_test.go
benashz: I believe with this change VSO will only take ownership of the destination secret if it does not have the generic VSO owner labels.

Allowing other Kinds to own the secret is an interesting idea, but I am not sure how that would work in the case where the "other" Kind is deleted - VSO would probably resolve that on the next reconciliation.

PR #546, opened 2024-01-10
Support excluding _raw from the destination

### LINKED ISSUE #424: Make _raw data in Kubernetes Secrets optional
Make it possible to prevent the addition of the `_raw` data to Kubernetes Secrets. This is causing issue with for example the Elastic-Cloud-on-Kubernetes Operator, since it checks the data in the secrets and fails for example if there are other data next to the license json.

Output of the operator logs:
```
"log.level":"ERROR", "message":"fatal exception while booting Elasticsearch", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"main","log.logger":"org.elasticsearch.bootstrap.Elasticsearch","elasticsearch.node.name":"elastic-es-masterdata-2","elasticsearch.cluster.name":"elastic","error.type":"java.lang.IllegalArgumentException","error.message":"unknown secure setting [_raw]
```

Enable / disable placement of these information using a flag or store it outside of the data section

benashz:
Add new transformation configurable `excludeRaw` to all syncable secret CRDs which controls whether or not the the `_raw` secret data is included in the destination secret.

Example:
```
spec:
  destination:
    transformation:
      excludeRaw: true
```

Adds a new commandline flag, `--global-rendering-options`, that takes a comma delimited set of options that provide control over global rendering options. The only option is `exclude-raw`.

```
  -global-rendering-options string
        Set global rendering options as a comma delimited string. Valid values are: [exclude-raw]
```

The global rendering options can also be set from Helm chart values:

```
--set 'controller.manager.globalRenderingOptions.excludeRaw=true'
```

Closes #424


rgarcia89:
Are there any negative implications with the exclude? If not, I wonder why the exclusion is not set to true by default.

benashz:
> Are there any negative implications with the exclude? If not, I wonder why the exclusion is not set to true by default.

@rgarcia89 not for VSO per-se but there could be for any user that is currently relying on this feature. I can see how having to configure this on all CRs could be burdensome, so it is probably worth making the feature opt-in somehow.

rgarcia89:
In that case it would be more attractive to make the default parameter configurable. Like for example the `defaultAuthMethod` for example.

benashz:
> In that case it would be more attractive to make the default parameter configurable. Like the `defaultAuthMethod` for example.

@rgarcia89 made this a command line configurable as well. See the updated description for more details.

rgarcia89:
@benashz this looks very good. Thanks!
I look very forward to the release

chart/crds/secrets.hashicorp.com_hcpvaultsecretsapps.yaml
```diff
@@ -169,6 +176,8 @@ spec:
                           - name
                           type: object
                         type: array
+                    required:
```
tomhjp, line 179: The behaviour is fine as it has a default, but from a generated documentation perspective this seems a little odd to be labelled as required. If you add `,omitempty` to the JSON tag will it stop labelling it as required?

docs/api/api-reference.md
```diff
@@ -318,6 +318,7 @@ _Appears in:_
 | `transformationRefs` _[TransformationRef](#transformationref) array_ | TransformationRefs contain references to template configuration from SecretTransformation |
 | `includes` _string array_ | Includes contains regex patterns of keys that should be included in the K8s Secret Data. FieldFilter can be used to filter the secret data that is stored in the K8s Secret Destination. Filters will not be applied to templated fields, those will always be included in the Destination K8s Secret. Exclusion filters are always applied first. |
 | `excludes` _string array_ | Excludes contains regex pattern for keys that should be excluded from the K8s Secret Data. FieldFilter can be used to filter the secret data that is stored in the K8s Secret Destination. Filters will not be applied to templated fields, those will always be included in the Destination K8s Secret. Exclusion filters are always applied first. |
+| `excludeRaw` _boolean_ | ExcludeRaw data from the destination Secret. Exclusion policy can be set globally by including 'exclude-raw` in the '--global-rendering-options' command line flag. The global policy always takes precedence over this configuration. |
```
tomhjp, line 321:
```suggestion
| `excludeRaw` _boolean_ | ExcludeRaw data from the destination Secret. Exclusion policy can be set globally by including 'exclude-raw` in the '--global-rendering-options' command line flag. If set, the command line flag always takes precedence over this configuration. |
```

internal/helpers/template.go
```diff
@@ -21,8 +21,14 @@ import (
 	"github.com/hashicorp/vault-secrets-operator/internal/template"
 )
 
-// regexCache provides a global LRU cache holding compiled regexes.
-var regexCache *lru.Cache[string, *regexp.Regexp]
+var (
+	// RenderOptionExcludeRaw sets the global sync option for controlling the exclusion
+	// of _raw from the destination secret.
+	// This is usually set from main via the command line arg --global-rendering-options
+	RenderOptionExcludeRaw bool
```
tomhjp, line 28: For this specific case, with the code as it currently is, this looks ok, but I would love to avoid mutable package-level variables as a general pattern. As we find more use-cases for passing flag config down through to templating functions, it would be great to establish a pattern that's more sustainable and safe. Maybe a stateful template helper struct that gets passed through to the clients that will consume it?

chart/crds/secrets.hashicorp.com_hcpvaultsecretsapps.yaml
benashz: Yes, it would be not be included in the required field set. I think we set the default to false to avoid some perpetual diff scenarios with Rancher.

docs/api/api-reference.md
benashz: This bit is auto generated from the structure member's doc strings. I can updated it there.

benashz: Updated in 8f2fa91c52b3e7b1b60a4f5b87c84868d35e166c

tomhjp (approved):

PR #551, opened 2024-01-17
Set CLI options from VSO_ environment variables

### LINKED ISSUE #533: OLM installation doesn't support setting flags for the operator
I want to install `vault-secret-operator` through OLM ([OperatorHub](https://developer.hashicorp.com/vault/docs/platform/k8s/vso/openshift#operatorhub)). I need to set some parameters available in the helm chart, such as [PersistenceModel](https://developer.hashicorp.com/vault/docs/platform/k8s/vso/helm#v-controller-manager-clientcache-persistencemodel). However, since it's configured as a [flag](https://github.com/hashicorp/vault-secrets-operator/blob/8a8c25953771637bc498dc4ebcceb100b15349f1/chart/templates/deployment.yaml#L68C16-L68C16) for the operator and operator doesn't support [`ARGS` as environment variable](https://github.com/operator-framework/operator-lifecycle-manager/blob/cee062283539d08f25333bfa53b2963725e35914/doc/contributors/design-proposals/operator-config.md#scenario-enable-debug-logs-for-an-operator), it seems that we can't set such configuration in the OLM mode.

Being able to set manager flags as environment variables. This can either be done via a separate env for every flag or supporting `ARGS` as an environment variable. 

Nothing specific.

**Additional context**

tvoran:
Adds support for setting some of the CLI options from environment variables, including:
- VSO_OUTPUT_FORMAT
- VSO_CLIENT_CACHE_SIZE
- VSO_CLIENT_CACHE_PERSISTENCE_MODEL
- VSO_MAX_CONCURRENT_RECONCILES

If both a command-line arg and environment variable are set for the same option, the environment variable takes precedence.

Also mentions the environment variables in the CLI flag help output.

Fixes #533


internal/options/env.go
```diff
@@ -0,0 +1,39 @@
...
+	ClientCachePersistenceModel string `split_words:"true"`
+
+	// MaxConcurrentReconciles is the VSO_MAX_CONCURRENT_RECONCILES environment variable option
+	MaxConcurrentReconciles *int `split_words:"true"`
+
+	// MaxConcurrentReconcilesVDS is the VSO_MAX_CONCURRENT_RECONCILES_VDS environment variable option
+	MaxConcurrentReconcilesVDS *int `split_words:"true"`
+
+	// MinRefreshAfterHVSA is the VSO_MIN_REFRESH_AFTER_HVSA environment variable option
+	MinRefreshAfterHVSA time.Duration `envconfig:"min_refresh_after_hvsa"`
```
benashz, line 33: We can probably exclude this option, since it is only meant for use in our test environments.

internal/options/env.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// ClientCachePersistenceModel is the VSO_CLIENT_CACHE_PERSISTENCE_MODEL
+	// environment variable option
+	ClientCachePersistenceModel string `split_words:"true"`
+
+	// MaxConcurrentReconciles is the VSO_MAX_CONCURRENT_RECONCILES environment variable option
+	MaxConcurrentReconciles *int `split_words:"true"`
+
+	// MaxConcurrentReconcilesVDS is the VSO_MAX_CONCURRENT_RECONCILES_VDS environment variable option
+	MaxConcurrentReconcilesVDS *int `split_words:"true"`
```
benashz, line 30: This one can be dropped as it is obsoleted by `max-concurrent-reconciles`

main.go
```diff
@@ -106,6 +108,32 @@ func main() {
...
 	flag.Parse()
 
+	// Parse environment variable options, prefixed with "VSO_"
+	if err := vsoEnvOptions.Parse(); err != nil {
+		os.Stderr.WriteString(fmt.Sprintf("Failed to process environment variable options: %q\n", err))
+		os.Exit(1)
+	}
+
+	// Set options from env if any are set
+	if vsoEnvOptions.OutputFormat != "" {
```
benashz, line 118: Should we document the env vars in the corresponding flags' usage?

benashz (approved):
Only a few suggestions to consider, otherwise looks great. I guess we will need to update the main VSO docs for these new env vars?

main.go
tvoran: Sure that makes sense. Added in e066367.

internal/options/env.go
tvoran: Gotcha, removed in e066367.

internal/options/env.go
tvoran: Cool, dropped in e066367.

tvoran:
> I guess we will need to update the main VSO docs for these new env vars?

Yep, absolutely, especially as it pertains to OperatorHub/OLM setups.

benashz (approved):
LGTM!

tvoran:
Thanks!

PR #552, opened 2024-01-18
Add support for userIDs in VaultPKISecret

czembower:
UID support in Vault core shipped in 1.13.x and was left out of the VSO implementation. I believe this addition makes VaultPKISecret exhaustive of all available pki/issue API parameters. This is my first PR for this project, apologies if I missed anything!


kschoche (commented):
Hi @czembower!
Thanks for filing this PR.

I think one connecting piece is missing [here](https://github.com/hashicorp/vault-secrets-operator/blob/0fd4c3bc66f92dbbcb21f93401a99a3222f14621/api/v1beta1/vaultpkisecret_types.go#L150):
```
diff --git a/api/v1beta1/vaultpkisecret_types.go b/api/v1beta1/vaultpkisecret_types.go
index 4f0fc7a..7fd3975 100644
--- a/api/v1beta1/vaultpkisecret_types.go
+++ b/api/v1beta1/vaultpkisecret_types.go
@@ -154,6 +154,7 @@ func (v *VaultPKISecret) GetIssuerAPIData() map[string]interface{} {
                "ip_sans":                 strings.Join(v.Spec.IPSans, ","),
                "uri_sans":                strings.Join(v.Spec.URISans, ","),
                "other_sans":              strings.Join(v.Spec.OtherSans, ","),
+               "user_ids":                strings.Join(v.Spec.UserIDs, ","),
                "ttl":                     v.Spec.TTL,
                "not_after":               v.Spec.NotAfter,
                "exclude_cn_from_sans":    v.Spec.ExcludeCNFromSans,
```

I'm not 100% sure of the PKI spec on whether or not including an empty string here will cause something to break, so we might have make it conditional on the spec supplying it?

I don't think we have any test coverage on these params actually working in the operator's tests which is why this was probably missed.

czembower:
I just committed a change that I think/hope addresses your concern!

benashz (approved):
LGTM!

PR #559, opened 2024-01-23
update api docs to include namespace requirements for auth components

kschoche:
Adds namespace requirements for vault auth components to clarify where the serviceaccount/secretRefs need to reside.


tvoran (approved):
Just some grammar nits 👍

PR #562, opened 2024-01-24
Reconcile on SecretTransformation updates.

benashz:
Allows for any changes made to a SecretTranretTransformation resource to result in a any referring syncable secret CRs to be "force" reconciled.

Each secrets controller watches for changes to the SecretTransformation. Upon update, all referring secret CR objects will be enqueued for reconciliation. In the case where `spec.destination.transformation.resync` is true, the secret will be synced using the updated SecretTransformation refs.


main.go
```diff
@@ -306,11 +309,13 @@ func main() {
...
-		Scheme:        mgr.GetScheme(),
-		Recorder:      mgr.GetEventRecorderFor("VaultDynamicSecret"),
-		ClientFactory: clientFactory,
-		HMACValidator: hmacValidator,
+		Client:         mgr.GetClient(),
+		Scheme:         mgr.GetScheme(),
+		Recorder:       mgr.GetEventRecorderFor("VaultDynamicSecret"),
+		ClientFactory:  clientFactory,
+		HMACValidator:  hmacValidator,
+		SyncRegistry:   controllers.NewSyncRegistry(),
```
kschoche, line 317: My one minor concern while reading through the code was that it looked like we were creating some lock contention on the `ResourceReferenceCache` between the various controllers which all run in parallel, but I'm happy to see that you thought ahead and are creating a separate cache for each controller!

kschoche (approved):
Excellent work!

benashz:
Thanks!

PR #563, opened 2024-01-25
WIP: watch for secret deletions

benashz:
Previously each secret controller could only detect the deletion of a destination K8s Secret that it owned on the next call to `Reconcile()`. This PR adds support for watching K8s secrets that it owns. Upon deletion, the owner resource will be enqueued for reconciliation with a random delay.


benashz:
Closing in favor of #584

PR #567, opened 2024-01-26
demo: add cross Vault namespace example

benashz:
Updates the demo to include an example of deploying VSO resources with cross Vault namespace authentication support. Secrets are synced from a secrets engine mounted in a separate Vault namespace.

Setup the demo with the following command:

```
make -f demo.mk VAULT_ENTERPRISE=true docker-build demo
```


benashz:
Closing in favor of #568

PR #568, opened 2024-01-26
demo: add cross Vault namespace example

benashz:
Updates the demo to include an example of deploying VSO resources with cross Vault namespace authentication support. Secrets are synced from a secrets engine mounted in a separate Vault namespace.

Setup the demo with the following command:

```
make -f demo.mk VAULT_ENTERPRISE=true docker-build demo
```

### DIFF
--- a/demo.mk
+++ b/demo.mk
@@ -36,6 +36,7 @@ demo-destroy: demo-delete-kind ## delete the kind cluster
 .PHONY: demo-infra-vault
 demo-infra-vault: ## Deploy Vault for the demo
 	$(MAKE) setup-vault \
+		VAULT_ENTERPRISE=$(VAULT_ENTERPRISE) \
 		TF_VAULT_STATE_DIR=$(TF_VAULT_STATE_DIR) \
 		TF_INFRA_STATE_DIR=$(TF_VAULT_STATE_DIR) \
 		K8S_CLUSTER_CONTEXT=$(K8S_CLUSTER_CONTEXT)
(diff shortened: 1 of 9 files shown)


demo.mk
```diff
@@ -53,6 +54,19 @@ demo-infra-app: demo-setup-kind ## Deploy Postgres for the demo
 		-var k8s_config_context=$(K8S_CLUSTER_CONTEXT) \
 		$(EXTRA_VARS) || exit 1 \
 
+.PHONY: demo-infra-app-plan
+demo-infra-app-plan: demo-setup-kind ## Deploy Postgres for the demo
+	@mkdir -p $(TF_APP_STATE_DIR)
+	rm -f $(TF_APP_STATE_DIR)/*.tf
+	cp $(DEMO_ROOT)/infra/app/*.tf $(TF_APP_STATE_DIR)/.
+	$(TERRAFORM) -chdir=$(TF_APP_STATE_DIR) init -upgrade
+	$(TERRAFORM) -chdir=$(TF_APP_STATE_DIR) plan \
```
kschoche, line 63:
```suggestion
	$(TERRAFORM) -chdir=$(TF_APP_STATE_DIR) apply \
```

I think you meant apply here?

demo/infra/app/cross-vault-ns.tf
```diff
@@ -0,0 +1,263 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+# demonstrates how to setup cross Vault namespace for VSO and K8s auth when var.vault_enterprise = true
+
+# sets the group policy application to work with with cross Vault namespace auth
+resource "vault_generic_endpoint" "sys-group-policy-application" {
+  count = var.vault_enterprise ? 1 : 0
```
kschoche, line 8: The count on just this resource is a little strange, should this test only run against vault enterprise?
No change required, just wondering about its purpose?

kschoche (approved):
LGTM, and glad this works as we expected :D 

I left one question and possibly one bug fix in the makefile.

tomcf-hcp:
Should we make any updates to demo/readme?

demo/infra/app/cross-vault-ns.tf
benashz: We only want to configure this policy if the demo is for Vault enterprise.

demo.mk
benashz: Pretty sure it needs to be plan, as this is a new make target:
```
demo-infra-app-plan
```

We already have apply in the `demo-infra-app` target

kschoche: Ah I forgot about that target, makes sense, thanks!

benashz:
> Should we make any updates to demo/readme?

Added in ff033f3b1b1670fcf66b9933990520dec29d2e6b

PR #574, opened 2024-01-29
ci: bump max tries for HCPVSA tests

benashz:
Pass a fairly large value for maxTries to
assertRemediationOnDestinationDeletion() to help mitigate the effects of the HVS' request rate limiter. Typically, the rate limiter is only ever triggered when there are four or more GH workflows running concurrently.


kschoche (approved):
LGTM

PR #588, opened 2024-02-09
API: resync is no longer needed.

benashz:
Removes obsolete `spec.destination.transformation.resync` field.


tomhjp (commented):
What's the upgrade story for this change if people have it set in their existing CRs and then update their CRDs?

benashz:
> What's the upgrade story for this change if people have it set in their existing CRs and then update their CRDs?

Hasn't been shipped yet, so we should be good.

tomhjp (approved):
Ah of course, thanks 👍

PR #589, opened 2024-02-09
Fix: include global templates only on empty templateRefs

benashz:
Ensure that a transformationRef's template overrides are honored.


thyton (approved):

benashz:
Thanks!

PR #590, opened 2024-02-09
Install the operator-sdk CLI and check `sdk-generate` in CI

tvoran:
Adds `sdk-generate` to `make manifests`, to ensure the custom resource definitions in the base ClusterServiceVersion are up to date and are checked in the build-pre-checks CI job. Also adds `HCPAuth` and `HCPVaultSecretsApp` to `PROJECT` along with the generated parts in the ClusterServiceVersion base.


benashz (approved):
LGTM!

Makefile
```diff
@@ -591,6 +592,23 @@ copywrite: ## Download copywrite locally if necessary.
 yq: ## Download yq locally if necessary.
 	@./hack/install_yq.sh
 
+.PHONY: operator-sdk
```
benashz, line 595: Looks good! I wonder if we could maybe update it to like the `yq` make target. That will ensure that we have expected version of the sdk, as well as guard against any GH rate-limiters. Might be worth doing now.

PROJECT
```diff
@@ -74,6 +74,24 @@ resources:
   kind: VaultTransit
   path: github.com/hashicorp/vault-secrets-operator/api/v1beta1
   version: v1beta1
+- api:
```
benashz, line 77: :+1:

benashz (commented):
Looks good! I just had one comment/suggestion about fetching the operator-sdk.

Makefile
```diff
@@ -507,20 +508,20 @@ set-image-ubi: kustomize copy-config ## Set the controller UBI image
...
 .PHONY: sdk-generate
-sdk-generate: copywrite
-	operator-sdk generate kustomize manifests -q
+sdk-generate: copywrite operator-sdk
+	$(OPERATOR_SDK) generate kustomize manifests -q
 	@$(COPYWRITE) headers &> /dev/null
 
 .PHONY: bundle
-bundle: manifests kustomize sdk-generate set-image-ubi yq ## Generate bundle manifests and metadata, then validate generated files.
+bundle: manifests kustomize sdk-generate set-image-ubi yq operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
```
thyton, line 516: This is out of my curiosity, not my actual code review :). 

I see `operator-sdk` is added to `sdk-generate`, and `sdk-generate` is added to `manifests`. I'm curious how `manifests` in `bundle` is executed differently that `sdk-generate` and `operator-sdk` remain in `bundle` (i.e why we don't include only `manifests` calling `sdk-generate` which also calls `operator-sdk`).

Makefile
tvoran: Sure, changed in 4ceb1e2

benashz: Thanks!

benashz:
LGTM!

PR #601, opened 2024-02-13
Chart: support setting imagePullPolicy

benashz:
Add support for setting the imagePullPolicy for the `manager` and `kube-rbac-proxy` containers. The default for both is `IfNotPresent`


thyton (approved):

benashz:
Thanks!

PR #605, opened 2024-02-15
Properly handle when the secret type changes

benashz:
The corev1.Secret.Type is an immutable field and cannot be updated, so the destination secret needs to be recreated whenever spec.destination.type changes.


internal/helpers/secrets.go
```diff
@@ -251,20 +251,46 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
...
+			}
+		} else {
+			dest.Type = secretType
+			logger.V(consts.LogLevelDebug).Info("Updating secret")
+			if err := client.Update(ctx, dest); err != nil {
+				return err
+			}
 		}
 	} else {
+		dest.Type = secretType
```
tvoran, line 293: Same question here since it looks like this was set on line 257, and wouldn't have changed yet?

internal/helpers/secrets.go
```diff
@@ -251,20 +251,46 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
...
+			}
+
+			dest.ResourceVersion = ""
+			dest.Generation = 0
+			dest.SetLabels(labels)
+			if err := client.Create(ctx, dest); err != nil {
+				return err
+			}
+		} else {
+			dest.Type = secretType
```
tvoran, line 286: Isn't this already set on 257?

internal/helpers/secrets.go
```diff
@@ -251,20 +251,46 @@ func SyncSecret(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Ob
...
+			}
+
+			// delete the secret
+			if err := client.Delete(ctx, dest); err != nil {
+				return err
+			}
+
+			dest.ResourceVersion = ""
+			dest.Generation = 0
+			dest.SetLabels(labels)
```
tvoran, line 281: May want to clear UID and CreationTimestamp too? Though I imagine those would both be overwritten by k8s.

tvoran (approved):

internal/helpers/secrets.go
benashz: It seems like those fields get created, so I think you are correct about them being overwritten.

```
 $ diff -u /tmp/before.yaml /tmp/after.yaml
--- /tmp/before.yaml    2024-02-15 14:26:40
+++ /tmp/after.yaml     2024-02-15 14:27:03
@@ -3,7 +3,7 @@
   .dockerconfigjson:  .....
 kind: Secret
 metadata:
-  creationTimestamp: "2024-02-15T19:26:34Z"
+  creationTimestamp: "2024-02-15T19:26:56Z"
   labels:
     app.kubernetes.io/component: secret-sync
     app.kubernetes.io/managed-by: hashicorp-vso
@@ -18,6 +18,6 @@
     kind: HCPVaultSecretsApp
     name: image-pull-secrets
     uid: f054ffbd-a2c4-4f2b-afc5-b32585b098fb
-  resourceVersion: "348003"
-  uid: 2a61b478-a07d-4735-a938-46e8abf38d37
-type: Opaque
+  resourceVersion: "348051"
+  uid: be705697-2d7a-4e95-b363-0a366fa7975d
+type: kubernetes.io/dockerconfigjson
```

internal/helpers/secrets.go
benashz: Good catch. Fixed in 5b74942ee405ee3c18198f44550567ce13cc57ee. I also noticed that the fake client does not seem to enforce immutable field constraints. Fixed in the same commit.

internal/helpers/secrets.go
benashz: Fixed in 5b74942ee405ee3c18198f44550567ce13cc57ee

PR #606, opened 2024-02-15
Build: retry GH fetch on transient errors

benashz:
Makes fetching build tools from GH a bit more robust by retrying on transient errors. 

Add support for setting curl's `--retry` option which will retry on any of the following HTTP response codes: 408, 429, 500, 502, 503, 504

Example of a related build failure:
https://github.com/hashicorp/vault-secrets-operator/actions/runs/7922121731/job/21629102539#step:6:19


thyton (approved):

thyton:
LGTM!

Issue #608, opened 2024-02-16
Secret sync fails after upgrading directly from v0.2.0 to v0.5.0


sebglon:
**Describe the bug**
After upgrading CRD and pod to 0.5.0, i have a sheme issue on VaultDynamicSecret:
```
VaultStaticSecret/monitoring/xxx-keys dry-run failed, reason: Invalid, error: VaultStaticSecret.secrets.hashicorp.com "xxx-keys" is invalid: status.lastGeneration: Required value
```

**Expected behavior**
Upgrade CRD
Upgrade helm chart to 0.5.0
Try to re-deploy existing resource

**Environment**
* Kubernetes version:  1.22.2
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): onPrem
  * Other configuration options or runtime services (istio, etc.):
* vault-secrets-operator version: v0.5.0

**Additional context**

benashz:
So I was able to reproduce the issue when upgrading directly from v0.2.0 -> v0.5.0. The VDS sync failure was caused by an attempt to add the finalizers to object being reconciled, before its status had been updated to match the new schema.

The workaround for now is to upgrade to v0.4.3 before going to v0.5.0. We will be releasing a fix that ensures that an upgrade from v0.2.0 to v0.5.x works seamlessly.

sebglon:
I tried to upgrade to v0.4.3 before v0.5.0 from v0.2.0 and i have the same issue

benashz:
> I tried to upgrade to v0.4.3 before v0.5.0 from v0.2.0 and i have the same issue

Would that be for the VaultStaticSecret or for VaultDynamicSecret? I believe for the VDS case, the incremental upgrade will work if `spec.revoke` is set. For the VSS, case the sync should succeed since VSO is not adding any finalizers.

We will be releasing v0.5.1 today, so there shouldn't be a need to do the incremental upgrades.

benashz:
Fixed in [v0.5.1](https://github.com/hashicorp/vault-secrets-operator/releases/tag/v0.5.1)

sebglon:
Always the same issue with v0.5.1
on dynamic and static

```
[...]

PR #609, opened 2024-02-16
Always add finalizers after status updates

benashz:
Make updating a resource's status more consistent for all secret controllers, by factoring out the Status.Update() calls to controller methods, and always setting the lastGeneraton from the reconciled object.

The root cause of #608 seems to be API schema differences when going directly from v0.2.0 to v0.5.0. In v0.5.0 we added finalizers the VPS, VSS, and HCPVSA controllers. Finalizers were being added before the object's status was updated, unfortunately that would result in a error that would prevent the secret from ever being  synced, and the object's status from ever being updated. The workaround is documented here: https://github.com/hashicorp/vault-secrets-operator/issues/608#issuecomment-1953010710

The solution is to add the finalizer after the status update. All controllers that have finalizers have been updated accordingly.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -250,6 +250,7 @@ func maybeAddFinalizer(ctx context.Context, c client.Client, o client.Object, fi
 		controllerutil.AddFinalizer(o, finalizer)
 		if err := c.Update(ctx, o); err != nil {
 			logger.Error(err, "Failed to add finalizer")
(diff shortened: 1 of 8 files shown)


thyton (approved):

thyton:
LGTM!

thyton (approved):
LGTM 👍

PR #626, opened 2024-02-27
chart values: mention global transformation option env variable

tvoran:
Also remove the docstring type for
`controller.manager.globalTransformationOptions` since it's not actually used in the chart as an array.


thyton (approved):

tvoran:
Thanks!

PR #629, opened 2024-02-29
Set VaultConnection headers on the vault client

### LINKED ISSUE #622: VSO failed to connect through CloudFlare Zero Trust
**Describe the bug**
We have our vault protected by CloudFlare Zero Trust. I have a client_id and client_secret tokens that I need to pass to go through the ZT. When using the CLI locally, the tokens works fine (using `vault login -header=cf-client=xxx ...`).
When adding those information as headers in the default vaultConnection, I get a 403:
```
Failed to check Vault seal status: Error making API request. URL: GET https://example.cloudflareaccess.com/cdn-cgi/access/login/vault.example.com Code: 403. Raw Message: <html> <head><title>403 Forbidden</title></head> <body> <center><h1>403 Forbidden</h1></center> <hr><center>cloudflare</center> </body> </html>
```

We're using helm charts with following info:
* vault-secrets-operator version: 0.4.3
* vault version: 0.27.0

The values.yaml (for the operator):
```
operator:
  enabled: true
  controller:
    replicas: 1
  defaultVaultConnection:
    enabled: true
    address: "https://vault.example.com"
    skipTLSVerify: false
    headers:
      CF-Access-Client-Id: "1f..."
      CF-Access-Client-Secret: "58..."
  defaultAuthMethod:
    enabled: false
```

When checking in my K8s cluster the vaultConnection resource, I can see the headers:
```
kind: VaultConnection
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: > {longobject}
  finalizers:
    - vaultconnection.secrets.hashicorp.com/finalizer
  generation: 4
  labels:
    app.kubernetes.io/instance: vault-dev
    app.kubernetes.io/name: operator
    app.kubernetes.io/version: 0.4.3
    argocd.argoproj.io/instance: vault-dev
    component: controller-manager
    control-plane: controller-manager
    helm.sh/chart: operator-0.4.3
  name: default
  namespace: vault
spec:
  address: 'https://vault.example.com'
  headers:
    CF-Access-Client-Id: 1f...
    CF-Access-Client-Secret: 58...
  skipTLSVerify: false
status:
  valid: true
```

**Expected behaviour**
The vault-secrets-operator will use the headers to bypass CloudFlare Zero Trust

**Environment**
* Kubernetes version: 
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): EKS, version 1.28

Is there something that I miss here? I haven't been able to find a good documentation on this specific setup =/

Also, To make it work, do I need to update additional resources (VaultAuth, VaultStaticSecret, ...)?
Thanks 🙏

tvoran:
Set Vault client headers from VaultConnection.Spec.Headers.

Fixes #622


thyton (approved):

tomhjp (approved):
LGTM

internal/vault/config.go
```diff
@@ -96,6 +98,9 @@ func MakeVaultClient(ctx context.Context, cfg *ClientConfig, client ctrlclient.C
 	if cfg.VaultNamespace != "" {
 		c.SetNamespace(cfg.VaultNamespace)
 	}
+	for k, v := range cfg.Headers {
```
benashz, line 101: I think we should move the custom header setting to come before line 98. That will ensure that `spec.headers` does not override `cfg.VaultNamespace` by setting `X-Vault-Namespace`.

internal/vault/config_test.go
```diff
@@ -72,6 +72,17 @@ func TestMakeVaultClient(t *testing.T) {
 			CACert:        nil,
 			expectedError: nil,
 		},
+		"headers": {
```
benashz, line 75: May want to add a test to ensure proper handling of the Vault namespace config e.g. `cfg.VaultNamespace` takes precedence over `X-Vault-Namespace`

benashz (approved):
Looking good. I added a suggestion to ensure that the Vault namespace from config always takes precedence over. We may also want to warn in the case where `X-Vault-Namespace` is being set in the custom headers. We could also make setting that header illegal.

internal/vault/config.go
tvoran: Sure, also made setting the namespace in VaultConnection an error condition in [10af3b4](https://github.com/hashicorp/vault-secrets-operator/pull/629/commits/10af3b436e5eecb59f858ab6fb1156e3493e936e)

internal/vault/config.go
```diff
@@ -93,6 +96,12 @@ func MakeVaultClient(ctx context.Context, cfg *ClientConfig, client ctrlclient.C
 		l.Error(err, "error setting up Vault API client")
 		return nil, err
 	}
+	if _, exists := cfg.Headers[vconsts.NamespaceHeaderName]; exists {
```
benashz, line 99: LGTM!

internal/vault/config_test.go
```diff
@@ -116,7 +139,25 @@ func TestMakeVaultClient(t *testing.T) {
...
+
+func makeVaultHttpHeaders(t *testing.T, namespace string, headers map[string]string) http.Header {
+	t.Helper()
+
+	h := make(http.Header)
+	for k, v := range headers {
+		h.Set(k, v)
+	}
+	h.Set("X-Vault-Request", "true")
+	if len(namespace) > 0 {
```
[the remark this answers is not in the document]
tvoran: [Yep](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/150#discussion_r860770659)

tvoran:
Thanks!

PR #636, opened 2024-03-04
VPS: return an error when the PKI secret is nil

benashz:


internal/vault/vault.go
```diff
@@ -20,6 +21,10 @@ type PKICertResponse struct {
 }
 
 func UnmarshalPKIIssueResponse(resp *api.Secret) (*PKICertResponse, error) {
+	if resp == nil {
+		return nil, fmt.Errorf("vault secret is nil")
```
tvoran, line 25: Just a thought:

```suggestion
		return nil, fmt.Errorf("vault secret response is nil")
```

tvoran (approved):

internal/vault/vault.go
benashz: Updated in 72eb03add62aa7cad8a77b4f2bbfdfd8318c4c4b

Issue #635, opened 2024-03-04
Unable to set up VaultAuthList


monicabiswal147:
Trying to set up the VaultAuthList using the API reference - https://developer.hashicorp.com/vault/docs/platform/k8s/vso/api-reference#vaultauthlist. My YAML for the same is as below:
`apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultAuthList
metadata:
  name: vault-auth-list
spec:
  items:
  - apiVersion: secrets.hashicorp.com/v1beta1
    kind: VaultAuth
    metadata:
      name: vault-auth
      namespace: sit
    spec:
      vaultConnectionRef: vso-connection
      method: kubernetes
      mount: sit
      kubernetes:
        role: sit-kubernetes
        serviceAccount: sit-sa`

I get an error - the server could not find the requested resource (post vaultauthlists.secrets.hashicorp.com)

tvoran:
Hi @monicabiswal147, the VaultAuthList type in our API reference docs is just the concrete kind used when getting a [collection of VaultAuth's](https://kubernetes.io/docs/reference/using-api/api-concepts/#collections) from the Kubernetes API. You'll also see it as the `listKind` in the CRD definition https://github.com/hashicorp/vault-secrets-operator/blob/8da51b765131b29307e712e046ec243e5e6d2a4c/chart/crds/secrets.hashicorp.com_vaultauths.yaml#L15

So in your case you'd want to create a separate VaultAuth for each item in the list.

I can certainly understand the confusion here, so we may want to omit the List types from the website docs, since they aren't really useful for creating things.

PR #637, opened 2024-03-05
Remove List types from API reference

tvoran:
Set List types to be ignored in docs/api/config.yaml and ran `make gen-api-ref-docs`.

Just thinking this could make the API reference docs more clear as to which resources can be created.

Related to #635


tomhjp (approved):
Makes sense to me 👍 one alternative I considered was whether we could make efforts to mark them as some form of read-only, but I think this way will cause less confusion for users.

benashz:
I wonder if updating documentation for the the *LIst* types to say that they are for read-only operations would be a better way to go? That way our API reference would still be complete.

PR #641, opened 2024-03-06
VDS: support configuring an explicit sync delay

benashz:
There are some secrets engines that do not provide a lease TTL that the VDS reconciler can use to compute a VDS instances enqueue horizon. This change adds the ability to set a sync duration explicity via the new configuration option: spec.refreshAfter. In the case where the lease TTL is known, spec.refreshAfter will be ignored if set.


thyton (approved):
LGTM 👍

Issue #631, opened 2024-03-01
Breaking changes to v1beta1 VaultStaticSecret and VaultConnection CRDs cause failed upgrades after v0.5.0


kennedn:
**Describe the bug**
Additional required fields have been added to the VaultConnection and VaultStaticSecret CRDS without bumping the api version:

https://github.com/hashicorp/vault-secrets-operator/blob/6de064379dc95617133056aca00213453a04f07b/chart/crds/secrets.hashicorp.com_vaultconnections.yaml#L59-L61

https://github.com/hashicorp/vault-secrets-operator/blob/6de064379dc95617133056aca00213453a04f07b/chart/crds/secrets.hashicorp.com_vaultstaticsecrets.yaml#L178-L181

The new fields have default values but defaults are only applied when new instances of the CRDs are created. This means any in-flight CRD instances that exist in a given cluster will now break the operator and block the upgrade:

**To Reproduce**
1. Deploy a version of VSO older than v0.5.0
2. Deploy a VaultConnection or VaultStaticSecrets CRD
3. Upgrade the VSO operator past v0.5.0
6. Operator fails to upgrade:

![MicrosoftTeams-image](https://github.com/hashicorp/vault-secrets-operator/assets/8060657/d04dcb01-6843-4813-9bb5-a9a4a84c8f54)

Application deployment:

```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultConnection
metadata:
  name: vault-connection-test
  namespace: default
spec:
  address: https://some-vault-url.com
  caCertSecretRef: some--ca-cert
---
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultStaticSecret
metadata:
  namespace: default
  name: vault-static-secret-test
spec:
  vaultAuthRef: vault-auth-test
  mount: some-namespace/secret
  type: kv-v2
  path: sample
  refreshAfter: 60s
  destination:
    create: true
    name: vault-static-secret-test
```

**Expected behavior**
Operator upgrades successfully

**Environment**

* Kubernetes version: 
* OpenShift 4.14
* 
* vault-secrets-operator version:
* v0.5.1

kennedn:
The workaround used to sidestep this bug was to backup all affected CRD instances, delete them in cluster, upgrade the operator, then reapply the backups. The missing default values are then applied upon creation of the CRD instances:

Backup affected CRD instances:
```shell
for crd in vaultconnection vaultstaticsecret; do
[...]

PR #645, opened 2024-03-07
Update the changelog with a known issue in v0.5.0

tvoran:
Just to warn folks that there may be an issue when upgrading to v0.5.0 via OperatorHub on OpenShift, with a link to [the issue discussing a workaround](https://github.com/hashicorp/vault-secrets-operator/issues/631).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Dependency Updates:
 
 ## 0.5.0 (February 15th, 2024)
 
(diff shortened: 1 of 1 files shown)


benashz:
Would you mind adding this to the the release docs as well: https://github.com/hashicorp/vault-secrets-operator/releases/edit/v0.5.0

tvoran:
@benashz Sure, will do.

PR #650, opened 2024-03-12
Make some recently required bool parameters optional

### LINKED ISSUE #631: Breaking changes to v1beta1 VaultStaticSecret and VaultConnection CRDs cause failed upgrades after v0.5.0

tvoran:
(Just putting this up as an option to address #631 if we wish.)

Makes some currently required `bool` parameters optional, but keeps the default of `false`, so that the value can effectively be set to `false` in k8s storage. The secret destination options were added in v0.5.0.

I was able to duplicate the upgrade error in #631 on OpenShift outside of OperatorHub with `operator-sdk run bundle...` and `operator-sdk run bundle-upgrade...`, which would fail when going from v0.4.3 to v0.5.x, but succeed when going from v0.4.3 to these changes.

<details>
  <summary>Details for posterity:</summary>

```bash
$ oc new-project vault-secrets-operator

$ operator-sdk run bundle -n vault-secrets-operator registry.connect.redhat.com/hashicorp/vault-secrets-operator-bundle:0.4.3

# deploy a VaultStaticSecret setup
...

# upgrade the operator
$ operator-sdk run bundle-upgrade -n vault-secrets-operator registry.connect.redhat.com/hashicorp/vault-secrets-operator-bundle:0.5.1

# ^ this will eventually timeout, with an error like this in the status of the subscription:
$ kubectl get subscription -n vault-secrets-operator vault-secrets-operator-v0-4-3-sub -o yaml
...
    message: 'error validating existing CRs against new CRD''s schema for "vaultstaticsecrets.secrets.hashicorp.com":
      error validating custom resource against new schema for VaultStaticSecret tenant-1/static-demo:
      [].spec.destination.overwrite: Required value'
...

# Build a bundle from this branch
$ export PREVIOUS_VERSION=v0.5.1
$ export TEST_VERSION=0.5.2-test-olm-upgrade
$ make ci-build ci-docker-build-ubi VERSION=${TEST_VERSION?}
$ docker tag hashicorp/vault-secrets-operator:${TEST_VERSION?}-ubi <docker-repo>:${TEST_VERSION?}-ubi
$ docker push <docker-repo>:${TEST_VERSION?}-ubi

$ make bundle IMG=<docker-repo>:${TEST_VERSION?} IMG_UBI=<docker-repo>:${TEST_VERSION?}-ubi VERSION=${TEST_VERSION?} USE_IMAGE_DIGESTS=true
$ make bundle-build BUNDLE_IMG=<docker-repo>-bundle:${TEST_VERSION?}
$ docker push <docker-repo>-bundle:${TEST_VERSION?}

# start over from the beginning, but do the bundle-upgrade with this branch, and the upgrade passes
...
$ operator-sdk run bundle-upgrade -n vault-secrets-operator docker.io/<docker-repo>-bundle:0.5.2-test-olm-upgrade
```

</details>

Closes #631

### DIFF
--- a/api/v1beta1/vaultconnection_types.go
+++ b/api/v1beta1/vaultconnection_types.go
@@ -22,7 +22,7 @@ type VaultConnectionSpec struct {
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 	// SkipTLSVerify for TLS connections.
 	// +kubebuilder:default=false
-	SkipTLSVerify bool `json:"skipTLSVerify"`
+	SkipTLSVerify bool `json:"skipTLSVerify,omitempty"`
 }
 
 // VaultConnectionStatus defines the observed state of VaultConnection
--- a/api/v1beta1/vaultconnection_types.go
+++ b/api/v1beta1/vaultconnection_types.go
@@ -22,7 +22,7 @@ type VaultConnectionSpec struct {
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 	// SkipTLSVerify for TLS connections.
 	// +kubebuilder:default=false
-	SkipTLSVerify bool `json:"skipTLSVerify,omitempty"`
+	SkipTLSVerify bool `json:"skipTLSVerify"`
 }
 
 // VaultConnectionStatus defines the observed state of VaultConnection
--- a/config/manifests/bases/vault-secrets-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/vault-secrets-operator.clusterserviceversion.yaml
@@ -168,4 +168,8 @@ spec:
   provider:
     name: HashiCorp
     url: https://www.hashicorp.com/
+  skips:
+  - vault-secrets-operator.v0.4.3
+  - vault-secrets-operator.v0.5.0
+  - vault-secrets-operator.v0.5.1
   version: 0.0.0-dev
--- a/config/manifests/bases/vault-secrets-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/vault-secrets-operator.clusterserviceversion.yaml
@@ -169,7 +169,6 @@ spec:
     name: HashiCorp
     url: https://www.hashicorp.com/
   skips:
-  - vault-secrets-operator.v0.4.3
   - vault-secrets-operator.v0.5.0
   - vault-secrets-operator.v0.5.1
   version: 0.0.0-dev
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -15,11 +15,11 @@ type Destination struct {
 	// Create the destination Secret.
 	// If the Secret already exists this should be set to false.
 	// +kubebuilder:default=false
-	Create bool `json:"create"`
+	Create bool `json:"create,omitempty"`
 	// Overwrite the destination Secret if it exists and Create is true. This is
 	// useful when migrating to VSO from a previous secret deployment strategy.
 	// +kubebuilder:default=false
-	Overwrite bool `json:"overwrite"`
+	Overwrite bool `json:"overwrite,omitempty"`
 	// Labels to apply to the Secret. Requires Create to be set to true.
 	Labels map[string]string `json:"labels,omitempty"`
 	// Annotations to apply to the Secret. Requires Create to be set to true.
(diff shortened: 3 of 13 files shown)


tomhjp (approved):
I'm supportive of this approach 👍 it always looked a bit wrong to me that these parameters were marked as required

api/v1beta1/vaultconnection_types.go
```diff
@@ -22,7 +22,7 @@ type VaultConnectionSpec struct {
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 	// SkipTLSVerify for TLS connections.
 	// +kubebuilder:default=false
-	SkipTLSVerify bool `json:"skipTLSVerify"`
+	SkipTLSVerify bool `json:"skipTLSVerify,omitempty"`
```
benashz, line 25: I think this one we want to keep as-is. We made this change under #527 - to avoid an issue with Rancher.

benashz (approved):
We should revert the change made to SkipTLSVerify, after that +1

benashz:
> I'm supportive of this approach 👍 it always looked a bit wrong to me that these parameters were marked as required

@tomhjp - yeah its unfortunate that fields with defaults are not excluded from the set of required fields, or that the validators do not take defaults into account.

api/v1beta1/vaultconnection_types.go
tvoran: We can certainly do that, especially since this has been in place since v0.4.3.

FWIW in my testing this behaves the same with or without `omitempty` when there's a default set. If the parameter is set to true/false in the CRD spec it comes back on a get. And if the parameter is unspecified it comes back as `false` on a get. Which I think satisfies #527 and #278.

config/manifests/bases/vault-secrets-operator.clusterserviceversion.yaml
```diff
@@ -168,4 +168,7 @@ spec:
   provider:
     name: HashiCorp
     url: https://www.hashicorp.com/
+  skips:
+  - vault-secrets-operator.v0.5.0
+  - vault-secrets-operator.v0.5.1
```
tvoran, line 173: Because of these I think we'll want to use PREVIOUS_VERSION=v0.4.3 with `make bundle`:

```
make bundle VERSION=${LATEST_TAG//v} USE_IMAGE_DIGESTS=true PREVIOUS_VERSION=v0.4.3
```

So that the resulting `build/bundle/manifests/vault-secrets-operator.clusterserviceversion.yaml` comes out like this:

```yaml
  skips:
    - vault-secrets-operator.v0.5.0
    - vault-secrets-operator.v0.5.1
  version: 0.5.2
  replaces: vault-secrets-operator.v0.4.3
```

PR #651, opened 2024-03-13
use go 1.21.8

tvoran:
We should probably use 1.21.8 or 1.22.1 here now: https://groups.google.com/g/golang-announce/c/5pwGVUPoMbg


tomhjp (approved):
LGTM - is the loop var fix the reason you didn't choose 1.22, or did you want to minimise risk for patch VSO releases? If it's the former, we can gate the behaviour change separately because it depends on a go.mod change: https://go.dev/blog/loopvar-preview

benashz (approved):
LGTM - I'm in favor of doing the full bump to go 1.22 in v0.6.0 since v0.5.2 is going out today. The version bump would include updating the version in go.mod.

PR #659, opened 2024-03-17
Fix: Overwriting tls.crt and tls.key when using secret type kubernetes.io/tls for VaultPKISecret (#658)

### LINKED ISSUE #658: Overwriting tls.crt and tls.key when using secret type kubernetes.io/tls for VaultPKISecret
**Describe the bug**
Currently it's not possible to create secret type kubernetes.io/tls and overwrite tls.key or tls.crt fields with transformations:

**To Reproduce**
1. Deploy some VaultPKISecret
```yaml
kind: VaultPKISecret
metadata:
  name: ingress-certs
  namespace: openshift-ingress
spec:
  destination:
    type: kubernetes.io/tls # - When using this type, vso ignores transformation on `tls.crt` field
    create: true
    name: ingress-certs
    overwrite: true
    # "\n" is required at the end of tls.crt - https://access.redhat.com/solutions/6984698
    transformation:
      templates:
        tls.crt: # this doesn't work
          text: |
            {{- printf "%s\n%s\n" (get .Secrets "certificate") (get .Secrets "issuing_ca") -}}
        tls-crt: # this works
          text: |
            {{- printf "%s\n%s\n" (get .Secrets "certificate") (get .Secrets "issuing_ca") -}}
  mount: pki_int
  role: ...
  commonName: "..."
  altNames: ["..."]
  ttl: 17520h
  format: pem
  expiryOffset: 720h
  vaultAuthRef: default
```

Application deployment:

```yaml
vault-secrets-operator:
  defaultVaultConnection:
    # toggles the deployment of the VaultAuthMethod CR
    enabled: true

    address: ...
    skipTLSVerify: false
    caCertSecret: root-ca

  controller:
    manager:
      resources:
        limits:
          cpu: 500m
          memory: 1024Mi

    kubeRbacProxy:
      resources:
        limits:
          cpu: 500m
          memory: 512Mi
```

**Expected behavior**
The ability to modify the tls.crt field in a Secret depends on the Secret's type. It works in the Opaque type but doesn't work in the kubernetes.io/tls type.

**Environment**
* OpenShift (4.12) v1.25.16+6df2177
* vault-secrets-operator version: 0.5.2

**Additional context**
RedHat OpenShift requires that tls.crt ends with newline.

- https://docs.openshift.com/container-platform/4.15/security/certificates/replacing-default-ingress-certificate.html
- https://access.redhat.com/solutions/6984698 (paywall)
image

jacob-faber:
Fixes #658


benashz:
Thanks for the contribution @jacob-faber - I would be more in favour of making this change a bit simpler. We could just say that if there are any transformations defined for the `destination` then skip the hard coded data rendering.

jacob-faber:
@benashz, Do you think https://github.com/hashicorp/vault-secrets-operator/blob/33c087f40dd4590fa76e14b78684c6134bf1e204/controllers/vaultpkisecret_controller.go#L232-L234 should be skipped too?

benashz (approved):
LGTM!

PR #685, opened 2024-04-11
Client factory: support client taints

benashz:
The secret reconcilers have no way of advising the ClientFactory that a Client might be invalid. Now a reconciler can taint the Client if a Vault operation fails for some reason. The common case is where a Vault request resulted in a 403 (forbidden) status code. In this case the reconciler can taint the client so that the next call to factory for the tainted client will have the factory validate that the client's token is still valid by performing reaching out to Vault's lookup lookup API. Client taints should be used sparingly, since they can increase the number of requests to Vault.


thyton (approved):
LGTM

benashz:
Closing in favor of #717

PR #686, opened 2024-04-11
Client factory: lock by client cache key

benashz:
Previously all calls to Get() were serialized using a single lock. This approach does not scale as the number of Vault client grows. With this change locking is done by ClientCacheKey, thereby reducing the overall contention for Vault clients.


thyton (approved):

benashz:
Closing in favor of #716 716

PR #702, opened 2024-04-23
add argo.Rollout support for RolloutRestartTarget

thyton:
add argo.Rollout support for RolloutRestartTarget

- add TestRolloutRestart unit test
- extend TestVaultDynamicSecret test cases to have []RolloutRestartTarget to test varying resources (deployment, argo.Rollout) 
- extend assertRolloutRestarts() to cover argo.Rollout


api/v1beta1/common.go
```diff
@@ -39,11 +39,12 @@ type Destination struct {
...
-// Supported resources: Deployment, DaemonSet, StatefulSet
+// Supported resources: Deployment, DaemonSet, StatefulSet, argo.Rollout
 type RolloutRestartTarget struct {
-	// +kubebuilder:validation:Enum={Deployment,DaemonSet,StatefulSet}
-	Kind string `json:"kind"`
-	Name string `json:"name"`
+	// +kubebuilder:validation:Enum={Deployment,DaemonSet,StatefulSet,argo.Rollout}
+	Kind       string `json:"kind"`
+	Name       string `json:"name"`
+	APIVersion string `json:"apiVersion,omitempty"`
```
thyton, line 47: I wonder how to add +kubebuilder:validation: for `APIVersion` based on `Kind` like {"", "argoproj.io/v1alpha1" } for argo.Rollout.

internal/helpers/rollout_restart.go
```diff
@@ -9,9 +9,11 @@ import (
 	"fmt"
 	"time"
 
+	argorolloutsv1alpha1 "github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
tvoran, line 16: These look like they're the same package? Probably should consolidate to `metav1`.

internal/helpers/rollout_restart.go
```diff
@@ -140,6 +151,10 @@ func patchForRolloutRestart(ctx context.Context, obj ctrlclient.Object, client c
 		}
 		t.Spec.Template.ObjectMeta.Annotations[AnnotationRestartedAt] = time.Now().Format(time.RFC3339)
 		return client.Patch(ctx, t, patch)
+	case *argorolloutsv1alpha1.Rollout:
+		patch := ctrlclient.MergeFrom(t.DeepCopy())
```
tvoran, line 155: I'm curious why this uses `MergeFrom()` instead of `StrategicMergeFrom()` like the other cases?

internal/helpers/rollout_restart_test.go
```diff
@@ -0,0 +1,176 @@
...
+			},
+			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
+				return assert.ErrorContains(t, err,
+					fmt.Sprintf("unsupported Kind %q", "invalid"), i...)
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rolloutRestartObj := createRolloutRestartObj(t, ctx, tt.args.namespace, tt.args.target, defaultClient)
```
tvoran, line 128: Should `t.Parallel()` be called inside the `t.Run()` block as well?

internal/helpers/rollout_restart_test.go
```diff
@@ -0,0 +1,176 @@
...
+			},
+			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
+				return assert.ErrorContains(t, err,
+					fmt.Sprintf("is paused, cannot restart it"), i...)
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.NoError(t, defaultClient.Create(ctx, tt.obj))
```
tvoran, line 171: Same question here about `t.Parallel()`.

api/v1beta1/common.go
tvoran, line 47: Could you elaborate on the need for the APIVersion field? It looks like there's only one version of argo's Rollout object right now.

For validation, would something like this work? Since it's not a required field:

```suggestion
        // +kubebuilder:validation:Enum={argoproj.io/v1alpha1}
	APIVersion string `json:"apiVersion,omitempty"`
```

internal/helpers/rollout_restart.go
thyton: Thank you! I've consolidated them.

internal/helpers/rollout_restart.go
thyton: `StrategicMergeFrom()` doesn't support CRDs. I've added a comment for the explaination https://github.com/hashicorp/vault-secrets-operator/blob/461cf779fab39327ab4a1460236b63a3b70bfff4/internal/helpers/rollout_restart.go#L163

api/v1beta1/common.go
thyton: > Could you elaborate on the need for the APIVersion field?

I've added some comments to elaborate 
https://github.com/hashicorp/vault-secrets-operator/blob/461cf779fab39327ab4a1460236b63a3b70bfff4/internal/helpers/rollout_restart.go#L109-L110
https://github.com/hashicorp/vault-secrets-operator/blob/461cf779fab39327ab4a1460236b63a3b70bfff4/api/v1beta1/common.go#L47-L51

thyton: > // +kubebuilder:validation:Enum={argoproj.io/v1alpha1}

I was thinking about it earlier. My concern is it doesn't validate `APIVersion` based on `Kind` (in this case argo.Rollout).

I wonder it's a good practice/benefit to add it as a first layer of validation. We have a `APIVersion` validation logic for each `Kind` in `RolloutRestartTarget()`.

test/integration/operator/terraform/argocd.tf
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+resource "helm_release" "argo_rollouts" {
+  namespace        = "argo"
+  create_namespace = true
+  name             = "argo-rollouts"
+  repository       = "https://argoproj.github.io/argo-helm"
+  chart            = "argo-rollouts"
+}
```
tvoran, line 10: We may want to specify a version for this chart so it's pinned for reproducibility?

test/integration/integration_test.go
```diff
@@ -663,80 +668,133 @@ func awaitRolloutRestarts(t *testing.T, ctx context.Context,
...
 	var errs error
 	var s *status
 
+	// TODO thy - remove before merge
+	//prettyprint := func(i interface{}, name string) {
+	//	fmt.Println(name)
+	//	b, err := json.MarshalIndent(i, "", "  ")
+	//	assert.Equal(t, nil, err)
+	//	fmt.Println(string(b))
+	//}
```
tvoran, line 702: Still needed?

api/v1beta1/common.go
tvoran: Maybe we don't need to add `APIVersion` to the CRD yet? I guess I'm having trouble understanding why someone would use it now, since there's only one argo.Rollout version served by the Argo project?

thyton: I'm seeing benefits in both removing APIVersion (1) and keeping it (2). I'm on the fence here. Wondering if you, @benashz, and @tomhjp have any thoughts.
1. Follows a practice of only writing code that is used and simplifies CRD.
2. Prepares the code framework for a future need of concurrent RolloutRestartTargets that have different APIVersions. 

Also for my learning, when should we start incorporating a future feature framework into the current development? (i.e. how far away in the future is a feature need to be considered for starting its framework incorporation?)

benashz: I am okay to punt on adding support for specifying the APIVersion for the target Kind. We can call YAGNI here. We should definitely leave behind some documentation that we will probably need to support APIVersion in the future.

When the APIVersion is bumped we will most definitely need it since, there can be Rollout instances of varying versions deployed to the cluster.

test/integration/integration_test.go
thyton: Sorry. I will remove these

api/v1beta1/common.go
thyton: I removed APIVersion and documented APIVersion consideration here
https://github.com/hashicorp/vault-secrets-operator/blob/1b79bbd54ef16af51327b7c5a8ff03f7fe78488a/api/v1beta1/common.go#L47-L50.

Let me know if I can improve it anyhow.

go.mod
```diff
@@ -7,6 +7,7 @@ toolchain go1.22.2
 require (
 	cloud.google.com/go/compute/metadata v0.3.0
 	github.com/Masterminds/sprig/v3 v3.2.3
+	github.com/argoproj/argo-rollouts v1.6.4
```
tvoran, line 10: BTW this is up to v1.6.6 now.

test/integration/integration_test.go
```diff
@@ -663,80 +668,120 @@ func awaitRolloutRestarts(t *testing.T, ctx context.Context,
...
-				annotations = o.Spec.Template.Annotations
-				tObj = &o
+			if minGeneration > 1 {
+				var o appsv1.Deployment
+				if assert.NoError(t, client.Get(ctx, tObjKey, &o)) {
+					annotations = o.Spec.Template.Annotations
+					tObj = &o
+				}
+				s = &status{o.Spec.Replicas, o.Status.ReadyReplicas}
 			}
```
tvoran, line 716: Am I reading it right that `assertRolloutRestarts() `now skips checking Deployments, StatefulSets, or ReplicaSets at all unless `minGeneration > 1`? From the [calls in the dynamic secrets tests](https://github.com/hashicorp/vault-secrets-operator/blob/d4a768e8abde9e0d687709c78ed5bce78c027230/test/integration/vaultdynamicsecret_integration_test.go#L319-L323), it looks like `awaitRolloutrestarts()` is expected to verify that the targets are deployed and ready if `minGeneration=1`, but I'm not seeing that here.

https://github.com/hashicorp/vault-secrets-operator/blob/d4a768e8abde9e0d687709c78ed5bce78c027230/test/integration/vaultdynamicsecret_integration_test.go#L319-L323

test/integration/integration_test.go
```diff
@@ -663,80 +668,120 @@ func awaitRolloutRestarts(t *testing.T, ctx context.Context,
...
+	ReadyReplicas int32
+}
+
+// assertRolloutRestarts asserts status and object state of each RolloutRestartTarget based on minGeneration.
+// If minGeneration==1, minimally assert each RolloutRestartTarget reached a healthy state after created
+// RolloutRestartTarget like argo.Rollout needs to reach a healthy state before we can effectively test
+// the rollout restart via a reconciler.Reconcile() - when argo.Rollout is making progress towards a healthy state
+// after created, any immediately triggered restart won't take effect
+// If minGeneration==2, assert each RolloutRestartTarget restarted, reached a healthy state, and has minGeneration
+// and appropriate restartAt.
```
tvoran, line 686: I'd suggest breaking up and formatting some of this text like below to make it easier to read. It also wouldn't hurt to wrap the lines at ~85 columns too (makes it easier to read in a diff).

```suggestion
// assertRolloutRestarts asserts status and object state of each RolloutRestartTarget based on minGeneration.
//
// If minGeneration==1, minimally assert each RolloutRestartTarget reached a healthy state after being created.
// RolloutRestartTarget's like argo.Rollout need to reach a healthy state before we can effectively test
// the rollout restart via a reconciler.Reconcile() - when argo.Rollout is making progress towards a healthy state
// after being created, any immediately triggered restart won't take effect.
//
// If minGeneration==2, assert each RolloutRestartTarget restarted, reached a healthy state, and has minGeneration
// and appropriate restartAt.
```

benashz (commented):
Looking good! I had a few initial comments/suggestions.

test/integration/integration_test.go
thyton: Thanks a lot for reformatting these!

test/integration/integration_test.go
thyton: After creation, the generation is 1 and there is no restartAt due to no secret syncs happening yet, so my thought was we could skip checking those. My idea mainly was to reuse `awaitRolloutrestarts()` and `assertRolloutRestarts()` to ensure the argo.Rollout resource reaches a healthy state. This led to using `minGeneration=1` to control `assertRolloutRestarts()`'s minimal checking (no generation and restartAt checking).

Looking back I think this pattern is odd. I've removed `if minGeneration > 1 {` condition. We will check the generation but not restartAt in the case `minGeneration=1`.

thyton: Resolving this conversation since I've reverted assertRolloutRestarts()'s behavior based on `minGeneration`.

tvoran (approved):

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -190,7 +190,17 @@ func TestVaultDynamicSecret(t *testing.T) {
 		existing           int
 	}{
 		{
-			name:     "existing-only",
+			name:           "one-existing-only",
```
benashz, line 193: Could we maybe set `withArgoRollout` on the `create-only` test below? Is there a problem running more than one test with argo rollouts?

thyton: I've removed `one-existing-only` test case and added `withArgoRollout: true` to both `existing-only` and `create-only` test cases. Tests passed locally.

thyton: Looks like CI workflow has problems with running more than one test with argo rollouts. Is it ok to have `withArgoRollout` in one test case?

benashz (approved):
LGTM - Great work!

Hige:
Argo doesn't have a resource like "argo.Rollout" in "Kind," there is simply "Rollout."
Example: https://github.com/argoproj/argocd-example-apps/blob/master/blue-green/templates/rollout.yaml#L2

Probably because of this, the pod restart does not work for me when secrets in the vault are changed, when I specify "argo.Rollout." Meanwhile, it works with "Deployment."

yaml
Copy code
rolloutRestartTargets:
  - kind: argo.Rollout
    name: {{ .Release.Name }}-app
Do you really need to use argo.Rollout instead of Rollout?

thyton:
> Argo doesn't have a resource like "argo.Rollout" in "Kind," there is simply "Rollout." Example: https://github.com/argoproj/argocd-example-apps/blob/master/blue-green/templates/rollout.yaml#L2
> 
> Probably because of this, the pod restart does not work for me when secrets in the vault are changed, when I specify "argo.Rollout." Meanwhile, it works with "Deployment."
> 
> ```
> 
> rolloutRestartTargets:
>   - kind: argo.Rollout
>     name: {{ .Release.Name }}-app
> ```
> 
> Do you really need to use argo.Rollout instead of Rollout?

@Hige We use `argo.Rollout` ([project/origin name].[kind]) to:
1. Separate it from the K8s builtin rollout kinds like Deployment.
2. Avoid future duplications if we support `Rollout` from a different project that is not `argo`. (foo.Rollout)

In the implementation of `argo.Rollout` support, we patch [the Rollout object's Spec.RestartAt](https://github.com/argoproj/argo-rollouts/blob/fdcb0d865732d1de0c3ac35bc8308bf3ea072201/pkg/apis/rollouts/v1alpha1/types.go#L80-L81) to indicate the restart request and cover that the object's generation reflecting the patch via [integration test](https://github.com/hashicorp/vault-secrets-operator/pull/702/files#diff-ca62391b26e30e01d1f8cee0eac7829ba8e9b5e22d80eab0e83881086fb4177bR710-R735).

Note: the Argo Rollouts controller (not VSO) reconciles argo.Rollout objects and is responsible for restarting the pods based on [the Rollout object's Spec.RestartAt](https://github.com/argoproj/argo-rollouts/blob/fdcb0d865732d1de0c3ac35bc8308bf3ea072201/pkg/apis/rollouts/v1alpha1/types.go#L80-L81)

Hige:
Thank you for the detailed response.
Why might a pod restart not work?

I am applying the following manifest:

```
---
kind: VaultStaticSecret
metadata:
  name: {{ .Release.Name }}-env
spec:
  vaultAuthRef: {{ .Release.Name }}-app
  type: kv-v2
  mount: secret
  path: {{ .Release.Name }}/config
  refreshAfter: 10s
  destination:
    name: vault-env
    create: true
    overwrite: true
  rolloutRestartTargets:
    - kind: argo.Rollout
      name: {{ .Release.Name }}-app

```

I change the env in the vault. I check that they are applied in the secret (this indeed works). I check if the pod restarted, but the pod did not restart and the "RestartAt" annotation was not added.

However, if I use the "Deployment" resource, everything works as expected.

Could you advise what can be done? What to check. There was no information about restarts in the "vault-secrets-operator" logs. Could the problem be on your side?

thyton:
@Hige Sorry for the experience you're having. 
> What to check. 

Based on [this code block](https://github.com/hashicorp/vault-secrets-operator/blob/5797542c8e5a476f0985525fbc2368abbde6f3e8/internal/helpers/rollout_restart.go#L60-L75), you could check the following:
1. K8s events via [kubectl events](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_events/)
I wonder if you see any K8s events indicating the reason as "RolloutRestartFailed" or "RolloutRestartTriggered".
```
kubectl events -A --watch -o json | jq 'select(.reason | startswith("RolloutRestart")) | .'
```
2. VSO logs
Note: Debug mode is required to see "Rollout restart succeeded"  
```
if errs != nil {
      logger.Error(errs, "Rollout restart failed", "targets", targets)
} else {
      logger.V(consts.LogLevelDebug).Info("Rollout restart succeeded", "total", len(targets))
}
```

I see that you use Helm to deploy VSO, so we could rule out potential issues with the manager RBAC to update argo.Rollout resources for now.

If you don't see that both 1 and 2's outputs indicate the rollout restart triggered, could you possibly redact and share? Otherwise, the Argo Rollouts controller's log is probably the next place to check why the pod did not restart.

> I check if the pod restarted, but the pod did not restart and the "RestartAt" annotation was not added.

We don't use the annotation update approach, which is used for Deployment, to trigger the restart for argo.Rollout. We patch [the Rollout object's Spec.RestartAt](https://github.com/argoproj/argo-rollouts/blob/fdcb0d865732d1de0c3ac35bc8308bf3ea072201/pkg/apis/rollouts/v1alpha1/types.go#L80-L81). Here is the [implementation](https://github.com/hashicorp/vault-secrets-operator/pull/702/files#diff-4c9a8f1818b817ef952f7baf30924710fff0c6e162eeda0f1013ba57d9ff81c9R170-R172) for more details.

PR #704, opened 2024-04-24
Replace the SyncController with source.Channel

benashz:
Update the VaultDynamicSecretReconciler to rely on the controller-runtime's source.Channel that is watched as a raw source. This ensures that there is only one reconciler work queue required to handle external reconciliation requests.


controllers/vaultdynamicsecret_controller.go
```diff
@@ -710,28 +684,23 @@ func (r *VaultDynamicSecretReconciler) vaultClientCallback(ctx context.Context,
...
-				Delay:        delay,
-				RequeueOnErr: true,
-				Request: reconcile.Request{
-					NamespacedName: types.NamespacedName{
-						Namespace: o.GetNamespace(),
-						Name:      o.GetName(),
-					},
-				},
+			evt := event.GenericEvent{
+				Object: o.DeepCopy(),
```
tvoran, line 691: IIRC events really only need the namespace and name of the object, so we might be able to avoid the deepcopy something like this:

```suggestion
				Object: &secretsv1beta1.VaultDynamicSecret{
					ObjectMeta: metav1.ObjectMeta{
						Namespace: o.Namespace,
						Name:      o.Name,
					},
				},
```

tvoran (approved):

controllers/vaultdynamicsecret_controller.go
benashz: Fixed in 3ee8cfac7d397fe70f2f1d263ce5b8e62317c33e

PR #716, opened 2024-04-29
CachingClientFactory: lock by client cache key

benashz:
Previously all calls to Get() were serialized using a single lock. This approach does not scale as the number of Vault client grows. With this change locking is done by ClientCacheKey, thereby reducing the overall contention for Vault clients.

### DIFF
--- a/internal/vault/client_factory.go
+++ b/internal/vault/client_factory.go
@@ -217,7 +217,7 @@ func (m *cachingClientFactory) onClientEvict(ctx context.Context, client ctrlcli
 		}
 	}
 
-	m.removeLock(cacheKey)
+	m.removeClientLock(cacheKey)
 }
 
 // Restore will attempt to restore a Client from storage. If storage is not enabled then no restoration will take place.
@@ -281,7 +281,7 @@ func (m *cachingClientFactory) isDisabled() bool {
 	return m.shutDown
 }
 
-func (m *cachingClientFactory) clientKeyLock(cacheKey ClientCacheKey) (*sync.RWMutex, bool) {
+func (m *cachingClientFactory) clientLock(cacheKey ClientCacheKey) (*sync.RWMutex, bool) {
 	m.clientLocksLock.Lock()
 	defer m.clientLocksLock.Unlock()
 	lock, ok := m.clientLocks[cacheKey]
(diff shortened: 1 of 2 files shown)


thyton (approved):
LGTM

internal/vault/client_factory.go
```diff
@@ -97,8 +97,11 @@ type cachingClientFactory struct {
 	clientCallbacks        []ClientCallbackHandler
 	callbackHandlerCh      chan Client
 	mu                     sync.RWMutex
+	muLocks                sync.RWMutex
+	encClientLock          sync.RWMutex
 	onceDoWatcher          sync.Once
 	callbackHandlerCancel  context.CancelFunc
+	cacheKeyLocks          map[ClientCacheKey]*sync.RWMutex
```
tvoran, line 104: Is this essentially a lock per client? Maybe naming this something like `clientLocks` would be more clear?

internal/vault/client_factory.go
```diff
@@ -270,6 +275,23 @@ func (m *cachingClientFactory) isDisabled() bool {
...
+	defer m.muLocks.Unlock()
+	lock, ok := m.cacheKeyLocks[cacheKey]
+	if !ok {
+		lock = &sync.RWMutex{}
+		m.cacheKeyLocks[cacheKey] = lock
+	}
+	return lock, ok
+}
+
+func (m *cachingClientFactory) removeLock(cacheKey ClientCacheKey) {
```
tvoran, line 289: This looks pretty specific to the cacheKey/client lock, so maybe something more along the lines of `removeCacheKeyLock()` or `removeClientLock`?

internal/vault/client_factory.go
```diff
@@ -278,8 +300,6 @@ func (m *cachingClientFactory) isDisabled() bool {
 //
 // Supported types for obj are: VaultDynamicSecret, VaultStaticSecret. VaultPKISecret
 func (m *cachingClientFactory) Get(ctx context.Context, client ctrlclient.Client, obj ctrlclient.Object) (Client, error) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
```
tvoran, line 282: Just from looking what else uses `mu.Lock()`, any concern with effects between Get() and prune() with respect to cacheKey locking? (i.e. should prune() be using cachKey locks?)

tvoran (approved):

internal/vault/client_factory.go
benashz: Good catch. I've added a read lock in d7d1b288c2c22332543c51175171f1f4ab9f4115

internal/vault/client_factory.go
benashz: Updated in d7d1b288c2c22332543c51175171f1f4ab9f4115

internal/vault/client_factory.go
benashz: Updated in b73710dd31d0aff71e7a57c9ffe4d9d4c321772e

PR #718, opened 2024-04-30
VDS: Selectively log calls to SyncRegistry.Delete()

benashz:


controllers/registry_test.go
```diff
@@ -475,7 +489,7 @@ func TestSyncRegistry(t *testing.T) {
 					case 0:
 						r.Add(ttt.objKey)
 					case 1:
-						r.Delete(ttt.objKey)
+						assert.Equal(t, r.Delete(ttt.objKey), ttt.wantHas)
```
tvoran, line 492: I think technically the expected value is supposed to be first. Doesn't change the behavior, but might make the failure message a little more clear.

```suggestion
						assert.Equal(t, ttt.wantHas, r.Delete(ttt.objKey))
```

tvoran (approved):

PR #722, opened 2024-05-02
Client: skip LifetimeWatcher validation for non-renewable auth tokens

benashz:
Previously calling Validate() on a defaultClient that was authenticated using a non-renewable Vault token, would result in an error when the Client was being fetched from the CachingClientFactory. The defaultClient does not start a LifetimeWatcher in this case, so any related validation checks are invalid. This PR corrects that.


thyton (approved):
LGTM

PR #731, opened 2024-05-08
Core: set the zap time encoder to include nanos

benashz:
Updates the zap logger to include timestamps with nano second resolution.

E.g:
```
2024-05-08T14:02:26.489440746Z  INFO    Stopping and waiting for caches
2024-05-08T14:02:26.489861624Z  INFO    Stopping and waiting for webhooks
2024-05-08T14:02:26.489883565Z  INFO    Stopping and waiting for HTTP servers
```


main.go
```diff
@@ -118,6 +119,7 @@ func main() {
 			"Valid values are: %v", []string{"exclude-raw"}))
 	opts := zap.Options{
 		Development: true,
+		TimeEncoder: zapcore.RFC3339NanoTimeEncoder,
```
tvoran, line 122: Curious why this needs to be nanosecond precision? I see that Vault itself seems to go down to millisecond for its logs (e.g. `2024-05-09T16:28:09.271Z`).

Would it make sense for this to be a development or debug mode option?

tvoran (approved):

main.go
benashz: > Curious why this needs to be nanosecond precision? I see that Vault itself seems to go down to millisecond for its logs (e.g. `2024-05-09T16:28:09.271Z`).

I think this follows the timestamp format of the K8s control plane using https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/time/format.go;l=112.

e.g:
```
2024-05-08T15:00:34.614182314Z stderr F 2024-05-08T15:00:34Z    INFO    Starting EventSource    {"controller": "vaultstaticsecret", "controllerGroup": "secrets.hashicorp.com", "controllerKind": "VaultStaticSecret", "source": "kind source: unknown type"}
```

Could go to millis. Not sure.

> 
> Would it make sense for this to be a development or debug mode option?

Perhaps. I always prefer to higher resolution by default.

thyton (approved):

benashz:
Closing as fixed in #778

PR #732, opened 2024-05-08
Use exponential backoffs on secret source errors.

benashz:
Previously, the back off duration was based on a fixed duration + some jitter. This PR introduces exponential back offs for all secret syncing controllers. The back off will be calculated and honored whenever an error is encountered while fetching from a secret source e.g: Vault, HCPVS. The back off configuration is controlled via some new command line arguments:

```
  -back-off-initial-interval duration
        Initial interval between retries on secret source errors. All errors are tried using an exponential backoff strategy. Also set from environment variable VSO_BACK_OFF_INITIAL_INTERVAL. (default 5s)
  -back-off-max-interval duration
        Maximum interval between retries on secret source errors. All errors are tried using an exponential backoff strategy. Also set from environment variable VSO_BACK_OFF_MAX_INTERVAL. (default 1m0s)
  -back-off-multiplier float
        Sets the multiplier for increasing the interval between retries on secret source errors. All errors are tried using an exponential backoff strategy. Also set from environment variable VSO_BACK_OFF_MULTIPLIER. (default 1.5)
  -back-off-randomization-factor float
        Sets the randomization factor to add jitter to the interval between retries on secret source errors. All errors are tried using an exponential backoff strategy. Also set from environment variable VSO_BACK_OFF_RANDOMIZATION_FACTOR. (default 0.5)
```

Or through the Helm chart values:
```yaml
    # Backoff settings for the controller manager. These settings control the backoff behavior
    # when the controller encounters an error while fetching secrets from the SecretSource.
    backOffOnSecretSourceError:
      # Initial interval between retries.
      # @type: duration
      initialInterval: "5s"
      # Maximum interval between retries.
      # @type: duration
      maxInterval: "60s"
      # Randomization factor to add jitter to the interval between retries.
      # @type: float
      randomizationFactor: 0.5
      # Sets the multiplier for increasing the interval between retries.
      # @type: float
      multiplier: 1.5
```

### DIFF
--- a/controllers/registry_test.go
+++ b/controllers/registry_test.go
@@ -6,7 +6,9 @@ package controllers
 import (
 	"sync"
 	"testing"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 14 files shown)


controllers/vaultdynamicsecret_controller.go
```diff
@@ -158,7 +159,10 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 	// indicates that the resource has been added to the SyncRegistry
 	// and must be synced.
 	case r.SyncRegistry.Has(req.NamespacedName):
+		// indicates that the resource has been added to the SyncRegistry
+		// and must be synced.
 		syncReason = "force sync"
+		syncReason = "last sync error"
```
thyton, line 165: Should we keep only line 165?

thyton (commented):
LGTM from my initial pass

controllers/vaultdynamicsecret_controller.go
thyton, line 165:
```suggestion
```
we probably won't need this line.

controllers/registry_test.go
```diff
@@ -503,3 +505,116 @@ func TestSyncRegistry(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			r := &BackOffRegistry{
+				m:    tt.m,
+				opts: tt.opts,
+			}
+			got, got1 := r.Get(tt.objKey)
+			assert.NotNilf(t, got, "Get(%v)", tt.objKey)
+			assert.Equalf(t, tt.want1, got1, "Get(%v)", tt.objKey)
+			assert.Greaterf(t, got.bo.NextBackOff(), time.Duration(0), "Get(%v)", tt.objKey)
+		})
```
tvoran, line 568: Might be good to test that the backoff in the registry is being incremented by `NextBackOff()`, i.e. do another `r.Get()` and check that `NextBackoff()` is greater than the first one.

controllers/vaultpkisecret_controller.go
```diff
@@ -183,9 +184,14 @@ func (r *VaultPKISecretReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		if err := r.updateStatus(ctx, o); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		r.SyncRegistry.Add(req.NamespacedName)
```
tvoran, line 188: It looks like the sync reason may need to be updated on line 123? I see the VDS controller uses "force sync" for this case.
https://github.com/hashicorp/vault-secrets-operator/blob/697604208121ce6e38fbad214538700ddec1b855/controllers/vaultpkisecret_controller.go#L122-L123

tvoran (approved):
👍 

I wonder if it would make sense to log the backoff settings on startup? Just to make it easier for users to see what's set.

controllers/registry_test.go
benashz: Added in 15e14c91bacf2ebdcfc5ad3888975587fe0287c3

controllers/vaultpkisecret_controller.go
benashz: Synced up in d10ffde439745a2ad2b55d1170e7cce99c10440a

benashz:
> 👍
> 
> I wonder if it would make sense to log the backoff settings on startup? Just to make it easier for users to see what's set.

Thanks! I made that change in 15e14c91bacf2ebdcfc5ad3888975587fe0287c3. I also added a new Prometheus metric that includes the same info:
```
vso_runtime_config{backOffInitialInterval="5s",backOffMaxInterval="1m0s",backOffMultiplier="1.50",backOffRandomizationFactor="0.50",clientCachePersistenceModel="direct-encrypted",clientCacheSize="10000",globalTransformationOptions="",maxConcurrentReconciles="100"} 1
```

PR #739, opened 2024-05-10
Call VDS Callbacks on VaultAuth and VaultConnection changes

benashz:
During some internal work we discovered that when updating a `VaultAuth` instance, any associated Vault Clients are pruned from the cache without notifying the `VaultDynamicSecrets` controller. This could potentially lead to premature revocation of secret leases, causing service disruptions for applications that rely on valid secret data.

Upon update to a `VaultAuth` or `VaultClient` instance, the `CachingClientFactory` will send any pruned Clients to its `callbackHandler` channel. Upon deletion of either resource, the Client will be removed from the cache and no callbacks will be called.


thyton:
LGTM. Let me know if/when the update for integration tests will be in. I can continue the review then.

benashz:
> LGTM. Let me know when the update for integration tests is in. I can continue the review then.

Updated with bb331e3023f62abe04b33ef43c99277e83814311

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -662,13 +663,69 @@ func TestVaultDynamicSecret_vaultClientCallback(t *testing.T) {
 	}
 
 	tests := []struct {
-		name     string
-		create   int
-		expected map[string]int
+		name        string
+		create      int
+		triggerFunc func(t *testing.T, reconciledObjs []*secretsv1beta1.VaultDynamicSecret)
+		expected    map[string]int
```
thyton, line 669: Looks like `expected` isn't used. Or do we have plan for using it in later work?

thyton (approved):
LGTM. Nice tests!

PR #749, opened 2024-05-14
update tlsConfig & namespaceSelector for metrics serviceMonitor

iamps5:
Update the comment to fix the [default selector label](https://github.com/hashicorp/vault-secrets-operator/blob/main/chart/templates/prometheus-servicemonitor.yaml#L25) in serviceMonitor
Added namespaceSelector field - this is required incase ServiceMonitor is deployed in any other namespace instead of VSO namespace
Added tlsConfig[] to add CA certificates, client certificates and client keys for SSL/TLS certificates validation on metrics' serviceMonitor. Without these we don't even have option to add these configuration in serviceMonitor

### DIFF
--- a/chart/templates/prometheus-servicemonitor.yaml
+++ b/chart/templates/prometheus-servicemonitor.yaml
@@ -23,6 +23,11 @@ spec:
   selector:
     matchLabels:
       control-plane: controller-manager
+  {{- if .Values.telemetry.serviceMonitor.namespaceSelector }}
+  namespaceSelector:
+    matchNames:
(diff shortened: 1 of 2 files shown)


iamps5:
@benashz can you please let me know if you are going to accept these changes? If yes, when? thanks

iamps5:
[Latest commit](https://github.com/hashicorp/vault-secrets-operator/pull/749/commits/0ef5dc560208c1f63090e15d430b21a1b949bfe1) -  has a quick fix for updating the image tag for `test-runner.yaml` - Since its doesn't contain any tag, it is pulling latest image which is mismatch in case someone is using old version of VSO

PR #753, opened 2024-05-16
Ensure that spec.hmacSecretData's value is honoured

benashz:
Previously, upon the first application of a VaultStaticSecret instance that had spec.hmacSecretData explicitly set to false, the K8s API would replace that value with default as is defined in the CRD's schema. This fix makes HMACSecretData a pointer receiver, which make K8s do the right thing.

- [x] add regression tests


fairclothjm (approved):
LGTM

benashz:
Thanks!

PR #755, opened 2024-05-16
Core: add support for setting maxElapsedTime backoff

benashz:
In #732, we add support for exp. backoffs on secret source errors. This PR adds the ability to configure the max elapsed time before giving up on syncing a secret resource.

```
  -back-off-max-elapsed-time duration
        Maximum elapsed time before giving up on secret source errors. All errors are tried using an exponential backoff strategy. Also set from environment variable VSO_BACK_OFF_MAX_ELAPSED_TIME.
```

Disabled by default.


fairclothjm (approved):
LGTM

benashz:
Thanks!

PR #757, opened 2024-05-17
Extend VaultAuthGlobalRef to support merge strategy configuration

benashz:
Updates #735 API changes. Previously the `vaultAuthGlobalRef` was a string, now it is a map that extends the configuration options that can be applied during the VaultAuth+VaultAuthGlobal merger.

Examples:
```yaml
---
kind: VaultAuth
metadata:
  name: default
spec:
  vaultAuthGlobalRef:
    mergeStrategy:
      headers: union
      params: union
    name: default
---
kind: VaultAuthGlobal
metadata:
  name: default
spec:
  defaultAuthMethod: kubernetes
  kubernetes:
    audiences:
    - vault
    mount: demo-auth-mount
    namespace: demo-ns
    role: auth-role
    serviceAccount: default
    tokenExpirationSeconds: 600
```


benashz:
Moved to #800

PR #762, opened 2024-05-21
Fix backoff registry tests

benashz:
Reduce the randomization factor to ensure that each backoff duration is greater than the last. With the underlying backoff library it is possible to have new backoff durations that are less than a previous one.

### DIFF
--- a/controllers/registry_test.go
+++ b/controllers/registry_test.go
@@ -514,8 +514,7 @@ func TestBackOffRegistry_Get(t *testing.T) {
 		m      map[client.ObjectKey]*BackOff
 		opts   []backoff.ExponentialBackOffOpts
 		objKey client.ObjectKey
-		want   *BackOff
-		want1  bool
+		want   bool
 	}{
 		{
 			name: "new",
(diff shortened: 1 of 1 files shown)


controllers/registry_test.go
```diff
@@ -539,34 +535,30 @@ func TestBackOffRegistry_Get(t *testing.T) {
...
-			assert.NotNilf(t, got, "Get(%v)", tt.objKey)
-			assert.Equalf(t, tt.want1, got1, "Get(%v)", tt.objKey)
-			last := got.bo.NextBackOff()
-			assert.Greaterf(t, last, time.Duration(0), "Get(%v)", tt.objKey)
-			assert.Greaterf(t, got.bo.NextBackOff(), last, "Get(%v)", tt.objKey)
+			bo, created := r.Get(tt.objKey)
+			assert.NotNilf(t, bo, "Get(%v)", tt.objKey)
+			assert.Equalf(t, tt.want, created, "Get(%v)", tt.objKey)
+			assert.Greaterf(t, bo.NextBackOff(), time.Duration(0), "Get(%v)", tt.objKey)
+			assert.Lessf(t, bo.NextBackOff(), bo.NextBackOff(), "Get(%v)", tt.objKey)
```
tvoran, line 561: Does this match how the backoff registry is being used? I thought there would be an `r.Get()` in-between calls to `bo.NextBackoff()`.

tvoran:
> With the underlying backoff library it is possible to have new backoff durations that are less than a previous one.

Is that just due to the randomization being used with the default options?

tvoran (approved):

controllers/registry_test.go
benashz: Great call out. Updated in cc31c2dc645af549833c5d5314fc2f8eed63a144

PR #763, opened 2024-05-21
Rename back-off to backoff

benashz:
Backoff is one word. Update all command line args to reflect that.


main.go
```diff
@@ -361,7 +361,7 @@ func main() {
 		SecretDataBuilder:          secretDataBuilder,
 		HMACValidator:              hmacValidator,
 		ClientFactory:              clientFactory,
-		BackOffRegistry:            controllers.NewBackOffRegistry(backOffOpts...),
+		BackOffRegistry:            controllers.NewBackOffRegistry(backoffOpts...),
```
thyton, line 364:
```suggestion
		BackoffRegistry:            controllers.NewBackOffRegistry(backoffOpts...),
```

thyton (approved):
LGTM! Thank you.
I know this PR is only scoped to command line arg changes. Should we also update other places (e.g. `BackOffRegistry` from the suggestion) to reflect backoff as one word in the future?

main.go
benashz: Good question. I decided to leave it as-is since the backoff library we are using spells  it 'BackOff'

PR #769, opened 2024-05-24
Core: call callbacks on invalidated Client

benashz:
Previously, invalidated Clients were removed from the Client cache without calling any of the registered ClientCallbacks. Now any callbacks registered with ClientCallbackOnCacheRemoval be invoked when a client is removed from cache.

This PR extends #717

### DIFF
--- a/internal/vault/client.go
+++ b/internal/vault/client.go
@@ -28,7 +28,7 @@ import (
 
 type ClientOptions struct {
 	SkipRenewal   bool
-	WatcherDoneCh chan<- Client
+	WatcherDoneCh chan<- *ClientCallbackHandlerRequest
 }
 
 func defaultClientOptions() *ClientOptions {
@@ -129,11 +129,10 @@ func NewClientFromStorageEntry(ctx context.Context, client ctrlclient.Client, en
 		return nil, fmt.Errorf("restored client's cacheKey %s does not match expected %s", cacheKey, entry.CacheKey)
 	}
 
-	if err := c.Validate(); err != nil {
-		return nil, err
-	}
+	c.Taint()
+	defer c.Untaint()
 
-	if _, err := c.Read(ctx, NewReadRequest("auth/token/lookup-self", nil)); err != nil {
+	if err := c.Validate(ctx); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


internal/vault/client_test.go
```diff
@@ -467,21 +474,9 @@ func Test_defaultClient_Validate(t *testing.T) {
 			lastWatcherErr: nil,
 			wantErr:        assert.NoError,
 		},
-		{
```
benashz, line 470: Removes the discovered duplicated test.

internal/vault/client_factory_test.go
```diff
@@ -174,3 +174,143 @@ func Test_cachingClientFactory_clientLocks(t *testing.T) {
...
+			cbOn:   ClientCallbackOnLifetimeWatcherDone,
+			cbFn: func(t *testing.T) (ClientCallback, *callbackResult) {
+				result := &callbackResult{}
+				return func(ctx context.Context, c Client) {
+					result.called = true
+					result.done = true
+				}, result
+			},
+			wantCalled: true,
+		},
```
thyton, line 220:
```suggestion
```
Looks like a duplicate of L193-206

internal/vault/client.go
```diff
@@ -129,11 +129,10 @@ func NewClientFromStorageEntry(ctx context.Context, client ctrlclient.Client, en
 		return nil, fmt.Errorf("restored client's cacheKey %s does not match expected %s", cacheKey, entry.CacheKey)
 	}
 
-	if err := c.Validate(); err != nil {
-		return nil, err
-	}
+	c.Taint()
+	defer c.Untaint()
```
thyton, line 133: I think with `defer c.Untaint()`, we always `Untaint()` either the client is valid or not. 
Should we `c.Untaint()` only when `c.Validate(ctx)` returns nil or before L139?

internal/vault/client_factory_test.go
thyton, line 220:
```suggestion
```
Looks like a duplicate of L193-206

thyton (commented):
Looking good to me. I have some comments.

internal/vault/client.go
benashz: Yeah, that was the prior approach. With the this change there should be no need to always untaint the client, since it was removed from the cache here: https://github.com/hashicorp/vault-secrets-operator/blob/a4d4f9f8f4397e9975eb5d340353e2ea404bccdf/internal/vault/client_factory.go#L443-L444

internal/vault/client_factory_test.go
benashz: Good catch! Fixed in 21436f9fdbe383db166502b785efc76c8700d3ee

Issue #494, opened 2023-12-05
Rotate Credentials button in Static Role in VDS doesn't work for pods No rolling updates.


toasahi:
![Group 1](https://github.com/hashicorp/vault-secrets-operator/assets/82006976/289f9653-de01-48d9-b11d-c7d3e20be53f)

**バグについて説明する**
The "Rotate credentials" button in the image changes the DataBase password, but not the pod environment variable (Secret). You will not be able to connect.
The reason is obvious: the values of "rotationPeriod" and "rotationSchedule" in the VaultStaticCredsMetaData are not changed from the values when VaultDynamicSecrets is applied.
Therefore, the pod does not perform a rolling update when the "Rotate credentials" button is pressed.
This will cause a big problem in the future.

**To Reproduce**
1. Enable database engine for postgres
2. Create static role
3. Apply the CRD of VSO.
4. Specify Deployment to be rotated for VDS
5. Press Rotate credential
6. Pods don't do rolling updates.

Application deployment:

```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultDynamicSecret
metadata:
  name: vso-db-demo
  namespace: default
spec:
  allowStaticCreds: true
  # Mount path of the secrets backend
  mount: db/postgres

  # Path to the secret
  path: static-creds/postgres-role

  # Where to store the secrets, end user will create the secret
  destination:
    create: true
    name: db-secret

  # Restart these pods when secrets rotated
  rolloutRestartTargets:
  - kind: Deployment
    name: postgres

  # Name of the CRD to authenticate to Vault
  vaultAuthRef: vault-auth
```

**期待される行動**
VDS(StaticRole) uses a single User, so when the RotateCredentials button is pressed, the Pod should do a rolling update and the Secret should be rewritten.

**Environment**
Kubernetes version:
EKS
vault: 1.15.1
vault-secrets-operator version: 0.4.0
postgres: 16.1.0

benashz:
Hi @toasahi, currently, VSO does not support the use-case described in this issue, rather it relies on the interval set from the last sync. We plan to extend VSO to support Vault's notification system to reconcile the sort of of event that you describe here.l Stay tuned.

Thanks,

Ben

toasahi:
Hi @benashz ,I am very excited about that feature.I look forward to future upgrades.

Thanks,

Asahi

Floppy012:
[...]

PR #771, opened 2024-05-25
VaultStaticSecret: support instant event-driven updates

tvoran:
Support for triggering a VaultStaticSecret (VSS) sync from Vault events instead of just on a refresh interval. Adds a new parameter in a new config block (`SyncConfig.InstantUpdates`) that enables VSS Reconcile() to spawns a goroutine with a websocket client that receives event notifications from Vault, and triggers Reconcile() for the VSS using the controller's source channel. Uses a registry based on go-cache to store which VSS' have event watchers running along with the event watcher's context cancel function and other control metadata.

Requirements:
* Vault Enterprise 1.16+
* Event [policy permissions](https://developer.hashicorp.com/vault/docs/concepts/events#policies) (see example below)

Caveats:
* There may be issues with the eventing client going stale with non-periodic Vault tokens (to be addressed in a future update)

Documentation update: https://github.com/hashicorp/vault/pull/27668

The websocket event client is built with the VaultStaticSecret's auth, so [extra permissions](https://developer.hashicorp.com/vault/docs/concepts/events#policies) will be needed on the policy for the associated VaultAuth role. 

Example for a kv-v2 secret "campaign" with the engine mounted under "kv-marketing" in the "us-west-org" namespace:

```shell
$ VAULT_NAMESPACE=us-west-org vault policy write marketing-read-only -<<EOF
path "kv-marketing/data/campaign" {
   capabilities = ["read", "list", "subscribe"]
   subscribe_event_types = ["kv*"]
}

path "sys/events/subscribe/kv*" {
   capabilities = ["read"]
}

EOF
```

And the corresponding VaultStaticSecret:

```yaml
kind: VaultStaticSecret
metadata:
  name: static-demo
  namespace: nicecorp
spec:
  namespace: us-west-org
  vaultAuthRef: event-demo-auth
  mount: kv-marketing
  type: kv-v2
  path: campaign
  refreshAfter: 1h   # <-- long refresh interval as a fallback
  destination:
    name: campaign-secret
    create: true
  syncConfig:
    instantUpdates: true  # <-- new option
```

Vault event subscription status and errors are emitted as k8s events on the VaultStaticSecret object. For example:

```console
$ kubectl describe vaultstaticsecret static-demo -n nicecorp
...
Events:
  Type     Reason               Age                   From               Message
  ----     ------               ----                  ----               -------
  Normal   SecretSynced         2m37s                 VaultStaticSecret  Secret synced
  Normal   SecretRotated        2m37s                 VaultStaticSecret  Secret synced
  Warning  EventWatcherError    16s (x18 over 2m37s)  VaultStaticSecret  Error while watching events: 
failed to connect to vault websocket: error returned when opening event stream web socket to 
wss://vault.vault.svc.cluster.local:8200/v1/sys/events/subscribe/kv%2A?json=true, ensure 
VaultAuth role has correct permissions and Vault is Enterprise version 1.16 or above: {"errors":["permission denied"]}
  Normal   EventWatcherStarted  6s                    VaultStaticSecret  Started watching events

```


Floppy012:
Is there any chance to implement this for Static Credentials of DB Secret Engines too? Or is this too far out of scope for this PR? (https://github.com/hashicorp/vault-secrets-operator/issues/494#issuecomment-1840689503)

tvoran:
Hi @Floppy012, DB Engine events are out of scope for this PR, but they're on our list for the future.

controllers/event_watcher_registry.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package controllers
+
+import (
+	"context"
+
+	gocache "github.com/patrickmn/go-cache"
```
benashz, line 9: I'm not familiar with this module. Do you think we should adopt it to replace some of the other internal caching implementations?

controllers/event_watcher_registry.go
```diff
@@ -0,0 +1,53 @@
...
+	StoppedCh chan struct{}
+	// LastGeneration is the generation of the VaultStaticSecret resource, used
+	// to detect if the event watcher needs to be recreated
+	LastGeneration int64
+	// LastClientID - vault client ID for the last successful connection, used
+	// to detect if the Vault client has changed since the event watcher started
+	LastClientID string
+}
+
+type EventWatcherRegistry struct {
```
benashz, line 28: nit: would be good to add some docs for this exported struct.

controllers/event_watcher_registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	_, isOpen := <-gotAgain.StoppedCh
+	assert.False(t, isOpen, "expected stoppedCh from registry item to be closed")
+
+	// Delete the event watcher
+	registry.Delete(itemName)
+
+	// Get the event watcher
+	gotFinally, ok := registry.Get(itemName)
+	assert.False(t, ok, "expected to not get event watcher, got one")
```
benashz, line 70: Could we also assert that the length of the registry matches what we expect at this point?

controllers/vaultstaticsecret_controller.go
```diff
@@ -6,27 +6,37 @@ package controllers
...
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"nhooyr.io/websocket"
```
benashz, line 20: Out of curiosity, is this the same module that we are using in agent?

controllers/vaultstaticsecret_controller.go
```diff
@@ -6,27 +6,37 @@ package controllers
...
 	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
 	"github.com/hashicorp/vault-secrets-operator/internal/consts"
 	"github.com/hashicorp/vault-secrets-operator/internal/helpers"
 	"github.com/hashicorp/vault-secrets-operator/internal/vault"
 )
 
-const vaultStaticSecretFinalizer = "vaultstaticsecret.secrets.hashicorp.com/finalizer"
+const (
+	vaultStaticSecretFinalizer = "vaultstaticsecret.secrets.hashicorp.com/finalizer"
+	eventPath                  = "/v1/sys/events/subscribe/kv*"
```
benashz, line 38: nit: may want to add `KV` somewhere in the name to differentiate it from other/future paths.

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
 	return nil
 }
 
+func (r *VaultStaticSecretReconciler) ensureEventWatcher(ctx context.Context, o *secretsv1beta1.VaultStaticSecret, c vault.Client) error {
+	logger := log.FromContext(ctx)
```
benashz, line 249: Could we also set the loggers name e.g: ` logger := log.FromContext(ctx).WithName("ensureEventWatcher")`

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+				Path     string `json:"path"`
+				Modified string `json:"modified"`
+			} `json:"metadata"`
+		} `json:"event"`
+		Namespace string `json:"namespace"`
+	} `json:"data"`
+}
+
+func (r *VaultStaticSecretReconciler) streamStaticSecretEvents(ctx context.Context, o *secretsv1beta1.VaultStaticSecret, wsClient *vault.WebsocketClient) error {
+	logger := log.FromContext(ctx)
```
benashz, line 415: nit: maybe name the logger based on this method's name?

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+			return nil
+		}
+	}
+	if meta != nil {
+		// The watcher is running, but the metadata or vault client has changed,
+		// so kill it
+		if meta.Cancel != nil {
+			meta.Cancel()
+			// Wait for the goroutine to stop and remove itself from the event registry
+			<-meta.StoppedCh
```
benashz, line 268: Do we need to worry about this blocking forever?

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+	updatedMeta := &eventWatcherMeta{
+		Cancel:         cancel,
+		LastClientID:   c.ID(),
+		LastGeneration: o.GetGeneration(),
+		StoppedCh:      stoppedCh,
+	}
+	// launch the goroutine to watch events
+	logger.V(consts.LogLevelDebug).Info("Starting event watcher", "meta", updatedMeta)
+	r.eventWatcherRegistry.Register(name, updatedMeta)
+	go r.getEvents(watchCtx, o, wsClient, stoppedCh)
```
benashz, line 289: Are okay to skip the error evaluation here?

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+
+	return nil
+}
+
+// unWatchEvents - If the VSS is in the registry, cancel its event watcher
+// context to close the goroutine, and remove the VSS from the registry
+func (r *VaultStaticSecretReconciler) unWatchEvents(o *secretsv1beta1.VaultStaticSecret) {
+	name := types.NamespacedName{Namespace: o.Namespace, Name: o.Name}
```
benashz, line 297: Could we use `client.ObjectKeyFromObject(o)` here instead.

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+		}
+		r.eventWatcherRegistry.Delete(name)
+	}
+}
+
+// getEvents calls streamStaticSecretEvents in a loop, collecting and responding
+// to any errors returned.
+func (r *VaultStaticSecretReconciler) getEvents(ctx context.Context, o *secretsv1beta1.VaultStaticSecret, wsClient *vault.WebsocketClient, stoppedCh chan struct{}) error {
+	logger := log.FromContext(ctx)
+	name := types.NamespacedName{Namespace: o.Namespace, Name: o.Name}
```
benashz, line 311: Could we use `client.ObjectKeyFromObject(o)` here instead.

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+			if err != nil {
+				return fmt.Errorf("failed to read from websocket: %w, message: %q",
+					err, string(message))
+			}
+			messageMap := eventMsg{}
+			err = json.Unmarshal(message, &messageMap)
+			if err != nil {
+				return fmt.Errorf("failed to unmarshal event message: %w", err)
+			}
+			logger.V(consts.LogLevelTrace).Info("Received message", "message type", msgType, "message", messageMap)
```
benashz, line 443: Is it possible that `messageMap` may contain confidential info?

benashz, line 443: nit: could we fold this long line?

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+			modified := messageMap.Data.Event.Metadata.Modified
+
+			if modified == "true" {
```
benashz, line 446: Should we validate the string boolean modified?
e.g:
```
			m, err := strconv.ParseBool(modified) ...
```

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+								Namespace: o.Namespace,
+								Name:      o.Name,
+							},
+						},
+					}
+				}
+			} else {
+				// This is an event we're not interested in, ignore it and
+				// carry on.
+				continue
```
benashz, line 471: Could it be beneficial to add a trace level log call here?

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+			meta.Cancel()
+		}
+		r.eventWatcherRegistry.Delete(name)
+	}
+}
+
+// getEvents calls streamStaticSecretEvents in a loop, collecting and responding
+// to any errors returned.
+func (r *VaultStaticSecretReconciler) getEvents(ctx context.Context, o *secretsv1beta1.VaultStaticSecret, wsClient *vault.WebsocketClient, stoppedCh chan struct{}) error {
+	logger := log.FromContext(ctx)
```
benashz, line 310: Could we name this logger?

benashz (commented):
This is looking really great! I had some questions/comments for your consideration.

controllers/vaultstaticsecret_controller.go
benashz, line 1: I wonder if it would be possible to add some Prometheus instrumentation to any of the new events stuff? That can be done as a follow-on PR.

controllers/vaultstaticsecret_controller.go
tvoran: Yeah, it's used in agent and vault's CLI: https://github.com/search?q=org%3Ahashicorp+%22nhooyr.io%2Fwebsocket%22+repo%3Ahashicorp%2Fvault&type=code

controllers/event_watcher_registry.go
tvoran: It might be something to consider? I grabbed it since it's used a few other places in HashiCorp, and in [Vault in particular](https://github.com/search?q=org%3Ahashicorp+github.com%2Fpatrickmn%2Fgo-cache+repo%3Ahashicorp%2Fvault&type=code).

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +245,242 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+eventLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			logger.V(consts.LogLevelDebug).Info("Context done, stopping getEvents",
+				"namespace", o.Namespace, "name", o.Name)
+			return nil
+		default:
+			if shouldBackoff {
+				time.Sleep(time.Second * 10)
```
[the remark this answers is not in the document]
tvoran: Sure, I used an exponential backoff here 2b28b51 with the same options as the Vault reads in the VSS Reconcile(), which includes a jitter factor I think.

controllers/event_watcher_registry.go
tvoran: Sure, and I un-exported it in 2b28b51 since it wasn't being used outside the package.

controllers/vaultstaticsecret_controller.go
tvoran: Good point, I added a wait with a timeout in 2b28b51.

controllers/vaultstaticsecret_controller.go
tvoran: I think so. It wasn't actually returning any errors so I removed that return in 2b28b51.

controllers/vaultstaticsecret_controller.go
tvoran: I think we're pretty safe here since it's just the Path, Namespace, and Modified fields from the event message.

controllers/vaultstaticsecret_controller.go
tvoran: Sure, added one in 2b28b51

controllers/vaultstaticsecret_controller.go
benashz: :+1:

controllers/vaultstaticsecret_controller.go
```diff
@@ -218,11 +246,255 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
+func (r *VaultStaticSecretReconciler) getEvents(ctx context.Context, o *secretsv1beta1.VaultStaticSecret, wsClient *vault.WebsocketClient, stoppedCh chan struct{}) {
+	logger := log.FromContext(ctx).WithName("getEvents")
+	name := client.ObjectKeyFromObject(o)
+	defer func() {
+		r.eventWatcherRegistry.Delete(name)
+		close(stoppedCh)
+	}()
+
+	// Use the same backoff options used for Vault reads in Reconcile()
+	retryBackoff := backoff.NewExponentialBackOff(r.BackOffRegistry.opts...)
```
benashz, line 323: probably should call `retryBackoff.Reset()`

benashz (approved):
Looks great! Only one suggestion to address.

benashz (approved):
Looks great! Only one suggestion to address.

controllers/vaultstaticsecret_controller.go
tvoran: It looks like backoff.NewExponentialBackOff() already calls Reset()? https://github.com/cenkalti/backoff/blob/v4.3.0/exponential.go#L100

PR #781, opened 2024-05-27
Fix versioned build

benashz:
It looks like we were not passing in the VSO version for the build workflow correctly. This PR should fix the versioned builds workflow.


thyton (approved):
LGTM

benashz:
Thanks!

PR #784, opened 2024-05-29
[0.7.0] fix typo in values.yaml for enabling servicemonitor for telemetry

rowansmithhc:
In 0.6.0 this was correct:

https://github.com/hashicorp/vault-secrets-operator/blob/v0.6.0/chart/values.yaml#L639

>    enabled: false

In 0.7.0 this changed:

https://github.com/hashicorp/vault-secrets-operator/blob/v0.7.0/chart/values.yaml#L724
>    nabled: false

In the current state it's not possible to enable the servicemonitor:

```
➜  vault-secrets-operator git:(main) ✗ grep -A 3 "Enable deployment of the Vault Secrets Operator ServiceMonitor CustomResource." values.yaml; helm template vso . | grep -i servicemonitor
    # Enable deployment of the Vault Secrets Operator ServiceMonitor CustomResource.
    # @type: boolean
    nabled: true

➜  vault-secrets-operator git:(main) ✗ grep -A 3 "Enable deployment of the Vault Secrets Operator ServiceMonitor CustomResource." values.yaml; helm template vso . | grep -i servicemonitor
    # Enable deployment of the Vault Secrets Operator ServiceMonitor CustomResource.
    # @type: boolean
    enabled: true

# Source: vault-secrets-operator/templates/prometheus-servicemonitor.yaml
kind: ServiceMonitor
```

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -721,7 +721,7 @@ telemetry:
 
     # Enable deployment of the Vault Secrets Operator ServiceMonitor CustomResource.
     # @type: boolean
-    nabled: false
+    enabled: false
 
     # Selector labels to add to the ServiceMonitor.
     # When empty, defaults to:


benashz:
Closing in favor of #786. Thank you for your contribution to HashiCorp!

tvoran:
@rowansmithhc I don't think this should prevent enabling the servicemonitor in 0.7.0? The parameters in values.yaml are defaults, though user-provided overrides aren't restricted to the parameters in values.yaml. So even though `telemetry.serviceMonitor.enabled` isn't listed in values.yaml, it can still be set on a helm deployment and the templated yaml will pick it up:

```
$ helm template --version 0.7.0 --set 'telemetry.serviceMonitor.enabled=true' hashicorp/vault-secrets-operator | grep -i ServiceMonitor
# Source: vault-secrets-operator/templates/prometheus-servicemonitor.yaml
kind: ServiceMonitor
```

rowansmithhc:
@tvoran yeah, that is correct. The use case I was referencing is not doing a command line based install, instead the chart is cloned/pushed to git and deployed via ArgoCD, so it would reference what is in the actual chart file in git, not what is in the stock helm chart file.

benashz:
Released in https://github.com/hashicorp/vault-secrets-operator/releases/tag/v0.7.1

PR #789, opened 2024-05-30
Helm: add support for auto upgrading CRDs

benashz:
Introduces a pre-upgrade hook that will upgrade/create any of the CRD manifests that are bundled in the vso docker image.

Other fixes:
- make the component label test more reliable by checking all documents individually.
- configure a backoff in the pre-delete "pdcc" Job.

Overview of the approach:

- Extend VSO to support upgrading the CRDs directly
- Include the CRD manifests in the VSO image
- Extend the Helm chart to deploy a pre-upgrade hook that invokes the upgrade process

Out of scope:
- support for rollback/downgrades.
- removal of obsolete CRDs

Update to the Helm values:
```yaml
# Configure the behaviour of Helm hooks.
hooks:
  # Resources common to all hooks.
  resources:
    limits:
      cpu: 500m
      memory: 128Mi
    requests:
      cpu: 10m
      memory: 64Mi
  # Configure the Helm pre-upgrade hook that handles custom resource definition (CRD) upgrades.
  upgradeCRDs:
    # Set to true to automatically upgrade the CRDs.
    # Disabling this will require manual intervention to upgrade the CRDs, so it is recommended to
    # always leave it enabled.
    # @type: boolean
    enabled: true

    # Limit the number of retries for the CRD upgrade.
    # @type: integer
    backoffLimit:  5

    # Set the timeout for the CRD upgrade. The operation should typically take less than 5s
    # to complete.
    # @type: string
    executionTimeout: 30s
```

### DIFF
--- a/chart/templates/hook-upgrade-crds.yaml
+++ b/chart/templates/hook-upgrade-crds.yaml
@@ -1,0 +1,111 @@
+{{- if .Values.hooks.upgradeCRDs.enabled -}}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "vso.chart.fullname" . }}-upgrade-crds
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/component: controller-manager
+{{ include "vso.chart.labels" . | indent 4 }}
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+    helm.sh/hook-weight: "1"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "vso.chart.fullname" . }}-upgrade-crds
+  labels:
+    app.kubernetes.io/component: rbac
+{{ include "vso.chart.labels" . | indent 4 }}
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+    helm.sh/hook-weight: "2"
+rules:
+  - apiGroups:
+    - apiextensions.k8s.io
+    resources:
+    - customresourcedefinitions
+    verbs:
+    - create
+    - delete
+    - get
+    - list
+    - patch
+    - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "vso.chart.fullname" . }}-upgrade-crds
+  labels:
+    app.kubernetes.io/component: rbac
+{{ include "vso.chart.labels" . | indent 4 }}
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+    helm.sh/hook-weight: "2"
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "vso.chart.fullname" . }}-upgrade-crds
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: {{ template "vso.chart.fullname" . }}-upgrade-crds
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ printf "%s-%s" "upgrade-crds" (include "vso.chart.fullname" .) | trunc 63 | trimSuffix "-" }}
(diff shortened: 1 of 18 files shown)


test/chart/chart_test.go
```diff
@@ -0,0 +1,282 @@
...
+			assert.Greater(t, updatedCRD.Generation, o.Generation)
+			assert.Equal(t, o.UID, updatedCRD.UID)
+		} else {
+			assert.Equal(t, int64(1), updatedCRD.Generation)
+		}
+
+		assert.Equal(t, wantCRD.Spec, updatedCRD.Spec, "CRD %q .spec mismatch", wantCRD.Name)
+		assert.Equal(t, wantCRD.Labels, updatedCRD.Labels, "CRD %q .metadata.labels mismatch", wantCRD.Name)
+		assert.Equal(t, wantCRD.Annotations, updatedCRD.Annotations, "CRD %q .metadata.annotations mismatch", wantCRD.Name)
+	}
```
tvoran, line 219: Might be good to compare `wantCRD.Status` as well if possible.

internal/utils/utils.go
```diff
@@ -52,3 +62,102 @@ func GetOwnerRefFromObj(owner ctrlclient.Object, scheme *runtime.Scheme) (metav1
...
+			}
+		} else {
+			logger.Info("Patching",
+				"name", crd.Name, "gvk", crd.GroupVersionKind(),
+				"uid", cur.GetUID(), "resourceVersion", cur.GetResourceVersion(),
+			)
+			patch := ctrlclient.MergeFrom(cur.DeepCopy())
+			cur.Spec = crd.Spec
+			cur.ObjectMeta.Annotations = crd.ObjectMeta.Annotations
+			cur.ObjectMeta.Labels = crd.ObjectMeta.Labels
```
tvoran, line 157: Any need to patch `cur.Status` as well?

internal/utils/utils_test.go
```diff
@@ -72,3 +82,217 @@ func TestGetCurrentNamespace(t *testing.T) {
...
+	builder := fake.NewClientBuilder().WithScheme(scheme)
+	tests := []struct {
+		name           string
+		c              client.Client
+		create         []*apiextensionsv1.CustomResourceDefinition
+		createDir      bool
+		manifests      []string
+		otherManifests []string
+		subDirs        int
+		canaries       int
```
tvoran, line 180: What are `canaries` in these tests? It looks like it's always set when `subDirs` is set in the test cases.

chart/templates/hook-upgrade-crds.yaml
```diff
@@ -0,0 +1,116 @@
...
+    - apiextensions.k8s.io
+    resources:
+    - customresourcedefinitions
+    verbs:
+    - create
+    - delete
+    - get
+    - list
+    - patch
+    - update
```
tvoran, line 43: Random thought: is there a way to limit this to just our CRDs? At least for the write verbs.

tvoran (approved):
Left a few minor questions, and I'm wondering if we need to check the `Status` block on upgrades, but on the whole 👍

internal/utils/utils.go
benashz: I don't think it should be necessary. The status should automatically be updated after we apply the patch.

chart/templates/hook-upgrade-crds.yaml
benashz: We would probably have to `resourceNames` statically to get something like what you are after here. That might be a bit brittle since we would have to update the set of names whenever we add new CRDs. Since this ClusterRole is created and destroyed for the lifetime of the hook, I think we are probably pretty safe here.

test/unit/hook-upgrade-crds.bats
```diff
@@ -0,0 +1,207 @@
...
+
+    local job
+    job="$(echo "${object}" | yq 'select(di == 3)' | tee /dev/stderr)"
+    [ "$(echo "${job}" | \
+      yq '.spec.backoffLimit == 10')" = "true" ]
+    [ "$(echo "${job}" | yq '.kind == "Job"')" = "true" ]
+    [ "$(echo "${job}" | \
+      yq '(.spec.template.spec.containers | length) == "1"')" = "true" ]
+    [ "$(echo "${job}" | \
+      yq '(.spec.template.spec.containers | length) == "1"')" = "true" ]
```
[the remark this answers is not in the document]
benashz: Fixed in 9b5c3a12a05ea0cce3a8938cbc98bbc4c2662de4

test/chart/chart_test.go
benashz: Fixed in 9b5c3a12a05ea0cce3a8938cbc98bbc4c2662de4

PR #798, opened 2024-06-04
Give concurrent lockers more time to get the lock

benashz:
Should resolve issues like: https://github.com/hashicorp/vault-secrets-operator/actions/runs/9370160932/job/25796550858#step:4:1943


thyton (approved):
LGTM

PR #799, opened 2024-06-04
CI: add done job for branch protection status checks

benashz:
Adds a job to denote the success/failure of the build workflow. Once this update is in we can make it a required check under the branch protections for `main`


thyton (approved):
LGTM!

PR #800, opened 2024-06-04
VaultAuth: extend the global ref API

benashz:
Extends the configuration options that can be applied during the VaultAuth+VaultAuthGlobal merger.

Examples:

```yaml
---
kind: VaultAuth
metadata:
  name: default
spec:
  vaultAuthGlobalRef:
    mergeStrategy:
      headers: union
      params: union
    name: default
---
kind: VaultAuthGlobal
metadata:
  name: default
spec:
  defaultAuthMethod: kubernetes
  kubernetes:
    audiences:
    - vault
    mount: demo-auth-mount
    namespace: demo-ns
    role: auth-role
    serviceAccount: default
    tokenExpirationSeconds: 600
```


thyton (approved):
LGTM. Really nice tests!

PR #801, opened 2024-06-05
Core: update backoff option docs

benashz:
Other fixes:
- call backoff.Reset() when creating a new BackOff(). That will ensure that the initial interval is honoured
- check that the backoff multiplier is greater than zero

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -181,6 +181,15 @@ controller:
 
     # Backoff settings for the controller manager. These settings control the backoff behavior
     # when the controller encounters an error while fetching secrets from the SecretSource.
+    # For example given the following settings:
+    #   initialInterval: 5s
+    #   maxInterval: 60s
+    #   randomizationFactor: 0.5
+    #   multiplier: 1.5
+    #
+    # The backoff retry sequence might be something like:
+    #  5.5s, 7.5s, 11.25s, 16.87s, 25.3125s, 37.96s, 56.95, 60.95s...
+    # @type: object
     backoffOnSecretSourceError:
       # Initial interval between retries.
       # @type: duration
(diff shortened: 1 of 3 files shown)


controllers/registry.go
```diff
@@ -262,8 +262,11 @@ func (r *BackOffRegistry) Get(objKey client.ObjectKey) (*BackOff, bool) {
 
 	entry, ok := r.m[objKey]
 	if !ok {
+		bo := backoff.NewExponentialBackOff(r.opts...)
+		// call Reset() to ensure that the initial interval is honoured.
+		bo.Reset()
```
tvoran, line 267: Do we still need the wrapper for backoff.Backoff on line 279? The comment there says it purposefully doesn't implement Backoff.Reset.

tvoran (approved):

controllers/registry.go
benashz: I had considered it. I am okay to leave it as-is since it may still prove useful if we ever want to add instrumentation.

chart/values.yaml
```diff
@@ -181,20 +181,37 @@ controller:
...
     # Backoff settings for the controller manager. These settings control the backoff behavior
     # when the controller encounters an error while fetching secrets from the SecretSource.
+    # For example given the following settings:
+    #   initialInterval: 5s
+    #   maxInterval: 60s
+    #   randomizationFactor: 0.5
+    #   multiplier: 1.5
+    #
+    # The backoff retry sequence might be something like:
+    #  5.5s, 7.5s, 11.25s, 16.87s, 25.3125s, 37.96s, 56.95, 60.95s...
```
tvoran, line 191: BTW, the rendered .mdx for the website isn't going to preserve these line breaks and formatting, so we may want to do something like:

````yaml
    # For example given the following settings:
    # ```yaml
    #   initialInterval: 5s
    #   maxInterval: 60s
    #   randomizationFactor: 0.5
    #   multiplier: 1.5
    # ```
    #
    # The backoff retry sequence might be something like:
    #  5.5s, 7.5s, 11.25s, 16.87s, 25.3125s, 37.96s, 56.95, 60.95s...
````

PR #805, opened 2024-06-07
CI: let unit-tests run in parallel to the build

benashz:
Since the build-done job determines pass/fail, we can update the build dependency to no longer require unit-tests. That allows us to get the results of the both integration unit tests independently. It also reduces the total workflow execution time by ~2m

Other fixes:
- fail build-done job on workflow cancellation.


thyton (approved):
LGTM!

PR #815, opened 2024-06-12
Core: delete resource status metrics upon object deletion

benashz:
Previously, VSO would orphan the `controller_resource_status` metric for CRs that have been deleted from the cluster. The fix is to delete the metric for the corresponding controller, namespace, and object name.


thyton (approved):
LGTM!

PR #817, opened 2024-06-13
Drop the tainted client gauge.

benashz:
In hindsight this metric doesn't make sense at this point. It also introduces another high cardinality metric that users probably don't want to have to manage. This metric is better implemented as a collector on the client cache itself, setting the total number of tainted clients would be more meaningful.


thyton (approved):
LGTM

PR #832, opened 2024-06-24
Add user-defined pre-delete hook annotation and labels for job

alextricity25:


chart/values.yaml
```diff
@@ -141,6 +141,14 @@ controller:
   # This value defines additional annotations for the deployment. This should be formatted as a YAML object (map)
   annotations: {}
 
+  # Controller pre-delete hook labels and annotations
```
benashz, line 144: It would be great if we could move the hooks configuration to https://github.com/hashicorp/vault-secrets-operator/blob/67fce39ae4a4929a712211e6a5908fbf5fa14236/chart/values.yaml#L786 .

benashz (commented):
This is looking good.  We will need to add some tests for it, though. It would be good to move the configuration to the `hooks` specific values  as well.

You can take a look at https://github.com/hashicorp/vault-secrets-operator/blob/main/test/unit/hook-upgrade-crds.bats for sample of what the tests might look like.

PR #850, opened 2024-07-12
Rejig VA status conditions

benashz:
Updates VaultAuthGlobal conditions to more in line with the other core K8s resources.


thyton (approved):
LGTM!

PR #854, opened 2024-07-16
VaultAuth: set valid status on VaultAuthGlobal deref error

benashz:
Other fixes:
- For all applicable types, make .Status.Types a boolean pointer so that it can be stored as false


controllers/hcpauth_controller.go
```diff
@@ -13,6 +13,7 @@ import (
 	hvsclient "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-06-13/client"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
```
tvoran, line 16: FWIW, my IDE is telling me [k8s.io/utils/pointer](https://pkg.go.dev/k8s.io/utils/pointer) is deprecated:

> "k8s.io/utils/pointer" is deprecated: Use functions in k8s.io/utils/ptr instead: ptr.To to obtain a pointer, ptr.Deref to dereference a pointer, ptr.Equal to compare dereferenced pointers.

tvoran (approved):

controllers/hcpauth_controller.go
benashz: Interesting. It looks like we are using the deprecated package in other places. I will open a separate migration PR. Thanks!

benashz: Migration PR: https://github.com/hashicorp/vault-secrets-operator/pull/856

PR #862, opened 2024-07-19
VC: add spec.timeout for Vault requests

benashz:
Extends the VaultConnection API to include the support for setting the Vault request time. It will be applied to all Vault Clients for the given VaultConnection.


thyton (approved):
LGTM

PR #873, opened 2024-07-26
Helm(hooks): honor imagePullPolicy and imagePullSecrets

benashz:
Update all Helm hook Jobs to Honor the the manager's `imagePullPolicy` and `imagePullSecrets`.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -194,9 +194,14 @@ spec:
       serviceAccountName: {{ include "vso.chart.fullname" . }}-controller-manager
       securityContext:
         {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+      {{- with .Values.controller.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
       - name: pre-delete-controller-cleanup
         image: {{ .Values.controller.manager.image.repository }}:{{ .Values.controller.manager.image.tag }}
+        imagePullPolicy: {{ .Values.controller.manager.image.pullPolicy }}
         args:
         - --uninstall
         - --pre-delete-hook-timeout-seconds={{ .Values.controller.preDeleteHookTimeoutSeconds }}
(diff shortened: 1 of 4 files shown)


chart/templates/deployment.yaml
```diff
@@ -194,9 +194,14 @@ spec:
       serviceAccountName: {{ include "vso.chart.fullname" . }}-controller-manager
       securityContext:
         {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+      {{- with .Values.controller.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
```
tvoran, line 200: The image pull secrets are already set on the deployment's service account (line 14), so I don't think this is necessary.

chart/templates/hook-upgrade-crds.yaml
```diff
@@ -88,9 +88,14 @@ spec:
       serviceAccountName: {{ template "vso.chart.fullname" . }}-upgrade-crds
       securityContext:
         {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+      {{- with .Values.controller.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
```
tvoran, line 94: How about setting the pull secrets on the service account, like we do for [the deployment's service account](https://github.com/hashicorp/vault-secrets-operator/blob/be6ff476c51d98aa78b3290e62ac3271813df38e/chart/templates/deployment.yaml#L14)?

i.e.

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ template "vso.chart.fullname" . }}-upgrade-crds
  labels:
{{ include "vso.chart.labels" . | indent 4 }}
  annotations:
    helm.sh/hook: pre-upgrade
    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
    helm.sh/hook-weight: "1"
{{ include "vso.imagePullSecrets" .}}

chart/templates/deployment.yaml
benashz: Reverted in c1dfa0095f554f9f65e94605cbcbf0e2ea0b71a3

chart/templates/hook-upgrade-crds.yaml
benashz: 👍 updated in c1dfa0095f554f9f65e94605cbcbf0e2ea0b71a3

tvoran (approved):

PR #891, opened 2024-08-14
Setup test infra in AWS

Zlaticanin:
This PR involves adding a testing infrastructure for AWS. It creates an EKS cluster and deploys VSO and Vault.

Prereq: Doormat AWS sandbox account

Before running any make commands, make sure to log in into doormat and export the AWS keys:
`doormat login && eval $(doormat aws -a <AWS ACCOUNT NAME> export)`

Make sure to export the path to your vault license:
`export TF_VAR_vault_license_path=<path to license file>`

Then to set up EKS cluster and deploy Vault and VSO:
`make -f scale-testing.mk create-eks`
`make -f scale-testing.mk deploy-workload`


test/integration/infra/scale-testing/main.tf
```diff
@@ -0,0 +1,136 @@
...
+
+module "vso-helm" {
+  source                       = "../../../modules/vso-helm"
+  operator_namespace           = var.operator_namespace
+  operator_image_repo          = var.operator_image_repo
+  operator_image_tag           = var.operator_image_tag
+  enable_default_connection    = var.enable_default_connection
+  enable_default_auth_method   = false
+  operator_helm_chart_path     = var.operator_helm_chart_path
+  k8s_vault_connection_address = var.k8s_vault_connection_address
```
tvoran, line 118: Is this being set somewhere? Looks like it defaults to `""`, and I keep getting an error about a missing address here:

```
│ Error: 1 error occurred:
│ 	* VaultConnection.secrets.hashicorp.com "default" is invalid: spec.address: Required value
```

test/integration/infra/scale-testing/variables.tf
```diff
@@ -0,0 +1,202 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+variable "kubernetes_version" {
+  type    = string
+  default = "1.26"
```
tvoran, line 6:
```suggestion
  default = "1.30"
```

test/integration/infra/scale-testing/variables.tf
```diff
@@ -0,0 +1,202 @@
...
+  default = "1.17-ent"
+}
+
+variable "vault_enterprise" {
+  type    = bool
+  default = true
+}
+
+variable "vault_chart_version" {
+  default = "0.27.0"
```
tvoran, line 190:
```suggestion
  default = "0.28.1"
```

test/integration/modules/vault/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+  default = "1.17-ent"
+}
+
+variable "vault_enterprise" {
+  type    = bool
+  default = true
+}
+
+variable "vault_chart_version" {
+  default = "0.27.0"
```
tvoran, line 50:
```suggestion
  default = "0.28.1"
```

test/integration/infra/scale-testing/main.tf
Zlaticanin: Yes, should be set to false. I updated it.

scale-testing.mk
```diff
@@ -0,0 +1,32 @@
...
+.PHONY: create-eks
+create-eks: ## Create a new EKS cluster
+	@mkdir -p $(TF_EKS_STATE_DIR)
+	rm -f $(TF_EKS_STATE_DIR)/*.tf
+	cp -v $(TF_EKS_SRC_DIR)/*.tf $(TF_EKS_STATE_DIR)/.
+	$(TERRAFORM) -chdir=$(TF_EKS_STATE_DIR) init -upgrade
+	$(TERRAFORM) -chdir=$(TF_EKS_STATE_DIR) apply -auto-approve \
+		-var region=$(AWS_REGION) \
+		-var kubernetes_version=$(EKS_K8S_VERSION) || exit 1
+	rm -f $(TF_EKS_STATE_DIR)/*.tfvars
```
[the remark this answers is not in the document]
Zlaticanin: Good idea, thank you! I added it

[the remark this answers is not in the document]
Zlaticanin: I am having some issues with `destroy-eks` target tho. I am getting insufficient permissions to destroy it. Any suggestions on how to resolve, considering that we can't update permissions on our doormat account

test/integration/infra/scale-testing/providers.tf
```diff
@@ -0,0 +1,26 @@
...
+  token                  = data.aws_eks_cluster_auth.cluster.token
+}
+
+provider "helm" {
+  kubernetes {
+    host                   = data.aws_eks_cluster.cluster.endpoint
+    cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority[0].data)
+    token                  = data.aws_eks_cluster_auth.cluster.token
+  }
+}
```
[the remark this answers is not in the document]
Zlaticanin: I've ran into this as well, but refreshing the AWS keys and re-running `create-eks` worked. 

I haven't considered this, but if folks think that's a better approach, I am happy to update it.

test/integration/infra/scale-testing/providers.tf
```diff
@@ -0,0 +1,26 @@
...
+    aws = {
+      source  = "hashicorp/aws"
+      version = "5.49.0"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = "2.16.1"
+    }
+  }
+}
```
tvoran, line 12: We should probably list the `helm` provider version here too.

test/integration/infra/scale-testing/providers.tf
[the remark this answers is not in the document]
tvoran: Yeah separating them seems to be the recommended approach: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources

tvoran, line 26: Looks like we also need to use the region variable here somewhere?

```suggestion
}

provider "aws" {
  region = var.region
}
```

scale-testing.mk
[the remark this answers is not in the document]
tvoran: I've been having issues with `destroy-eks` failing too, though it seems to get stuck on deleting k8s resources or one of the helm charts. I don't think it's actually a AWS permissions issue, since removing any of the `kubernetes_*` or `helm_release` resources from the terraform.tfstate file allows it to proceed correctly for me.

test/integration/modules/vault/metrics-server.tf
tvoran, line 1: I think `metrics-server` and `kube-prometheus` could live in their own module? Since they aren't directly related to vault.

test/integration/infra/scale-testing/providers.tf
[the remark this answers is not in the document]
Zlaticanin: I tried adding some dependencies to wait for the EKS cluster to be up and running before we try to deploy anything, but I didn't have any luck making it work. I went ahead and separated them into two different runs

test/integration/modules/vault/metrics-server.tf
Zlaticanin: I moved them to a new module. I am not entirely sure how we plan on using metrics here. I don't have much experience at all with that. So please let me know if there is something else I should do.

scale-testing.mk
```diff
@@ -0,0 +1,63 @@
...
+	@echo Getting license from VAULT_LICENSE
+	$(eval _VAULT_LICENSE=$(shell printenv VAULT_LICENSE))
+else ifdef VAULT_LICENSE_PATH
+	@echo Getting license from VAULT_LICENSE_PATH
+	$(eval _VAULT_LICENSE=$(shell cat $(VAULT_LICENSE_PATH)))
+	@test -n "$(_VAULT_LICENSE)" || ( echo vault license is empty; exit 1)
+else
+	$(error no valid vault license source provided, choices are VAULT_LICENSE_CI, VAULT_LICENSE, VAULT_LICENSE_PATH)
+endif
+endif
```
tvoran, line 33: Do we need another set-vault-license target here? Does the `include ./Makefile` not bring it in from the main makefile?

test/integration/infra/scale-testing/providers.tf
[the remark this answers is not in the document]
tvoran: It's working better for me too!

scale-testing.mk
```diff
@@ -0,0 +1,63 @@
...
+
+.PHONY: create-eks
+create-eks: set-vault-license ## Create a new EKS cluster
+	@mkdir -p $(TF_EKS_STATE_DIR)
+ifeq ($(VAULT_ENTERPRISE), true)
+    ## ensure that the license is *not* emitted to the console
+	@echo "vault_license = \"$(_VAULT_LICENSE)\"" > $(TF_EKS_STATE_DIR)/license.auto.tfvars
+endif
```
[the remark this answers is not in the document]
Zlaticanin: Yeah, that makes sense. I can update.
I put it as a dependency on deploy-workload, so that we can only run that make target and it will spin up an eks cluster and deploy vso and vault.

scale-testing.mk
Zlaticanin: We don't, you are right. Should be included, Iwill remove. Thank you

scale-testing.mk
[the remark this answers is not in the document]
Zlaticanin: Is there a way where we can set make targets to remove those resources before destroying the cluster? I am having issues with the destroy targets :/

test/integration/infra/scale-testing/deployments/providers.tf
```diff
@@ -0,0 +1,34 @@
...
+  backend = "local"
+
+  config = {
+    path = "../../eks-cluster/state/terraform.tfstate"
+  }
+}
+
+provider "kubernetes" {
+  host                   = data.terraform_remote_state.eks.outputs.cluster_endpoint
+  cluster_ca_certificate = base64decode(data.terraform_remote_state.eks.outputs.cluster_certificate_authority)
```
thyton, line 24: If we intend to scale-test on other clusters beside EKS (GKE and AKS), I think we probably should create variables for cluster-specific values (e.g. `cluster_ca_certificate`) for computing its value based on the cluster type we use. We could refactor in a different PR. We can have an env var to the cluster type for scale-testing in `scale-testing.mk`.

scale-testing.mk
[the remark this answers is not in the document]
Zlaticanin: Splitting them into two runs resolves this issue. I can now destroy it without any issues.

scale-testing.mk
```diff
@@ -0,0 +1,45 @@
...
+
+include ./Makefile
+
+.PHONY: create-eks
+create-eks: set-vault-license ## Create a new EKS cluster
+	@mkdir -p $(TF_EKS_STATE_DIR)
+ifeq ($(VAULT_ENTERPRISE), true)
+    ## ensure that the license is *not* emitted to the console
+	@echo "vault_license = \"$(_VAULT_LICENSE)\"" > $(TF_EKS_STATE_DIR)/license.auto.tfvars
+endif
```
tvoran, line 23: Do we still need `set-vault-license` on the eks target at all? Looks like we'll want to move the `@echo "vault_license ...` block down into `deploy-workload` as well?

test/integration/infra/scale-testing/deployments/main.tf
```diff
@@ -0,0 +1,30 @@
...
+
+  manager_extra_args = [
+    "-min-refresh-after-hvsa=3s",
+    "-zap-log-level=6"
+  ]
+}
+
+module "vault" {
+  source               = "../../../../modules/vault"
+  vault_license_path   = var.vault_license_path
```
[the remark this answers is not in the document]
Zlaticanin: Good idea! I will add it

scale-testing.mk
Zlaticanin: Good catch! We don't need `set-vault-license` on that target anymore.  I will update and move the license block. Thank you 🙏

tvoran (commented):
👍 

Sometime in the future we may want to look at combining the EKS terraform in `test/integration/infra/eks/` with the scale testing eks setup.

tvoran (approved):

scale-testing.mk
```diff
@@ -32,6 +28,10 @@ endif
 .PHONY: deploy-workload
 deploy-workload: set-vault-license ## Deploy the workload to the EKS cluster
 	@mkdir -p $(TF_DEPLOY_STATE_DIR)
+ifeq ($(VAULT_ENTERPRISE), true)
+    ## ensure that the license is *not* emitted to the console
+	@echo "vault_license = \"$(_VAULT_LICENSE)\"" > $(TF_EKS_STATE_DIR)/license.auto.tfvars
```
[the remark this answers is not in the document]
Zlaticanin: Ah yes, good catch again! Thank you 🙏

PR #889, opened 2024-08-14
HVS: update to 2023-11-28 API

tvoran:
Updates hashicorp/hcp-sdk-go to v0.160.0 and switches to the [2023-11-28 API](https://github.com/hashicorp/hcp-sdk-go/blob/v0.106.0/clients/cloud-vault-secrets/preview/2023-11-28/models/secrets20231128_open_secret.go#L20).

(This is a prerequisite to supporting rotating and dynamic secret types in HVS.)

PR #893, opened 2024-08-16
HVS: rotating secret support

tvoran:
Initial support for rotating secret types in HVS. Fetches them along with static secrets on the same open app secrets API call.

Depends on #889 

Integration tests will be added in a future PR.


thyton (approved):
LGTM

tvoran:
Thanks!

PR #898, opened 2024-08-27
VSS/instant-updates: more stable event watcher

tvoran:
Pass a dereferenced VSS object to the vault event watcher goroutine to avoid the `EventWatcherStarted` k8s event being emitted without a name or namespace. Also wait for the `EventWatcherStarted` k8s event before proceeding with the instant updates tests, to ensure the Vault event subscription is setup before the tests proceed with modifying Vault.

Disclaimer: I'm not totally sure why passing a dereferenced VSS object seems to fix the issue. I thought maybe it was something to do with garbage collection, but I haven't been able to confirm that suspicion.


thyton (approved):
LGTM

tvoran:
Thanks!

PR #906, opened 2024-09-05
VC: update spec.timeout to be a string

benashz:
Previously, spec.timeout was a *metav1.Duration. This caused some challenges when creating/updating VaultConnection instances via VSO's API, when the value was 60s or greater the marshalled value would no longer be in valid duration notation and the API request would fail during field validation.


HamzaShili65 (approved):
LGTM!

PR #909, opened 2024-09-10
[PSA-2298] bump kube-rbac-proxy to 0.18.1

mcollao-hc:
Bumping kube-rbac-proxy to 0.18.1 which solves multiple vulnerabilities.

https://github.com/brancz/kube-rbac-proxy/releases/tag/v0.18.1

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -121,7 +121,7 @@ controller:
     image:
       pullPolicy: IfNotPresent
       repository: gcr.io/kubebuilder/kube-rbac-proxy
-      tag: v0.15.0
+      tag: v0.18.1
 
     # Configures the default resources for the kube rbac proxy container.
     # For more information on configuring resources, see the K8s documentation:
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -120,7 +120,7 @@ controller:
     # Image sets the repo and tag of the kube-rbac-proxy image to use for the controller.
     image:
       pullPolicy: IfNotPresent
-      repository: gcr.io/kubebuilder/kube-rbac-proxy
+      repository: quay.io/brancz/kube-rbac-proxy
       tag: v0.18.1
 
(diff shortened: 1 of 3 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -19,7 +19,7 @@ spec:
         # capabilities:
         #   drop:
         #     - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.18.1
```
benashz, line 22: That should be:
```suggestion
        image: quay.io/brancz/kube-rbac-proxy:v0.18.1
```

chart/values.yaml
```diff
@@ -121,7 +121,7 @@ controller:
     image:
       pullPolicy: IfNotPresent
       repository: gcr.io/kubebuilder/kube-rbac-proxy
```
benashz, line 123:
```suggestion
      repository: quay.io/brancz/kube-rbac-proxy
```

benashz (commented):
Thanks for the contribution. It looks like we also need to update the repository. The tag you have specified does not exist on `gcr.io`. I have provided some suggested updates to address that.

We also need to to update the version here: https://github.com/hashicorp/vault-secrets-operator/blob/649d7cff7f18541da6c673c46b6a72362fc152e4/Makefile#L7

config/default/manager_auth_proxy_patch.yaml
mcollao-hc: Changed, thanks!

chart/values.yaml
mcollao-hc: Changed, thanks!

benashz (approved):
LGTM!

PR #917, opened 2024-09-13
HVS: basic dynamic secrets support 

tvoran:
Fetches all dynamic secret key/value pairs along with the rest of the secrets in an HVS App, every `RefreshAfter` or at the `renewalPercent` of the TTL of the dynamic secrets, whichever comes first.

Adds a `SyncConfig` for HVS to the HCPVaultSecretsAppSpec to allow specifying a custom `renewalPercent` for dynamic secrets, otherwise the default is 67% of the TTL.

```yaml
kind: HCPVaultSecretsApp
metadata:
  name: myapp
spec:
  appName: sample-app
  hcpAuthRef: my-hc-org
  refreshAfter: 2h
  destination:
    name: myapp-secrets
    create: true
  syncConfig:
    dynamic:
      renewalPercent: 85
```

Adds the last observed state of each dynamic secret to HCPVaultSecretsAppStatus:

```yaml
...
status:
  dynamicSecrets:
  - createdAt: "2024-09-12T17:02:50.742Z"
    expiresAt: "2024-09-12T18:02:50.742Z"
    name: gcp
    ttl: 3600s
  lastGeneration: 2
  secretMAC: la2TiOJ1XBvPmB1z3T/mINiwc0/AslBfSyOceN9UCtA=
...
```

Note: Caching dynamic secret responses in order to avoid constantly creating new secrets every Reconcile() is planned for a follow-on PR.


api/v1beta1/hcpvaultsecretsapp_types.go
```diff
@@ -32,6 +32,37 @@ type HCPVaultSecretsAppSpec struct {
...
+// HVSSyncConfig configures sync behavior from HVS to VSO
+type HVSSyncConfig struct {
+	// Dynamic configures sync behavior for dynamic secrets.
+	Dynamic *HVSDynamicSyncConfig `json:"dynamic,omitempty"`
+}
+
+// HVSDynamicSyncConfig configures sync behavior for HVS dynamic secrets.
+type HVSDynamicSyncConfig struct {
+	// RenewalPercent is the percent out of 100 of a dynamic secret's TTL when
+	// new secrets are generated. Defaults to 67 percent minus jitter.
```
tvoran, line 48: Note that this is slightly different than [VaultDynamicSecret's renewalPercent](https://github.com/hashicorp/vault-secrets-operator/blob/d31cc625d686817a0f6d5f4f4ceb98b090fe8820/api/v1beta1/vaultdynamicsecret_types.go#L42-L48), since that one defaults to 67 percent *plus* to up 10% jitter.

api/v1beta1/hcpvaultsecretsapp_types.go
```diff
@@ -32,6 +32,37 @@ type HCPVaultSecretsAppSpec struct {
...
 	Destination Destination `json:"destination"`
+	// SyncConfig configures sync behavior from HVS to VSO
+	SyncConfig *HVSSyncConfig `json:"syncConfig,omitempty"`
+}
+
+// HVSSyncConfig configures sync behavior from HVS to VSO
+type HVSSyncConfig struct {
+	// Dynamic configures sync behavior for dynamic secrets.
+	Dynamic *HVSDynamicSyncConfig `json:"dynamic,omitempty"`
+}
```
tvoran, line 43: I debated whether to reuse the existing SyncConfig{} type from VaultStaticSecret, but since HVS is a bit different than Vault I went with a new type here but still named "SyncConfig" in the HVSApp spec.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -319,3 +340,93 @@ func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	fullTTL := time.Time(dynamicInstance.ExpiresAt).Sub(time.Time(dynamicInstance.CreatedAt))
+	renewPoint := fullTTL * time.Duration(renewPercent) / 100
+	renewTime := time.Time(dynamicInstance.CreatedAt).Add(renewPoint)
+	howLongUntilRenewTime := renewTime.Sub(now)
+
+	if howLongUntilRenewTime < requeueAfter {
+		nextRequeue = howLongUntilRenewTime
+	}
+	if nextRequeue < 0 {
+		nextRequeue = 1 * time.Second
```
thyton, line 428: Could we have a constant `defaultNextRequeue` for 1s?

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -131,25 +131,46 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	if err != nil {
-		logger.Error(err, "Get App Secret", "appName", o.Spec.AppName)
+		logger.Error(err, "Get App Secrets", "appName", o.Spec.AppName)
+		entry, _ := r.BackOffRegistry.Get(req.NamespacedName)
+		return ctrl.Result{
+			RequeueAfter: entry.NextBackOff(),
+		}, nil
+	}
+
+	dynamicSecrets, err := getHVSDynamicSecrets(ctx, c, o.Spec.AppName)
```
thyton, line 141: Double checking my understanding: It looks like we don't have a way currently to know if there is any dynamic/static secrets in HVS App Secrets, so VSO blindly requests HVS.

thyton (approved):
LGTM. Nice tests!

api/v1beta1/hcpvaultsecretsapp_types.go
thyton: I think new `HVSSyncConfig` for HVS makes sense.

controllers/hcpvaultsecretsapp_controller.go
tvoran: Yeah so the `c.OpenAppSecrets()` call before this fetches all the static and rotating secrets in the HVS App. Then this function first queries the HVS App secret metadata with `c.ListAppSecrets()`, and then fetches each dynamic secret individually with `c.OpenAppSecret`. Each call to `c.OpenAppSecret` generates and returns a new set of credentials and TTL for the dynamic secret.

We're planning to cache the dynamic secret responses in the future, and then we'll know which dynamic secrets need to be fetched based on their TTLs.

controllers/hcpvaultsecretsapp_controller.go
tvoran: Sure, added one in fd0656f

tvoran:
Thanks!

controllers/hcpvaultsecretsapp_controller.go
thyton: Thank you!

PR #924, opened 2024-09-18
CI: use dependabot groups for Go deps

benashz:
Follows the work done in vault-csi-provider project: https://github.com/hashicorp/vault-csi-provider/blob/f8fdfd37090297e07a2796b44023bb6c02f12793/.github/dependabot.yaml#L15-L22

Hopefully this will reduce the number of Dependabot PRs and make those PRs more reliable in the case where some dependency update is incomplete.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -15,6 +15,14 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      gomod-breaking:
+        update-types:
+          - major
+      gomod-backward-compatible:
+        update-types:
+          - minor
+          - patch
     # Disable version updates and only check security updates for github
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -15,6 +15,14 @@ updates:
...
     schedule:
       interval: "weekly"
+    groups:
+      gomod-breaking:
+        update-types:
+          - major
+      gomod-backward-compatible:
+        update-types:
+          - minor
+          - patch
```
tvoran, line 25: Looks like this should go under the `package-ecosystem: "gomod"` section?

benashz: Oops. Good catch. Should be fixed in 384cf15de492225407e114ed5044f843484d456a

tvoran (approved):

PR #963, opened 2024-11-04
Add controller.extraLabels to pre-upgrade hook job

satyamkapoor:
With release auto CRD upgrade coming in with release [0.8.0](https://github.com/hashicorp/vault-secrets-operator/releases/tag/v0.8.0), extraLabels were missed being honored in the helm pre-upgrade hook job. This PR, add's it.


benashz:
@satyamkapoor thank you for the PR. Would you mind expanding a bit on how these Job labels would be used? The upgrade CRDs job is consistent with https://github.com/hashicorp/vault-secrets-operator/blob/3b751445f42b2a92958b88d2cc725b588b71122f/chart/templates/deployment.yaml#L170, in terms of label configuration.

PR #968, opened 2024-11-11
bump k8s version

hashiblaum:
remove version 1.26.15 and add 1.31.2

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -344,7 +344,7 @@ jobs:
       - run: echo "setting versions"
     outputs:
       # JSON encoded array of k8s versions
-      K8S_VERSIONS: '["1.31.2", "1.30.0", "1.29.4", "1.28.9", "1.27.13"]'
+      K8S_VERSIONS: '["1.31.2", "1.30.6", "1.29.10", "1.28.9", "1.27.13"]'
       VAULT_N: "1.17.2"
       VAULT_N_1: "1.16.6"
       VAULT_N_2: "1.15.12"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -344,7 +344,7 @@ jobs:
       - run: echo "setting versions"
     outputs:
       # JSON encoded array of k8s versions
-      K8S_VERSIONS: '["1.30.0", "1.29.4", "1.28.9", "1.27.13", "1.26.15"]'
+      K8S_VERSIONS: '["1.31.2", "1.30.6", "1.29.10", "1.28.9", "1.27.13"]'
```
tvoran, line 347: It looks like there are some slightly newer versions available now for [1.28](https://hub.docker.com/r/kindest/node/tags?name=1.28) and [1.27](https://hub.docker.com/r/kindest/node/tags?name=1.27):

```suggestion
      K8S_VERSIONS: '["1.31.2", "1.30.6", "1.29.10", "1.28.15", "1.27.16"]'
```

tvoran (commented):
May want to update the kind versions too:

https://github.com/hashicorp/vault-secrets-operator/blob/e9783a0f3e0a48e5ef4aa18357eed71482ae4d5b/.github/workflows/build.yaml#L321

https://github.com/hashicorp/vault-secrets-operator/blob/e9783a0f3e0a48e5ef4aa18357eed71482ae4d5b/.github/actions/integration-test/action.yml#L70

tvoran (approved):

PR #977, opened 2024-11-27
Fix pattern: regex to support days (d)

0megam:
While Vault API supports setting ttl and expiryOffset with d suffix CRD is not supporting that.


0megam:
Could someone please review this PR ?
This blocking us from upgrading our VSO to the latest version, because huge amount of services in our k8s clusters are using days(d) in expiryOffset parameter and we do not want to maintain our own version of CRDs.

benashz:
In order to maintain consistency between the other types, specifying the `expiryOffset` in days is not supported. The Vault PKI API supports setting a certificate TTL with the units in days. That is addressed in #990.

PR #980, opened 2024-12-02
Create Goroutine to Cleanup Orphaned Shadow Secrets

jaireddjawed:
…secrets

### DIFF
--- a/controllers/hcpvaultsecretsapp_controller.go
+++ b/controllers/hcpvaultsecretsapp_controller.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	httptransport "github.com/go-openapi/runtime/client"
@@ -90,6 +91,21 @@ type HCPVaultSecretsAppReconciler struct {
 	BackOffRegistry             *BackOffRegistry
 }
 
+var orphanedShadowSecretManager = NewOrphanedShadowSecretManager()
+
+type OrphanedShadowSecretManager struct {
+	// activeCleanupRoutines is a map of HCPVaultSecretsApp namespaced names, where the corresponing value is true
+	// when a cleanup routing is running to delete orphaned shadow secrets for the HCPVaultSecretsApp if the app was deleted.
+	activeCleanupRoutines map[string]bool
+	mu                    sync.Mutex
+}
+
+func NewOrphanedShadowSecretManager() *OrphanedShadowSecretManager {
+	return &OrphanedShadowSecretManager{
+		activeCleanupRoutines: make(map[string]bool),
+	}
+}
+
 // +kubebuilder:rbac:groups=secrets.hashicorp.com,resources=hcpvaultsecretsapps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=secrets.hashicorp.com,resources=hcpvaultsecretsapps/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=secrets.hashicorp.com,resources=hcpvaultsecretsapps/finalizers,verbs=update
@@ -121,6 +137,45 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
 	if o.GetDeletionTimestamp() != nil {
 		logger.Info("Got deletion timestamp", "obj", o)
 		return ctrl.Result{}, r.handleDeletion(ctx, o)
+	}
+
+	shouldInitiateShadowSecretCleanupRoutine := false
+
+	orphanedShadowSecretManager.mu.Lock()
+	// since each HVPVaultSecretsApp has a unique NamespacedName, we can use the it to keep track
+	// of whether a cleanup routine is running for the HCPVaultSecretsApp already
+	if !orphanedShadowSecretManager.activeCleanupRoutines[req.NamespacedName.String()] {
+		orphanedShadowSecretManager.activeCleanupRoutines[req.NamespacedName.String()] = true
+		shouldInitiateShadowSecretCleanupRoutine = true
+	}
+	orphanedShadowSecretManager.mu.Unlock()
+
+	// start a cleanup routine to periodically check if an HCPVaultSecretsApp has been deleted
+	// so that any orphaned shadow secrets can be deleted
+	if shouldInitiateShadowSecretCleanupRoutine {
+		go func() {
+			// check for cleanup every 5 minutes
+			ticker := time.NewTicker(5 * time.Minute)
+
+			for {
+				select {
+				case <-ticker.C:
+					orphanedShadowSecretsRemoved, err := r.cleanupShadowSecrets(ctx, req)
+					if err != nil {
+						logger.Error(err, "Failed to cleanup shadow secrets", "obj", o)
+					}
+
+					// if the orphaned shadow secrets have been removed, remove the HCPVaultSecretsApp from the active cleanup routines
+					// and stop the ticker so that the cleanup is not run again
+					if orphanedShadowSecretsRemoved {
+						orphanedShadowSecretManager.mu.Lock()
+						delete(orphanedShadowSecretManager.activeCleanupRoutines, req.NamespacedName.String())
+						orphanedShadowSecretManager.mu.Unlock()
+						ticker.Stop()
+					}
+				}
+			}
+		}()
 	}
 
 	var requeueAfter time.Duration
@@ -322,6 +377,26 @@ func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts c
 			builder.WithPredicates(&secretsPredicate{}),
 		).
 		Complete(r)
+}
+
+// returns true if the HCPVaultSecretsApp was deleted and the shadow secret was cleaned up
+func (r *HCPVaultSecretsAppReconciler) cleanupShadowSecrets(ctx context.Context, req ctrl.Request) (bool, error) {
+	o := &secretsv1beta1.HCPVaultSecretsApp{}
+	if err := r.Client.Get(context.Background(), req.NamespacedName, o); err != nil {
+		if apierrors.IsNotFound(err) {
+			// since the HCPVaultSecretsApp does not exist, we return true to indicate that the shadow secret was cleaned up
+			return true, nil
+		}
+
+		return false, err
+	}
+
+	// attempt to delete the shadow secret if the HCPVaultSecretsApp has been deleted
+	if o.GetDeletionTimestamp() != nil {
+		return true, r.handleDeletion(ctx, o)
+	}
+
+	return false, nil
 }
 
 func (r *HCPVaultSecretsAppReconciler) hvsClient(ctx context.Context, o *secretsv1beta1.HCPVaultSecretsApp) (hvsclient.ClientService, error) {


controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -123,6 +139,45 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, r.handleDeletion(ctx, o)
 	}
 
+	shouldInitiateShadowSecretCleanupRoutine := false
+
+	orphanedShadowSecretManager.mu.Lock()
+	// since each HVPVaultSecretsApp has a unique NamespacedName, we can use the it to keep track
+	// of whether a cleanup routine is running for the HCPVaultSecretsApp already
+	if !orphanedShadowSecretManager.activeCleanupRoutines[req.NamespacedName.String()] {
```
jaireddjawed, line 147: I didn't see a way where we were keeping track of HCPVaultSecretsApps in a list or map already. I believe that the map is necessary here because we don't want to create a new goroutine every time `Reconcile()` is executed. We only need one goroutine per HVS app to periodically check if the HVS app was deleted so that we can attempt to remove orphaned shadow secrets.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -123,6 +129,39 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if HCPVaultSecretsAppMap[req.NamespacedName.String()] == nil {
+		HCPVaultSecretsAppMap[req.NamespacedName.String()] = o
+	}
+
+	if !initializedOrphanedShadowSecretCleanup {
+		initializedOrphanedShadowSecretCleanup = true
+
+		go func() {
+			// check for cleanup every 1 hour
+			ticker := time.NewTicker(1 * time.Hour)
```
jaireddjawed, line 142: I wasn't sure what an appropriate interval should be, so I just set it to run once every hour for now.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -123,6 +129,39 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, r.handleDeletion(ctx, o)
 	}
 
+	// keep track of the HCPVaultSecretsApp for the cleanup goroutine
+	if HCPVaultSecretsAppMap[req.NamespacedName.String()] == nil {
```
jaireddjawed, line 133: I create a map here instead of using the existing `HCPVaultSecretsAppList` because it made it easier to keep track of any HVS apps (I would have needed a loop here instead of simply an if statement).

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -123,6 +129,39 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		HCPVaultSecretsAppMap[req.NamespacedName.String()] = o
+	}
+
+	if !initializedOrphanedShadowSecretCleanup {
```
jaireddjawed, line 137: I think that this was the quickest way to prevent a new goroutine coming up every time a CRUD operation is ran on an HVS app. We only need one because the goroutine iterates through every HVS app that had Reconcile() executed on it to check for a possibility of orphaned secrets.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +299,57 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 }
 
+func (r *HCPVaultSecretsAppReconciler) startShadowSecretCleanupRoutine(ctx context.Context) {
+	logger := log.FromContext(ctx)
+	// cleanup of orphaned shadow secrets runs once every hour
+	ticker := time.NewTicker(1 * time.Hour)
+
+	for {
+		select {
+		case <-ticker.C:
```
lursu, line 309: Should maybe add a ctx done case to know when to close this loop out.

jaireddjawed: @lursu I believe that the loop is supposed to run indefinitely to check if any orphaned secrets exist at any point, right?

jaireddjawed: Discussed with @lursu during our sync today and we believe that it would be good to add a comment here to make sure others know that this is a long-running process.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +299,58 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		}
+
+		// delete the HCPVaultSecretsApp if it no longer exists
+		if o.GetDeletionTimestamp() != nil {
+			if err := r.handleDeletion(ctx, o); err != nil {
+				return err
+			}
+
+			logger.Info("Deleted orphaned resources associated with HCPVaultSecretsApp", "app", o.Name)
+		}
```
tvoran, line 348: It seems like this will need to handle the case where the Get() error is just that the object isn't found ([example here](https://github.com/hashicorp/vault-secrets-operator/blob/857159f9578a48c07d25158b0d9f65ba0897efb3/controllers/hcpvaultsecretsapp_controller.go#L118)). It's entirely possible that the HCPVaultSecretsApp is gone from k8s, but the hvs-dynamic-secret-cache secret is leftover, in which case there'd be nothing to call GetDeletionTimestamp() on.

If the HCPVaultSecretsApp is found, we should also compare the UID label ("vso-ownerRefUID") value to see if it matches the UID of the HCPVaultSecretsApp. If they don't match only the secret should be deleted.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +299,58 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		case <-ticker.C:
+			err := r.cleanupOrphanedShadowSecrets(ctx)
+			if err != nil {
+				logger.Error(err, "Failed to cleanup orphaned shadow secrets")
+			}
+		}
+	}
+}
+
+func (r *HCPVaultSecretsAppReconciler) cleanupOrphanedShadowSecrets(ctx context.Context) error {
```
tvoran, line 319: This would be a great function to test with the fake k8s client (look for calls to `testutils.NewFakeClientBuilder()` for examples). It allows you to specify the k8s objects (i.e. HCPVaultSecretsApp's and/or Secret's) that will come back from calls to client.List and client.Get for a given test case.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -123,6 +128,12 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, r.handleDeletion(ctx, o)
 	}
 
+	if !orphanedShadowSecretCleanupInitialized {
+		go r.startShadowSecretCleanupRoutine(ctx)
+		// prevent the shadow secret cleanup goroutine from being started more than once
+		orphanedShadowSecretCleanupInitialized = true
+	}
```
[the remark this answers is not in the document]
jaireddjawed: Got it, I'll move it into the `SetupWithManager()` method.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -75,6 +76,8 @@ var (
 
 	hvsaLabelPrefix = fmt.Sprintf("%s.%s", "hcpvaultsecretsapps",
 		secretsv1beta1.GroupVersion.Group)
+
+	labelOwnerRefUID = fmt.Sprintf("%s/vso-ownerRefUID", secretsv1beta1.GroupVersion.Group)
```
tvoran, line 80: Might be better to make this public in helpers/secrets.go and use it from there so it's not duplicated here.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +295,66 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}, nil
 }
 
+func (r *HCPVaultSecretsAppReconciler) startShadowSecretCleanupRoutine(ctx context.Context) {
+	logger := log.FromContext(ctx)
```
tvoran, line 299: May want to add a name to this logger using WithName:

```suggestion
	logger := log.FromContext(ctx).WithName("startShadowSecretCleanupRoutine")
```

controllers/hcpvaultsecretsapp_controller_test.go
```diff
@@ -1224,3 +1228,130 @@ func Test_makeShadowObjKey(t *testing.T) {
...
+		isHCPVaultSecretsAppDeletionExpected bool
+		isShadowSecretDeletionExpected       bool
+	}{
+		"HCPVaultSecretsApp and shadow secret marked for deletion, HCPVaultSecretsApp is owner of shadow secret": {
+			o: hvsApp.DeepCopy(),
+			secret: &corev1.Secret{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       VaultDynamicSecret.String(),
+					APIVersion: secretsv1beta1.GroupVersion.Version,
+				},
```
tvoran, line 1259: Does TypeMeta need to be set on the Secret? These values don't look right; this isn't a secretsv1beta1 object.

controllers/hcpvaultsecretsapp_controller_test.go
```diff
@@ -1224,3 +1228,130 @@ func Test_makeShadowObjKey(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace:         common.OperatorNamespace,
+					Name:              "shadow-secret",
+					DeletionTimestamp: &deletionTimestamp,
+					Finalizers:        []string{vaultDynamicSecretFinalizer},
+					Labels: map[string]string{
+						"app.kubernetes.io/component": "hvs-dynamic-secret-cache",
```
tvoran, line 1289: Might be good to have a case with a Secret in the operator namespace that doesn't have this label, and verify that it's still around after the cleanup is run.

tvoran (commented):
Just some minor comments, this is looking pretty good.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +295,66 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			if err := r.handleDeletion(ctx, o); err != nil {
+				logger.Error(err, "Failed to handle deletion of HCPVaultSecretsApp", "app", o.Name)
+				return err
+			}
+
+			logger.Info("Deleted orphaned resources associated with HCPVaultSecretsApp", "app", o.Name)
+		} else if apierrors.IsNotFound(err) || secret.GetDeletionTimestamp() != nil {
+			if err := r.Client.Delete(ctx, &secret); err != nil {
+				logger.Error(err, "Failed to delete secret", "secret", secret.Name)
+				return err
```
benashz, line 348: It looks like an error on one deletion will prevent the pruning of other secrets.

I believe we already have some code for pruning the secret cache. I wonder if we could use that instead? 

https://github.com/hashicorp/vault-secrets-operator/blob/336f6ca2973424c1df5fec2b87701c06d8fcc193/vault/cache_storage.go#L144

controllers/hcpvaultsecretsapp_controller_test.go
jaireddjawed: Wasn't sure if it was needed. I'll remove it.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: Edited to re-use the existing `labelOwnerRefUID` variable in helpers/secrets.go.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -306,6 +373,12 @@ func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts c
 		r.BackOffRegistry = NewBackOffRegistry()
 	}
 
+	if !orphanedShadowSecretCleanupInitialized {
```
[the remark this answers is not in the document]
jaireddjawed: @benashz Do you mean that its odd that I am starting the cleanup orphaned secrets process in the SetupWithManager() method, or that it's odd that I'm using a boolean to start the cleanup process? I believe you meant the latter based on the second part of your comment. I made a quick edit and wanted to see if this is what you were looking for. 

```
func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options) error {
	r.referenceCache = newResourceReferenceCache()
	if r.BackOffRegistry == nil {
		r.BackOffRegistry = NewBackOffRegistry()
	}

	mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
		go r.startShadowSecretCleanupRoutine(ctx)
		return nil
	}))
    
    // ...
 ```

File: controllers/hcpvaultsecretsapp_controller.go (hunk 13 above)
Comment (jaireddjawed) [reply controllers/hcpvaultsecretsapp_controller.go:348]: Good catch! I think changing it to `continue` instead of returning would be the best approach because it will continue pruning secrets after the failed deletion of one, but it will eventually come back to the failed deletion during the next cleanup round.

Comment (jaireddjawed) [reply controllers/hcpvaultsecretsapp_controller.go:347]: `ClientCacheStorage` isn't being used at all in `hcpvaultsecretsapp_controller.go`, so I thought it would be easier to do it this way. If I were to add `ClientCacheStorage`, would it look like the following?

* Add `ClientCacheStorage` to `HCPVaultSecretsAppReconciler`
```
type HCPVaultSecretsAppReconciler struct {
	// ...
	ClientCacheStorage          vault.ClientCacheStorage
	StorageConfig               *vault.ClientCacheStorageConfig
}
```

* Initialize `ClientCacheStorage` in `SetupWithManager()`
```
func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options) error {
// ...
storageConfig := vault.DefaultClientCacheStorageConfig()
	var metricsRegistry prometheus.Registerer
	if r.CollectClientCacheMetrics {
		metricsRegistry = ctrlmetrics.Registry
	}
	clientCacheStorage, err := vault.NewDefaultClientCacheStorage(context.Background(), r.Client, storageConfig, metricsRegistry)
	if err != nil {
		return fmt.Errorf("failed to create client cache storage: %w", err)
	}
	r.ClientCacheStorage = clientCacheStorage
}
```

And then simply use `Prune` from `ClientCacheStorage` in the cleanup method that I wrote, right?

Comment (jaireddjawed) [reply controllers/hcpvaultsecretsapp_controller.go:347]: Discussed with @benashz today and we decided that this change would be too involved for a simple ticket.

File: controllers/hcpvaultsecretsapp_controller.go (hunk 2 above)
Comment (jaireddjawed) [reply controllers/hcpvaultsecretsapp_controller.go]: Discussed with @benashz. We will use a default value of once every hour but will let a user pass a command line flag to change the cleanup time.

File: internal/options/env.go [hunk 15]
```diff
@@ -49,6 +49,9 @@ type VSOEnvOptions struct {
 
 	// ClientCacheNumLocks is VSO_CLIENT_CACHE_NUM_LOCKS environment variable option
 	ClientCacheNumLocks *int `split_words:"true"`
+
+	// CleanupOrphanedShadowSecretInterval is VSO_CLEANUP_ORPHANED_SHADOW_SECRETS_INTERVAL environment variable option
+	CleanupOrphanedShadowSecretInterval time.Duration `split_words:"true"`
```
benashz, line 54: Shouldn't this be limited this to HVS only? It seems pretty generic, which is fine.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -80,14 +83,16 @@ var (
...
-	BackOffRegistry             *BackOffRegistry
+	Scheme                              *runtime.Scheme
+	Recorder                            record.EventRecorder
+	SecretDataBuilder                   *helpers.SecretDataBuilder
+	HMACValidator                       helpers.HMACValidator
+	MinRefreshAfter                     time.Duration
+	referenceCache                      ResourceReferenceCache
+	GlobalTransformationOptions         *helpers.GlobalTransformationOptions
+	BackOffRegistry                     *BackOffRegistry
+	CleanupOrphanedShadowSecretInterval time.Duration
```
benashz, line 94: nit: does this need to be member of this structure? Could we maybe pass it in when calling SetupWithManager()?

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+func (r *HCPVaultSecretsAppReconciler) startOrphanedShadowSecretCleanup(ctx context.Context) error {
+	var err error
+
+	r.once.Do(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				if ctx.Err() != nil {
+					err = ctx.Err()
+				}
```
benashz, line 305: It doesn't look like we are breaking/returning here? I think this might lead to a Go routine being leaked.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			// runs the cleanup process once every hour or as specified by the user
+			case <-time.After(r.CleanupOrphanedShadowSecretInterval):
+				err = r.cleanupOrphanedShadowSecrets(ctx)
```
benashz, line 308: Is there any reason assign to err here? I wonder if logging the error might be a better option?

main.go
```diff
@@ -547,14 +555,15 @@ func main() {
...
-		GlobalTransformationOptions: globalTransOptions,
+		Client:                              mgr.GetClient(),
+		Scheme:                              mgr.GetScheme(),
+		Recorder:                            mgr.GetEventRecorderFor("HCPVaultSecretsApp"),
+		SecretDataBuilder:                   secretDataBuilder,
+		HMACValidator:                       hmacValidator,
+		MinRefreshAfter:                     minRefreshAfterHVSA,
+		BackOffRegistry:                     controllers.NewBackOffRegistry(backoffOpts...),
+		GlobalTransformationOptions:         globalTransOptions,
+		CleanupOrphanedShadowSecretInterval: cleanupOrphanedShadowSecretInterval,
```
benashz, line 566: might be good to pass this to SetupWithManager() instead of extending the set of exported members.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	})
+
+	return fmt.Errorf("shadow secret cleanup error err=%s", err)
+}
+
+func (r *HCPVaultSecretsAppReconciler) cleanupOrphanedShadowSecrets(ctx context.Context) error {
+	logger := log.FromContext(ctx).WithName("cleanupOrphanedShadowSecrets")
+
+	// filtering only for dynamic secrets
+	dynamicSecretLabelSelector := client.MatchingLabels{"app.kubernetes.io/component": "hvs-dynamic-secret-cache"}
```
benashz, line 320: nit: only used once, maybe inline it below

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		logger.Error(err, "Failed to list shadow secrets")
+		return err
+	}
+
+	for _, secret := range secrets.Items {
+		o := &secretsv1beta1.HCPVaultSecretsApp{}
+
+		namespace := secret.Labels[hvsaLabelPrefix+"/namespace"]
+		name := secret.Labels[hvsaLabelPrefix+"/name"]
+		namespacedName := types.NamespacedName{Namespace: namespace, Name: name}
```
benashz, line 333: we are not guarding against unset labels or where a label value is empty. I think we probably want to continue if that is the case. You may also be able to use a client.HasLabels matcher in your call to List() above.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+
+		// get the HCPVaultSecretsApp instance that that the shadow secret belongs to (if applicable)
```
benashz, line 335: typo: double "that"

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	dynamicSecretLabelSelector := client.MatchingLabels{"app.kubernetes.io/component": "hvs-dynamic-secret-cache"}
+
+	secrets := corev1.SecretList{}
+	if err := r.List(ctx, &secrets, client.InNamespace(common.OperatorNamespace), dynamicSecretLabelSelector); err != nil {
+		logger.Error(err, "Failed to list shadow secrets")
+		return err
+	}
+
+	for _, secret := range secrets.Items {
+		o := &secretsv1beta1.HCPVaultSecretsApp{}
```
benashz, line 329: nit: probably want to move this variable to after you have the object key.

controllers/hcpvaultsecretsapp_controller.go
benashz, line 333: nit: we usually use the `objKey` variable name here

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		// no errors are returned in the loop because this could block the deletion of other
+		// orphaned shadow secrets that are further along in the list
+		err := r.Get(ctx, namespacedName, o)
+		if err != nil && !apierrors.IsNotFound(err) {
+			logger.Error(err, "Error getting resource from k8s", "secret", secret.Name)
+			continue
+		}
+
+		// if the HCPVaultSecretsApp has been deleted, and the shadow secret belongs to it, delete both
+		if o.GetDeletionTimestamp() != nil && o.GetUID() == types.UID(secret.Labels[helpers.LabelOwnerRefUID]) {
```
benashz, line 345: What if secret labels is missing the UID key?

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,74 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		name := secret.Labels[hvsaLabelPrefix+"/name"]
+		namespacedName := types.NamespacedName{Namespace: namespace, Name: name}
+
+		// get the HCPVaultSecretsApp instance that that the shadow secret belongs to (if applicable)
+		// no errors are returned in the loop because this could block the deletion of other
+		// orphaned shadow secrets that are further along in the list
+		err := r.Get(ctx, namespacedName, o)
+		if err != nil && !apierrors.IsNotFound(err) {
+			logger.Error(err, "Error getting resource from k8s", "secret", secret.Name)
+			continue
```
benashz, line 341: you may want to think about collecting all errors using `errors.Join()`

benashz (commented):
Getting there! I added some more feedback for your consideration.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: Assumed that it always would be included. Will update the if statement to protect from a nil pointer error.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: I agree. There's little to no need to return the error here because it may block the deletion of other secrets in a list if one gets an error in the deletion for some reason.

controllers/hcpvaultsecretsapp_controller.go
benashz: I don't think that would cause a nil pointer panic, I think it would just be like comparing something to an empty string

internal/options/env.go
jaireddjawed: Added HVSA into the environment variable name.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: I changed it to add an extra check to make sure that the LabelOwnerRefUID is not empty.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: I changed it to add an extra check to make sure that the LabelOwnerRefUID is not empty.

jaireddjawed: I checked and the definition of labelOwnerRefID states that it cannot be blank on a secret. I think we are good to leave this as is?

`labelOwnerRefUID is used as the primary key when listing the Secrets owned by a specific VSO object. It should be included in every Secret that is created by VSO.`

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: Included a client.HasLabels to ensure that the secrets iterated in the loop have these labels.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,78 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	var err error
+
+	r.once.Do(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				if ctx.Err() != nil {
+					err = ctx.Err()
+				}
+				break
```
tvoran, line 306: I think this `break` will only break out of the `select`, but still be in the `for` loop? Probably want do something like `return` here or use a label with `break` to target the `for` loop: https://go.dev/ref/spec#Break_statements

controllers/hcpvaultsecretsapp_controller_test.go
```diff
@@ -1224,3 +1227,158 @@ func Test_makeShadowObjKey(t *testing.T) {
...
+						helpers.LabelOwnerRefUID:       "hvsApp4UID",
+					},
+				},
+			},
+			isShadowSecretDeletionExpected: false,
+		},
+	}
+
+	ctx := context.Background()
+	clientBuilder := testutils.NewFakeClientBuilder().Build()
```
tvoran, line 1341: Is this being used for each test case? I wonder if it should be created fresh for each test case so there aren't side effects from leftover k8s objects (i.e. move it inside the t.Run() test function).

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: Good catch! I missed that.

controllers/hcpvaultsecretsapp_controller_test.go
jaireddjawed: I originally had it in the for loop, but Ben told me to move it out so that we can re-use the same ctx for each test case. @benashz please advise. Thanks!

tvoran: Oh I'm referring to just the fake client (`clientBuilder`); using the same ctx is fine.

main.go
```diff
@@ -214,6 +215,10 @@ func main() {
 			"All errors are tried using an exponential backoff strategy. "+
 			"The value must be greater than zero. "+
 			"Also set from environment variable VSO_BACKOFF_MULTIPLIER.")
+	flag.DurationVar(&cleanupOrphanedShadowSecretInterval, "cleanup-orphaned-shadow-secrets-interval", 1*time.Hour,
+		"The time interval between each execution of the cleanup process for cached shadow secrets"+
+			"associated with a deleted HCPVaultSecretsApp. "+
+			"Also set from environment variable VSO_HVSA_CLEANUP_ORPHANED_SHADOW_SECRET_INTERVAL.")
```
benashz, line 221: nit: ideally the env-var's name should match the CLI option.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,78 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			// otherwise, delete the single shadow secret if it has a deletion timestamp
+			if err := helpers.DeleteSecret(ctx, r.Client, objKey); err != nil {
+				errs = errors.Join(errs, fmt.Errorf("failed to delete shadow secret %s: %w", secret.Name, err))
+			}
+
+			logger.Info("Deleted orphaned shadow secret", "secret", secret.Name)
+		}
+	}
+
+	logger.Error(errs, "Failed during cleanup of orphaned shadow secrets")
```
benashz, line 365: It looks like we are logging unconditionally here.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -300,12 +377,17 @@ func (r *HCPVaultSecretsAppReconciler) updateStatus(ctx context.Context, o *secr
...
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options) error {
+func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts controller.Options, cleanupOrphanedShadowSecretInterval time.Duration) error {
 	r.referenceCache = newResourceReferenceCache()
 	if r.BackOffRegistry == nil {
 		r.BackOffRegistry = NewBackOffRegistry()
 	}
 
+	mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
benashz, line 386: Did we test this with more than one VSO replica?

internal/options/env_test.go
```diff
@@ -23,32 +23,34 @@ func TestParse(t *testing.T) {
...
+				"VSO_MAX_CONCURRENT_RECONCILES":                    "10",
+				"VSO_BACKOFF_INITIAL_INTERVAL":                     "1s",
+				"VSO_BACKOFF_MAX_INTERVAL":                         "60s",
+				"VSO_BACKOFF_MAX_ELAPSED_TIME":                     "24h",
+				"VSO_BACKOFF_RANDOMIZATION_FACTOR":                 "0.5",
+				"VSO_BACKOFF_MULTIPLIER":                           "2.5",
+				"VSO_GLOBAL_TRANSFORMATION_OPTIONS":                "gOpt1,gOpt2",
+				"VSO_GLOBAL_VAULT_AUTH_OPTIONS":                    "vOpt1,vOpt2",
+				"VSO_CLIENT_CACHE_NUM_LOCKS":                       "10",
+				"VSO_HVSA_CLEANUP_ORPHANED_SHADOW_SECRET_INTERVAL": "1h",
```
benashz, line 38: nit: name should align with the command line option equivalent

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,78 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		// if the HCPVaultSecretsApp has been deleted, and the shadow secret belongs to it, delete both
+		if o.GetDeletionTimestamp() != nil && o.GetUID() == types.UID(secret.Labels[helpers.LabelOwnerRefUID]) {
+			if err := r.handleDeletion(ctx, o); err != nil {
+				errs = errors.Join(errs, fmt.Errorf("failed to handle deletion of HCPVaultSecretsApp %s: %w", o.Spec.AppName, err))
+			}
+
+			logger.Info("Deleted orphaned resources associated with HCPVaultSecretsApp", "app", o.Name)
+		} else if apierrors.IsNotFound(err) || secret.GetDeletionTimestamp() != nil {
+			// otherwise, delete the single shadow secret if it has a deletion timestamp
+			if err := helpers.DeleteSecret(ctx, r.Client, objKey); err != nil {
```
tvoran, line 357: I think `objKey` is going to be the HCPVaultSecretsApp here; should be the secret instead?

controllers/hcpvaultsecretsapp_controller_test.go
```diff
@@ -1224,3 +1228,159 @@ func Test_makeShadowObjKey(t *testing.T) {
...
+		},
+		"deleted-secret-hvsapp-not-found": {
+			secret: &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: common.OperatorNamespace,
+					Name:      "shadowSecret3",
+					Labels: map[string]string{
+						"app.kubernetes.io/component": "hvs-dynamic-secret-cache",
+					},
+					DeletionTimestamp: &deletionTimestamp,
```
tvoran, line 1305: We should also add a similar test case where DeletionTimestamp is not set.

controllers/hcpvaultsecretsapp_controller_test.go
```diff
@@ -1224,3 +1228,159 @@ func Test_makeShadowObjKey(t *testing.T) {
...
+
+			if tt.isHCPVaultSecretsAppDeletionExpected {
+				deletedHVSApp := &secretsv1beta1.HCPVaultSecretsApp{}
+				err := r.Get(ctx, ctrlclient.ObjectKeyFromObject(tt.o), deletedHVSApp)
+				assert.True(t, apierrors.IsNotFound(err))
+			}
+
+			if tt.isShadowSecretDeletionExpected {
+				deletedSecret := &corev1.Secret{}
+				err := r.Get(ctx, makeShadowObjKey(tt.secret), deletedSecret)
```
tvoran, line 1377: Seems like this should be `ctrlclient.ObjectKeyFromObject(tt.secret)`? Otherwise it's not going to fetch the secret in the test case. And could probably just do this Get() once then do the if/else.

controllers/hcpvaultsecretsapp_controller.go
jaireddjawed: Yes, it ran only on the leader as expected.

controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -288,6 +293,90 @@ func (r *HCPVaultSecretsAppReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+
+func (r *HCPVaultSecretsAppReconciler) cleanupOrphanedShadowSecrets(ctx context.Context) {
+	logger := log.FromContext(ctx).WithName("cleanupOrphanedShadowSecrets")
+	var errs error
+
+	namespaceLabelKey := hvsaLabelPrefix + "/namespace"
+	nameLabelKey := hvsaLabelPrefix + "/name"
+
+	// filtering only for dynamic secrets, also checking if namespace and name labels are present
+	secrets := secretsv1beta1.VaultDynamicSecretList{}
```
tvoran, line 325: I'm not following; why is this controller interacting with VaultDynamicSecret's? Those aren't used with HCPVaultSecretApp's, and the shadow secrets in question here should be k8s Secret's, right?

main.go
```diff
@@ -214,6 +215,10 @@ func main() {
 			"All errors are tried using an exponential backoff strategy. "+
 			"The value must be greater than zero. "+
 			"Also set from environment variable VSO_BACKOFF_MULTIPLIER.")
+	flag.DurationVar(&cleanupOrphanedShadowSecretInterval, "cleanup-orphaned-shadow-secret-interval", 1*time.Hour,
+		"The time interval between each execution of the cleanup process for cached shadow secrets"+
+			"associated with a deleted HCPVaultSecretsApp. "+
+			"Also set from environment variable VSO_HVSA_CLEANUP_ORPHANED_SHADOW_SECRET_INTERVAL.")
```
tvoran, line 221: I wonder if we should add `hvsa` to the end of this to match the other flags? That seems to be the convention the other flags are following (like `min-refresh-after-hvsa`).

It's getting pretty long, so I think we could drop "orphaned" without losing too much meaning:

```suggestion
	flag.DurationVar(&cleanupOrphanedShadowSecretInterval, "cleanup-shadow-secret-interval-hvsa", 1*time.Hour,
		"The time interval between each execution of the cleanup process for cached shadow secrets"+
			"associated with a deleted HCPVaultSecretsApp. "+
			"Also set from environment variable VSO_CLEANUP_SHADOW_SECRET_INTERVAL_HVSA.")
```

internal/options/env.go
```diff
@@ -49,6 +49,9 @@ type VSOEnvOptions struct {
 
 	// ClientCacheNumLocks is VSO_CLIENT_CACHE_NUM_LOCKS environment variable option
 	ClientCacheNumLocks *int `split_words:"true"`
+
+	// CleanupOrphanedShadowSecretInterval is VSO_HVSA_CLEANUP_ORPHANED_SHADOW_SECRETS_INTERVAL environment variable option
+	HVSACleanupOrphanedShadowSecretInterval time.Duration `split_words:"true"`
```
tvoran, line 54: Then a similar thought here to match the command line flag:

```suggestion

	// CleanupShadowSecretIntervalHVSA is VSO_CLEANUP_SHADOW_SECRETS_INTERVAL_HVSA environment variable option
	CleanupShadowSecretIntervalHVSA time.Duration `split_words:"true"`
```

tvoran (commented):
Almost there, this worked locally for me following the reproduction steps in the jira card.

tvoran (commented):
Looks good to me, thanks for addressing the error logging! 👍 

(Since auto-merge is enabled I'll hold off approving until @benashz has had a chance to look at the changes.)

PR #982, opened 2024-12-09
Core: reduce memory footprint

benashz:
Previously, VSO was caching and registering full object watchers on K8s Secrets. While watching K8s Secrets is necessary for automated remediation in the case where a destination secret is deleted from the cluster, doing so can result in OOM conditions for the operator, since each Secret's data contributes to the operator's total memory.

This change does the following:
- disables the caching K8s Secrets in the manager's client.
- only watches for Secret metadata changes.

Remaining tasks:
- [x] Add integration regression tests - see #984

Reproduction steps:
Run:
```
kind create cluster --name oom
helm install --wait --namespace vault-secrets-operator-system --create-namespace vault-secrets-operator hashicorp/vault-secrets-operator
kubectl create namespace oom-secrets
dd if=/dev/urandom of=1m.random bs=1m count=1
# should be enough to trigger the OOM killer
for n in {0..90}; do kubectl create --namespace oom-secrets secret generic sec-$n --from-file /tmp/1m.random ; done
```

Wait for the manager to be OOM killed

Test:
```
make docker-build load-docker-image KIND_CLUSTER_NAME=oom
helm upgrade --set controller.manager.image.tag=0.0.0-dev --namespace vault-secrets-operator-system --create-namespace vault-secrets-operator hashicorp/vault-secrets-operator
```

Wait for VSO to come up, it should *not* be OOM killed

### DIFF
--- a/controllers/hcpvaultsecretsapp_controller.go
+++ b/controllers/hcpvaultsecretsapp_controller.go
@@ -314,6 +314,11 @@ func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts c
 			&secretsv1beta1.SecretTransformation{},
 			NewEnqueueRefRequestsHandlerST(r.referenceCache, nil),
 		).
+		// In order to reduce the operator's memory usage, we only watch for the
+		// Secret's metadata. That is sufficient for us to know when a Secret is
+		// deleted. If we ever need to access to the Secret's data, we can always fetch
+		// it from the API server in a RequestHandler, selectively based on the Secret's
+		// labels.
 		WatchesMetadata(
 			&corev1.Secret{},
 			&enqueueOnDeletionRequestHandler{
--- a/controllers/hcpvaultsecretsapp_controller.go
+++ b/controllers/hcpvaultsecretsapp_controller.go
@@ -314,7 +314,7 @@ func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts c
 			&secretsv1beta1.SecretTransformation{},
 			NewEnqueueRefRequestsHandlerST(r.referenceCache, nil),
 		).
-		Watches(
+		WatchesMetadata(
 			&corev1.Secret{},
 			&enqueueOnDeletionRequestHandler{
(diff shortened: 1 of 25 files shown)


controllers/hcpvaultsecretsapp_controller.go
```diff
@@ -314,7 +314,7 @@ func (r *HCPVaultSecretsAppReconciler) SetupWithManager(mgr ctrl.Manager, opts c
 			&secretsv1beta1.SecretTransformation{},
 			NewEnqueueRefRequestsHandlerST(r.referenceCache, nil),
 		).
-		Watches(
+		WatchesMetadata(
```
tvoran, line 317: We may want to add some comments for the use of `WatchesMetadata` so we don't forget why it's necessary.

tvoran (approved):

tvoran:
The [doc string for WatchesMetadata](https://github.com/kubernetes-sigs/controller-runtime/blob/013f46fbca88cf495a783cdde6dda9ab5c3cd0b9/pkg/builder/controller.go#L177-L203) makes it sound like we should use `metav1.PartialObjectMetadata{}` instead of `corev1.Secret{}` for Get's where possible. I'm wondering if we need to do that for functions like [getSecretExists()](https://github.com/hashicorp/vault-secrets-operator/blob/c17c40e347d7b3c9d9f3d992bea28c434956dbe2/helpers/secrets.go#L376-L387)?

benashz:
> The [doc string for WatchesMetadata](https://github.com/kubernetes-sigs/controller-runtime/blob/013f46fbca88cf495a783cdde6dda9ab5c3cd0b9/pkg/builder/controller.go#L177-L203) makes it sound like we should use `metav1.PartialObjectMetadata{}` instead of `corev1.Secret{}` for Get's where possible. I'm wondering if we need to do that for functions like [getSecretExists()](https://github.com/hashicorp/vault-secrets-operator/blob/c17c40e347d7b3c9d9f3d992bea28c434956dbe2/helpers/secrets.go#L376-L387)?

Yeah, that might make those calls a bit lighter weight. We can probably address that in follow-on work.

PR #996, opened 2025-01-03
CRT: suppress GO-2022-0635

benashz:
Suppress GO-2022-0635 as it is of low severity and VSO isn't using the affected functionalities

Upgrading to latest version of go-secure-stdlib is not possible at this time. The required functionality was inadvertently dropped from github.com/hashicorp/go-secure-stdlib/awsutil during the migration to aws-sdk-go-v2.

Fixes: https://github.com/hashicorp/crt-workflows-common/actions/runs/12589451224/job/35089570078#step:9:1

### DIFF
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -4,13 +4,28 @@
 container {
 	dependencies = true
 	alpine_secdb = true
-	secrets      = true
+	secrets {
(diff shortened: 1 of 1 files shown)


fairclothjm (approved):
LGTM!

PR #1013, opened 2025-02-12
Add Kubernetes Client QPS and Burst Configuration

jaireddjawed:
In clusters with many VaultStaticSecrets, the default Kubernetes client QPS of 5 and Burst of 10 is insufficient, leading it to throttle its communications with the Kubernetes API. This can cause brand-new deployments to roll out very slowly as their pods wait for their Kubernetes secrets to be created from their VaultStaticSecrets.

This PR addresses this by allowing users to change the Kubernetes client QPS and Burst values through environment variables/cli arguments and adds them as chart values.


internal/options/env.go
```diff
@@ -49,6 +49,12 @@ type VSOEnvOptions struct {
 
 	// ClientCacheNumLocks is VSO_CLIENT_CACHE_NUM_LOCKS environment variable option
 	ClientCacheNumLocks *int `split_words:"true"`
+
+	// KubeClientQPS is the VSO_KUBE_CLIENT_QPS environment variable option
+	KubeClientQPS float64 `split_words:"true"`
+
+	// KubeClientBurst is the VSO_KUBE_CLIENT_BURST environment variable option
+	KubeClientBurst *int `split_words:"true"`
```
benashz, line 57: shouldn't this be `*uint`

chart/values.yaml
```diff
@@ -552,10 +566,10 @@ metricsService:
   # https://kubernetes.io/docs/concepts/services-networking/service/
   # @type: map
   ports:
-  - name: https
-    port: 8443
-    protocol: TCP
-    targetPort: https
+    - name: https
```
benashz, line 569: Is this whitespace change necessary?

jaireddjawed: Was likely something I missed when pushing up the new changes.

benashz (approved):

PR #1016, opened 2025-02-20
fix(rollout-restart): adding field manager vault-secrets-operator to the restartAt annotation

### LINKED ISSUE #921: Rollouts trigger a double rollout for Flux managed workloads
If a workload (e.g. Deployment or StatefulSet) is deployed into a cluster via [Flux](https://fluxcd.io/), then setting the `restartedAt` annotation will trigger a second rollout at some point in the future when Flux reconciles the configuration.

This is also a problem if you use kubectl (or any other tooling) to trigger a rollout or otherwise modify the deployment template. 
The [solution](https://fluxcd.io/flux/faq/#why-are-kubectl-edits-rolled-back-by-flux) is to configure the field manager as `flux-client-side-apply` when setting the annotation.

Make the Field Manager configurable when applying the rollout restart annotation.

I'm not sure where this would be configured ideally, I can see 3 options at least
a) As a config flag/env var on the operator pod itself
b) A field in the CRD e.g. `spec.rolloutRestartTargets[{ kind: 'Deployment', name: 'foobar', fieldManager: 'flux-client-side-apply'}]`
c) An annotation on the target workload - `vso.secrets.hashicorp.com/fieldManager: 'flux-client-side-apply'`

I've tested this works by just hardcoding the field manager and deploying the custom operator to a dev environment

BradErz:
Should fix #921


helpers/rollout_restart_test.go
```diff
@@ -164,10 +164,13 @@ func assertPatchedRolloutRestartObj(t *testing.T, ctx context.Context, obj ctrlc
 	switch o := obj.(type) {
 	case *appsv1.Deployment:
 		restartAt = o.Spec.Template.ObjectMeta.Annotations[AnnotationRestartedAt]
+		assertFieldManagerSet(t, obj, FieldManagerName)
 	case *appsv1.StatefulSet:
 		restartAt = o.Spec.Template.ObjectMeta.Annotations[AnnotationRestartedAt]
+		assertFieldManagerSet(t, obj, FieldManagerName)
 	case *appsv1.DaemonSet:
 		restartAt = o.Spec.Template.ObjectMeta.Annotations[AnnotationRestartedAt]
+		assertFieldManagerSet(t, obj, FieldManagerName)
```
benashz, line 173: For completeness, I wonder if we should assert that the field manager was not added in the ArgoRollout case below?

helpers/rollout_restart.go
```diff
@@ -120,37 +123,40 @@ func patchForRolloutRestart(ctx context.Context, obj ctrlclient.Object, client c
...
-		patch := ctrlclient.MergeFrom(t.DeepCopy())
+		patch = ctrlclient.MergeFrom(t.DeepCopy())
 		t.Spec.RestartAt = &metav1.Time{Time: time.Now()}
-		return client.Patch(ctx, t, patch)
+		// we arent setting the field owner here because we probably dont want to own the RestartAt field of the ArgoRollout spec
 	default:
 		return fmt.Errorf("unsupported type %T for rollout-restart patching", t)
 	}
+
+	return client.Patch(ctx, obj, patch, patchOptions...)
```
benashz, line 161: Previously, we were returning from each case. I am not opposed to this refactoring but we should guard against a `nil` `patch` here.

helpers/rollout_restart_test.go
```diff
@@ -185,3 +188,14 @@ func assertPatchedRolloutRestartObj(t *testing.T, ctx context.Context, obj ctrlc
 		"restartAt should be after beforeRolloutRestart",
 		attr, restartAtTime, "beforeRolloutRestart", beforeRolloutRestart)
 }
+
+func assertFieldManagerSet(t *testing.T, obj ctrlclient.Object, manager string) {
+	found := false
```
benashz, line 193:
```suggestion
	t.Helper()
	found := false
```

helpers/rollout_restart_test.go
BradErz: Thanks for the initial pass @benashz. Actually i discovered the test itself is not testing anything regarding the field manager. Because its passed in as an option to the `client.Patch` function. 

Im not super familiar with the kubernetes client; it seems to be a fake implementation of the actual Kubernetes client. Do you have an example of how to inspect the request potentially to make sure the options are really added to the request? I guess in other apps i would use a mock of the client and expect 

As for the test for making sure its not applied to the argo rollout, makes complete sense. Will do it when i figure out how to actually inspect the options ahaha.

BradErz:
After double-checking the installation of VSO in our development environment (0.7.1). It seems it's already adding the FieldManager for the `restartedAt` annotation.

Im not familiar enough with controllers probably to understand what exactly is adding it.. but something is adding the field already.

```yaml
kubectl get deployment -n deseb --show-managed-fields -o json my-test-app | jq '.metadata.managedFields[] | select(.manager == "vault-secrets-operator")'
{
  "apiVersion": "apps/v1",
  "fieldsType": "FieldsV1",
  "fieldsV1": {
    "f:spec": {
      "f:template": {
        "f:metadata": {
          "f:annotations": {
            ".": {},
            "f:vso.secrets.hashicorp.com/restartedAt": {}
          }
        }
      }
    }
  },
  "manager": "vault-secrets-operator",
  "operation": "Update",
  "time": "2025-02-21T11:47:17Z"
}
```

Would you be able to double-check on your side @benashz if you already see the field, too? From my limited experience with the controller-runtime client, I thought the only way to add that was using the the`ctrlclient.FieldOwner(FieldManagerName)` like I did in this PR or using the wrapper client `client = ctrlclient.WithFieldOwner(client, FieldManagerName)`. But maybe it's simply added automatically somewhere in the controller-runtime initialisation.

So potentially, this PR this can just be closed alongside with #921.

koschos:
@benashz Please take a look—I’ve explained here why it can be resolved.
https://github.com/hashicorp/vault-secrets-operator/issues/921#issuecomment-2676857238

Issue #882, opened 2024-08-07
multiple Owner reference on secret with cluster-api and clusterresourceset


sebglon:
**Is your feature request related to a problem? Please describe.**
we deploy some secret generated by vaultstaticsecret with the clusterresourceset (cluster-api).
but VSS fail to sync secret due to 2 reference owner set on the secret

**Describe the solution you'd like**
VSO have to check if the owner is present on the list but not exact match

**Describe alternatives you've considered**
-

**Additional context**
here is an error sync log
```
Failed to update k8s secret: invalid ownerReferences, refs=[]v1.OwnerReference{v1.OwnerReference{APIVersion:"secrets.hashicorp.com/v1beta1", Kind:"VaultStaticSecret", Name:"osc-cloud-contr
oller-manager-secret", UID:"8d29d9be-6076-4a43-bb6a-1e33bfb50ea6", Controller:(*bool)(nil), BlockOwnerDeletion:(*bool)(nil)}, v1.OwnerReference{APIVersion:"addons.cluster.x-k8s.io/v1beta1", Kind:"ClusterResourceSet", Name:"osc-cloud-controller-manager", U
ID:"b3f06448-83dd-4a7e-8fa5-7ff58c09340d", Controller:(*bool)(nil), BlockOwnerDeletion:(*bool)(nil)}}
not the owner of the destination Secret devops-b/osc-cloud-controller-manager-secret
```

sebglon:
Maybe adding some parameters on the owner like:
```
  ownerReferences:
  - apiVersion: secrets.hashicorp.com/v1beta1
    kind: VaultStaticSecret
    name: image-credentials-provider
    uid: b23b06af-7423-40eb-9c37-e993deb88dd0
    blockOwnerDeletion: false
    controller: true
  - apiVersion: addons.cluster.x-k8s.io/v1beta1
    kind: ClusterResourceSet
    name: image-credentials
    uid: 840da7a0-fe56-4da7-9c06-2abb6ef42138
```

sebglon:
Here is the context with cluster-api:
```
---
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultStaticSecret
metadata:
  name: image-credentials-provider
spec:
  vaultAuthRef: cluster-api-${cluster_env}
  mount: secret
  type: kv-v2
  path: cluster-api/${cluster_env}/image-credentials-provider/
  refreshAfter: 5s
  destination:
    name: image-credentials-provider
[...]

PR #1044, opened 2025-04-01
feat(secret): support multiple ownerRefs

sebglon:
Fix: #882

Support multiple OwnerReferences. 
K8S doc [here](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md)

For compatibilty with existing deployment, i can't add the ownerRefences.contoller on the check (with a deep-equal).
But if we have a way to upgrade all existing secrets managed by VSO, we can use the deepEqual.


sebglon:
Hello, @benashz,
Can you help me on this MR? for me it's ok.
This is required to use with Kafka and cluster-api operators.

PR #899, opened 2024-08-29
Extend helm chart to improve controller availability

MM53:
While installing the operator in our setup we did some modifications to the resources to improve the availability of the controller Pods. Since those changes might also help others, it would be great to include them in the Helm chart.

Following changes were applied:
* Add a value to set the `priorityClassName` of the Deployment and Jobs. This supports proper handling of scheduling and eviction in case of low resources
* Allow to configure `topologySpreadConstraints` of the controller Pods. This provides a more detailed control of Pod placement than `affinity` rules.
* Add an optional `PodDisruptionBudget` to avoid evicting all controller Pods at once in case there are multiple replicas.

Please let me know what you think about these changes.

PR #1050, opened 2025-04-15
Add priorityClassName, topologySpreadConstraints and podDisruptionBudget to Helm Chart

jaireddjawed:
Takes the work done in https://github.com/hashicorp/vault-secrets-operator/pull/899 (adds `priorityClassName`, `topologySpreadConstraints` and `PodDisruptionBudget` as fields to the helm chart) and adds tests.


digivava (commented):
This seems like it should have an "improvements" changelog entry. Or do we typically add those at a different point? (newbie-to-the-team question)

chart/values.yaml
```diff
@@ -66,6 +70,34 @@ controller:
...
+
+  # Configure the PodDisruptionBudget for the controller deployment.
+  podDisruptionBudget:
+
+    # toggles the deployment of the PodDisruptionBudget for the controller.
+    # @type: boolean
+    enabled: false
+
+    # Sets the maximum number of pods that can be unavailable during the eviction.
+    # This field cannot be set if minAvailable is set.
```
digivava, line 92: Does Kubernetes ensure this when you try to create a PodDisruptionBudget resource with both `maxUnavailable` and `minAvailable` specified at the same time, or do we need to restrict people from doing that ourselves?

chart/values.yaml
```diff
@@ -66,6 +70,34 @@ controller:
...
+    # @type: integer
+    maxUnavailable: 0
+
+    # Sets the numer of pods that must be available during the eviction.
+    # This field cannot be set if maxUnavailable is set.
+    # @type: integer
```
digivava, line 98: `maxUnavailable` and `minAvailable` can actually both be expressed as [percentages](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#rounding-logic-when-specifying-percentages).

So maybe this would be better as `@type: string`, and then change the conditions on the PDB template yaml to allow for things like "50%" as well?

jaireddjawed:
> This seems like it should have an "improvements" changelog entry. Or do we typically add those at a different point? (newbie-to-the-team question)

I agree that this belongs to "improvements". For VSO, we add the changelog in a separate PR after all PRs that will be in the new version have been approved. You can see an example [here](https://github.com/hashicorp/vault-secrets-operator/pull/1024).

chart/templates/_helpers.tpl
```diff
@@ -345,3 +345,17 @@ clientCache numLocks
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+topologySpreadConstraints appends the "vso.chart.selectorLabels" to .Values.controller.topologySpreadConstraints if no labelSelector was specified
```
digivava, line 350: Since we're also adding topology spread constaints, not just priority class and pod disruption budget, can you modify the PR name so that it's included as well? Just so that we don't forget to mention this field when we make the changelog, as some users may be excited to learn about it. (assuming the PR name is what gets used for the changelog entries)

chart/values.yaml
```diff
@@ -66,6 +70,34 @@ controller:
...
+    # toggles the deployment of the PodDisruptionBudget for the controller.
+    # @type: boolean
+    enabled: false
+
+    # Sets the maximum number of pods that can be unavailable during the eviction.
+    # This field cannot be set if minAvailable is set.
+    # @type: integer
+    maxUnavailable: 0
+
+    # Sets the numer of pods that must be available during the eviction.
```
digivava, line 96: typo here: should be "number"

chart/templates/deployment.yaml
```diff
@@ -161,6 +164,10 @@ spec:
       affinity:
       {{- toYaml .Values.controller.affinity | nindent 8 }}
       {{- end }}
+      {{- if .Values.controller.topologySpreadConstraints }}
+      topologySpreadConstraints:
+      {{- include "vso.topologySpreadConstraints" . | nindent 8 }}
```
digivava, line 169: Can you help me understand why we're using `include` here, but `toYaml` for the "tolerations" array up above? (You may be doing the right thing, I just wanted to double-check since it's been awhile since I did any Go templating stuff.)

test/unit/deployment.bats
```diff
@@ -1350,3 +1350,62 @@ load _helpers
   actual=$(echo "$object" | yq 'contains(["--kube-client-burst=2000"])' | tee /dev/stderr)
   [ "${actual}" = "true" ]
 }
+
+#--------------------------------------------------------------------
+# podDisruptionBudget
```
digivava, line 1355: Are there any tests we could add for the other fields you've added here as well? (priority class and topology spread constraints)

chart/templates/deployment.yaml
jaireddjawed: @digivava `include` takes a yaml template (in this case "vso.topologySpreadConstraints" in _helpers.tpl) and generates yaml using the template.

`toYaml` takes the raw data given and converts it into yaml format. For example, `{{- toYaml .Values.controller.tolerations | nindent 8 }}` takes `tolerations` from the values.yaml file and formats it.

digivava (approved):
Looks good and all tests are passing! Thanks for getting this across the finish line.

PR #1054, opened 2025-04-22
Fix VDS bug where 2 leases are being generated on initial deployment

Zlaticanin:
This PR addresses fixes a bug where two leases were being generated during the initial deployment of VaultDynamicSecret resources. The problem happened because of how the reconciliation logic is handling the scenario where a lease already existed but was being treated as a new lease creation.

With this fix, the operator checks for the existence of a lease and the new behavior is:
1. If the lease exists and is renewable, it will be renewed instead of creating a new lease.
2. If the lease exists but is not renewable, a new lease will be created to ensure valid credentials are available.

This fix prevents the creation of unnecessary leases and ensures that only one active lease exists for ach dynamic secret.

Logs posted in the [ticket](https://hashicorp.atlassian.net/browse/VAULT-35251).

### DIFF
--- a/controllers/vaultdynamicsecret_controller.go
+++ b/controllers/vaultdynamicsecret_controller.go
@@ -396,6 +396,21 @@ func (r *VaultDynamicSecretReconciler) syncSecret(ctx context.Context, c vault.C
 ) (*secretsv1beta1.VaultSecretLease, bool, error) {
 	logger := log.FromContext(ctx).WithName("syncSecret")
 
+	// check if lease already exists
+	if o.Status.SecretLease.ID != "" {
+		logger.V(consts.LogLevelDebug).Info("Lease already exists", "leaseID", o.Status.SecretLease.ID)
+		// if the lease is renewable, renew it
+		if o.Status.SecretLease.Renewable {
+			secretLease, err := r.renewLease(ctx, c, o)
+			if err != nil {
+				logger.Error(err, "Failed to renew lease")
+				return nil, false, err
+			}
+			o.Status.SecretLease = *secretLease
+			return secretLease, false, nil
+		}
+	}
+
 	resp, err := r.doVault(ctx, c, o)
 	if err != nil {
 		return nil, false, err


controllers/vaultdynamicsecret_controller.go
```diff
@@ -150,6 +150,15 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 	o.Status.VaultClientMeta.CacheKey = clientCacheKey.String()
 	o.Status.VaultClientMeta.ID = vClient.ID()
 
+	if !o.Spec.AllowStaticCreds && o.Status.LastGeneration != o.GetGeneration() && o.Status.SecretLease.ID == "" {
```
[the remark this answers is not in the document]
Zlaticanin: Hey @benashz. Thank you for looking at this! 🙏 
I added two comments on the [jira ticket](https://hashicorp.atlassian.net/browse/VAULT-35251) kinda explaining the thought process: 
First comment is where I highlighted the logs, and second comment are pasted logs.
Please let me know what you think and what would be a better approach to solve this.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -123,6 +123,13 @@ func (r *VaultDynamicSecretReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, r.handleDeletion(ctx, o)
 	}
 
+	if addedFinalizer, err := maybeAddFinalizer(ctx, r.Client, o, vaultDynamicSecretFinalizer); err != nil {
+		return ctrl.Result{}, err
+	} else if addedFinalizer {
+		// the finalizer was added, requeue the request.
+		return ctrl.Result{Requeue: true}, nil
+	}
```
tvoran, line 131: There's a comment in maybeAddFinalizer():
https://github.com/hashicorp/vault-secrets-operator/blob/e024c7566022fcddfdd695127666f73663fe801f/controllers/common.go#L248-L249

Any concern with API validation errors now that maybeAddFinalizer() is being called before updateStatus()?

Zlaticanin: Thank you for reviewing! My understanding for the comment is that an API validation error would happen if both metadata (finalizers) and status of a resource are updated in the same reconcile pass. But in our case, we are calling `maybeAddFinalizer()` first, and if a finalizer is added, we exit early and requeue. So we shouldn't be touching the status in the same loop, which would avoid the validation issue because the next reconcile handles the status update separately. Please let me know if I am misunderstanding!

PR #1076, opened 2025-06-04
Fix allowednamespaces is ignored if vaultauth is called default and resides in the vso k8s namespace.

jaireddjawed:


benashz (changes requested):
This change seems invalid to me.

Issue #903, opened 2024-09-02
VSO constantly rotates secret when VaultStaticSecret template text is empty string 


JackCheng01:
**Describe the bug**
VSO constantly rotates secret when VaultStaticSecret template text is empty string , and triggers rollout restart for target resource.

**To Reproduce**

1. create VaultStaticSecret having empty text in templates
e.g.
```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultStaticSecret
metadata:
  annotations:
  name: empty-text
  namespace: default
spec:
  destination:
    create: true
    name: empty-text
    overwrite: true
    transformation:
      excludeRaw: true
      excludes:
      - .*
      templates:
        emptyString:
          text: ""
        emptyStringFromGoTemplte:
          text: '{{- printf "%s" "" -}}'
    type: Opaque
  hmacSecretData: true
  mount: dummy
  namespace: dummy
  path: dummy
  refreshAfter: 30s
  rolloutRestartTargets:
  - kind: Deployment
    name: dummy
  type: kv-v2
```

2. Monitor events for created VaultStaticSecret

```sh
k events --for vaultstaticsecret.secrets.hashicorp.com/empty-text
LAST SEEN   TYPE     REASON                    OBJECT                         MESSAGE
31s         Normal   SecretSynced              VaultStaticSecret/empty-text   Secret synced
6s          Normal   RolloutRestartTriggered   VaultStaticSecret/empty-text   Rollout restart triggered for {Deployment dummy}
6s          Normal   SecretRotated             VaultStaticSecret/empty-text   Secret synced
```

**Expected behavior**

VSO should not rotate secret for VaultStaticSecret when there's no solid change.

**Environment**
* Kubernetes version: 
  * Server Version: v1.29.1-gke.1589020
* vault-secrets-operator version: 0.8.1.0

alitto:
We are seeing a similar behaviour with the following VaultStaticSecret when the referenced secret in vault has no keys (empty value `{}`).

```yaml
apiVersion: secrets.hashicorp.com/v1beta1
kind: VaultStaticSecret
metadata:
  creationTimestamp: '2024-09-20T11:48:27Z'
  finalizers:
    - vaultstaticsecret.secrets.hashicorp.com/finalizer
  generation: 1
spec:
  destination:
    create: true
    name: api.secrets
    overwrite: false
    transformation:
      excludeRaw: true
  hmacSecretData: true
  mount: development
  path: my-secret
[...]

PR #1141, opened 2025-10-25
Marshaling normalization for HMAC comparison

ollaw:
Fixes [#903](https://github.com/hashicorp/vault-secrets-operator/issues/903)

## Problem

When a **Vault secret becomes empty**, a `VaultStaticSecret` with the transformation `excludeRaw: true` and a set of targets set in `rolloutRestartTargets` fields, triggers infinite restarts of target Pods.

The cause of it seems to be the way VSO checks if the secrets has changed and must be updated, an specifically:

1. When a secret on Vault becomes empty, VSO triggers the first _valid_ reconciliation. The destination secret shouldn't have any data anymore (since `excludeRaw: true`) and thus VSO applies the patch with `data: {}` since `{}` it's what is basically coming from Vault.
2. Kubernetes won't save the `data` field on the Secret since is empty.
3. At the next reconciliation, to understand if there are diffs, VSO reads the secret from Vault again and compares the HMAC of the Kubernetes secret with the HMAC of the Vault secret.
4. The HMAC of the two secret is different for the way the emptiness is represented (`nil` for Kubernetes, `{}` for Vault) and the update process is repeated, triggering again the restart of the Pods.

## Changes

This PR normalizes HMAC computation to ensure consistent comparison between Vault and Kubernetes secrets.
If the secret’s data field is nil, it is now initialized to an empty map before computing the HMAC.

## PCI review checklist


benashz:
Closing in favor of #1200 

@ollaw thank you for contributing to HashiCorp!

PR #1146, opened 2025-11-08
Generalize VaultStaticSecret eventing to support future dynamic secrets

jaireddjawed:
## Description

With the introduction of Vault event notifications, VSO can now stream events from Vault in real time, enabling near-immediate updates to static secrets when they are modified or deleted. We plan to extend this capability to dynamic secrets as well.

Because the core event-streaming logic in VSO is largely shared between VaultStaticSecret and VaultDynamicSecret, this PR refactors and generalizes portions of the existing static-secret eventing implementation. These changes lay the groundwork for supporting real-time updates for dynamic secrets without altering current behavior. The PR that introduces instant updates for dynamic secrets is located [here](https://github.com/hashicorp/vault-secrets-operator/pull/1159). Testing has been done on static secrets to ensure behavior remains consistent.

<details>
<summary>Setup KIND cluster with a kv secret and event notifications enabled</summary>

```
#!/usr/bin/env bash
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: BUSL-1.1

make setup-kind
make setup-integration-test

set -e
#!/usr/bin/env bash
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: BUSL-1.1

set -e

cat <<EOF | kubectl -n vault exec -i vault-0 -- sh -e
vault --version
vault secrets disable kvv2/
vault secrets enable -path=kvv2 kv-v2
vault kv put kvv2/secret1 username="db-readonly-username" password="db-secret-password"

cat <<EOT > /tmp/policy.hcl
path "kvv2/data/secret1" {
   capabilities = ["read", "list", "subscribe"]
   subscribe_event_types = ["kv*"]
}
path "sys/events/subscribe/*" {
    capabilities = ["read"]
}
EOT
vault policy write demo /tmp/policy.hcl

# setup the necessary auth backend
vault auth disable kubernetes
vault auth enable kubernetes
vault write auth/kubernetes/config \
    kubernetes_host=https://kubernetes.default.svc
vault write auth/kubernetes/role/demo \
    bound_service_account_names=default \
    bound_service_account_namespaces=tenant-1,tenant-2 \
    policies=demo \
    ttl=1h
EOF

for ns in tenant-{1,2} ; do
    kubectl delete namespace --wait --timeout=30s "${ns}" &> /dev/null || true
    kubectl create namespace "${ns}"
done

make build docker-build deploy-kind

kubectl apply -f config/samples/secrets_v1beta1_vaultconnection.yaml
kubectl apply -f config/samples/secrets_v1beta1_vaultauth.yaml
kubectl apply -f config/samples/secrets_v1beta1_vaultstaticsecret.yaml

```

</details>

<details>
<summary>VaultConnection CRD</summary>

```
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: BUSL-1.1

kind: VaultConnection
metadata:
  labels:
    app.kubernetes.io/name: vaultconnection
    app.kubernetes.io/instance: vaultconnection-sample
    app.kubernetes.io/part-of: vault-secrets-operator
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/created-by: vault-secrets-operator
  name: vaultconnection-sample
  namespace: tenant-1
spec:
  address: http://vault.vault.svc.cluster.local:8200
  skipTLSVerify: true
---
kind: VaultConnection
metadata:
  labels:
    app.kubernetes.io/name: vaultconnection
    app.kubernetes.io/instance: vaultconnection-sample
    app.kubernetes.io/part-of: vault-secrets-operator
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/created-by: vault-secrets-operator
  name: vaultconnection-sample
  namespace: tenant-2
spec:
  address: http://vault.vault.svc.cluster.local:8200
  skipTLSVerify: true
```

</details>

<details>
<summary>VaultAuth CRD</summary>

```
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: BUSL-1.1

kind: VaultAuth
metadata:
  labels:
    app.kubernetes.io/name: vaultauth
    app.kubernetes.io/instance: vaultauth-sample
    app.kubernetes.io/part-of: vault-secrets-operator
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/created-by: vault-secrets-operator
  name: vaultauth-sample
  namespace: tenant-1
spec:
  vaultConnectionRef: vaultconnection-sample
  method: kubernetes
  mount: kubernetes
  kubernetes:
    role: demo
    serviceAccount: default
---
kind: VaultAuth
metadata:
  labels:
    app.kubernetes.io/name: vaultauth
    app.kubernetes.io/instance: vaultauth-sample
    app.kubernetes.io/part-of: vault-secrets-operator
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/created-by: vault-secrets-operator
  name: vaultauth-sample
  namespace: tenant-2
spec:
  vaultConnectionRef: vaultconnection-sample
  method: kubernetes
  mount: kubernetes
  kubernetes:
    role: demo
    serviceAccount: default
```

</details>

<details>
<summary>VaultStaticSecret CRD</summary>

```
---
apiVersion: v1
kind: Secret
metadata:
  name: secret1
  namespace: tenant-1
type: Opaque
---
kind: VaultStaticSecret
metadata:
  namespace: tenant-1
  name: vaultstaticsecret-sample-tenant-1
spec:
  # namespace: cluster1/tenant-1
  vaultAuthRef: vaultauth-sample
  mount: kvv2
  type: kv-v2
  path: secret1
  refreshAfter: 10m
  destination:
    name: secret1
  syncConfig:
    instantUpdates: true
```

</details>

## Manual Testing

<details>
<summary>Get the Secret Stored in k8s To Confirm The Original Value of the Secret</summary>

```
{
    "apiVersion": "v1",
    "data": {
        "_raw": "eyJkYXRhIjp7InBhc3N3b3JkIjoiZGItc2VjcmV0LXBhc3N3b3JkIiwidXNlcm5hbWUiOiJkYi1yZWFkb25seS11c2VybmFtZSJ9LCJtZXRhZGF0YSI6eyJjcmVhdGVkX3RpbWUiOiIyMDI1LTEyLTAzVDA2OjMyOjEzLjM0OTc1NzYzOVoiLCJjdXN0b21fbWV0YWRhdGEiOm51bGwsImRlbGV0aW9uX3RpbWUiOiIiLCJkZXN0cm95ZWQiOmZhbHNlLCJ2ZXJzaW9uIjoxfX0=",
        "password": "ZGItc2VjcmV0LXBhc3N3b3Jk",
        "username": "ZGItcmVhZG9ubHktdXNlcm5hbWU="
    },
    "kind": "Secret",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"secret1\",\"namespace\":\"tenant-1\"},\"type\":\"Opaque\"}\n"
        },
        "creationTimestamp": "2025-12-03T06:33:30Z",
        "name": "secret1",
        "namespace": "tenant-1",
        "resourceVersion": "815",
        "uid": "de782965-1c56-4dfa-ad5d-a4ae396aaa8a"
    },
    "type": "Opaque"
}
```

</details>

<details>
<summary>Perform an update to the password kv store in Vault and ensured that the password changed in k8s</summary>

```
{
    "apiVersion": "v1",
    "data": {
        "_raw": "eyJkYXRhIjp7InBhc3N3b3JkIjoiZGItc2VjcmV0LXBhc3N3b3JkLTEyMyIsInVzZXJuYW1lIjoiZGItcmVhZG9ubHktdXNlcm5hbWUifSwibWV0YWRhdGEiOnsiY3JlYXRlZF90aW1lIjoiMjAyNS0xMi0wM1QwNzowMDowMi45NzA0ODQzODNaIiwiY3VzdG9tX21ldGFkYXRhIjpudWxsLCJkZWxldGlvbl90aW1lIjoiIiwiZGVzdHJveWVkIjpmYWxzZSwidmVyc2lvbiI6Mn19",
        "password": "ZGItc2VjcmV0LXBhc3N3b3JkLTEyMw==",
        "username": "ZGItcmVhZG9ubHktdXNlcm5hbWU="
    },
    "kind": "Secret",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"secret1\",\"namespace\":\"tenant-1\"},\"type\":\"Opaque\"}\n"
        },
        "creationTimestamp": "2025-12-03T06:57:32Z",
        "name": "secret1",
        "namespace": "tenant-1",
        "resourceVersion": "1126",
        "uid": "b5bc1ce0-8299-45c8-9a3e-a9cb65c8398d"
    },
    "type": "Opaque"
}
```

</details>

<details>
<summary>Perform a rollback to confirm that the secret reverts to the original on a near-instant basis</summary>

```
{
    "apiVersion": "v1",
    "data": {
        "_raw": "eyJkYXRhIjp7InBhc3N3b3JkIjoiZGItc2VjcmV0LXBhc3N3b3JkIiwidXNlcm5hbWUiOiJkYi1yZWFkb25seS11c2VybmFtZSJ9LCJtZXRhZGF0YSI6eyJjcmVhdGVkX3RpbWUiOiIyMDI1LTEyLTAzVDA3OjAyOjQ3Ljc3NjcwNjMzNVoiLCJjdXN0b21fbWV0YWRhdGEiOm51bGwsImRlbGV0aW9uX3RpbWUiOiIiLCJkZXN0cm95ZWQiOmZhbHNlLCJ2ZXJzaW9uIjo0fX0=",
        "password": "ZGItc2VjcmV0LXBhc3N3b3Jk",
        "username": "ZGItcmVhZG9ubHktdXNlcm5hbWU="
    },
    "kind": "Secret",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"secret1\",\"namespace\":\"tenant-1\"},\"type\":\"Opaque\"}\n"
        },
        "creationTimestamp": "2025-12-03T06:57:32Z",
        "name": "secret1",
        "namespace": "tenant-1",
        "resourceVersion": "1426",
        "uid": "b5bc1ce0-8299-45c8-9a3e-a9cb65c8398d"
    },
    "type": "Opaque"
}
```

</details>

<details>
<summary>Perform deletion of the username field and confirm that it has been removed from k8s on a near-instant basis</summary>

```
{
    "apiVersion": "v1",
    "data": {
        "_raw": "eyJkYXRhIjp7InBhc3N3b3JkIjoiZGItc2VjcmV0LXBhc3N3b3JkIn0sIm1ldGFkYXRhIjp7ImNyZWF0ZWRfdGltZSI6IjIwMjUtMTItMDNUMDc6MDQ6NTQuNzc3Njc5ODhaIiwiY3VzdG9tX21ldGFkYXRhIjp7Im1ldGFkYXRhLWtleSI6InZhbHVlIn0sImRlbGV0aW9uX3RpbWUiOiIiLCJkZXN0cm95ZWQiOmZhbHNlLCJ2ZXJzaW9uIjo1fX0=",
        "password": "ZGItc2VjcmV0LXBhc3N3b3Jk"
    },
    "kind": "Secret",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"secret1\",\"namespace\":\"tenant-1\"},\"type\":\"Opaque\"}\n"
        },
        "creationTimestamp": "2025-12-03T06:57:32Z",
        "name": "secret1",
        "namespace": "tenant-1",
        "resourceVersion": "1658",
        "uid": "b5bc1ce0-8299-45c8-9a3e-a9cb65c8398d"
    },
    "type": "Opaque"
}
```

</details>

## PCI review checklist

### DIFF
--- a/controllers/instant_updates.go
+++ b/controllers/instant_updates.go
@@ -191,11 +191,8 @@ func (cfg *InstantUpdateConfig) getEvents(ctx context.Context, o client.Object, 
 				retryBackoff.Reset()
 				continue
 			}
-			if ctx.Err() != nil {
-				return
-			}
 			if strings.Contains(err.Error(), "use of closed network connection") ||
-				strings.Contains(err.Error(), "context canceled") {
+				ctx.Err() != nil {
 				// The connection and/or context was closed, so we should
 				// exit the goroutine (and the defer will remove this from
 				// the registry)
@@ -231,15 +228,13 @@ func (cfg *InstantUpdateConfig) getEvents(ctx context.Context, o client.Object, 
 				logger.Error(
 					fmt.Errorf("failed to get event watcher metadata for VaultStaticSecret"),
 					"key", name.String())
+				cfg.enqueueForReconcile(name)
 				return
 			}
 
 			meta.LastClientID = newClient.ID()
 			cfg.Registry.Register(name, meta)
 
-			shouldBackoff = false
-			errorCount = 0
-			retryBackoff.Reset()
 		}
 	}
 }


api/v1beta1/vaultstaticsecret_types.go
```diff
@@ -72,7 +72,7 @@ type VaultStaticSecretCollectable struct {
 // SyncConfig configures sync behavior from Vault to VSO
 type SyncConfig struct {
 	// InstantUpdates is a flag to indicate that event-driven updates are
-	// enabled for this VaultStaticSecret
+	// enabled for VaultStaticSecret/VaultDynamicSecret
```
tvoran:
```suggestion
	// enabled for this VaultStaticSecret/VaultDynamicSecret
```

api/v1beta1/vaultdynamicsecret_types.go
```diff
@@ -61,6 +61,8 @@ type VaultDynamicSecretSpec struct {
 	RolloutRestartTargets []RolloutRestartTarget `json:"rolloutRestartTargets,omitempty"`
 	// Destination provides configuration necessary for syncing the Vault secret to Kubernetes.
 	Destination Destination `json:"destination"`
+	// SyncConfig configures sync behavior from Vault to VSO
+	SyncConfig *SyncConfig `json:"syncConfig,omitempty"`
```
tvoran: In the future we may want to have a separate DynamicSyncConfig type if we need options that aren't applicable to VaultStaticSecret's, but this is probably fine for now.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -631,6 +647,242 @@ func (r *VaultDynamicSecretReconciler) awaitVaultSecretRotation(ctx context.Cont
 	return staticCredsMeta, resp, nil
 }
 
+func (r *VaultDynamicSecretReconciler) ensureEventWatcher(ctx context.Context, o *secretsv1beta1.VaultDynamicSecret, c vault.Client) error {
+	logger := log.FromContext(ctx).WithName("ensureEventWatcher")
+	name := client.ObjectKeyFromObject(o)
+
+	meta, ok := r.eventWatcherRegistry.Get(name)
+	if ok {
+		// The watcher is running, and if the VSS object has not been updated,
```
tvoran:
```suggestion
		// The watcher is running, and if the VDS object has not been updated,
```

controllers/vaultdynamicsecret_controller.go
```diff
@@ -631,6 +647,242 @@ func (r *VaultDynamicSecretReconciler) awaitVaultSecretRotation(ctx context.Cont
...
+	meta, ok := r.eventWatcherRegistry.Get(name)
+	if ok {
+		if meta.Cancel != nil {
+			meta.Cancel()
+		}
+		r.eventWatcherRegistry.Delete(name)
+	}
+}
+
+// getEvents calls streamStaticSecretEvents in a loop, collecting and responding
```
tvoran: This is calling `streamDynamicSecretEvents` now.

controllers/vaultdynamicsecret_controller.go
```diff
@@ -631,6 +647,242 @@ func (r *VaultDynamicSecretReconciler) awaitVaultSecretRotation(ctx context.Cont
...
+			if modified {
+				namespace := strings.Trim(messageMap.Data.Namespace, "/")
+				path := messageMap.Data.Event.Metadata.Path
+
+				// Filter out KV v1/v2 events since these should be handled by VaultStaticSecretController
+				if isKVSecretPath(path) {
+					logger.V(consts.LogLevelDebug).Info("KV secret event received, ignoring (should be handled by VaultStaticSecretController)",
+						"namespace", namespace, "path", path)
+					continue
+				}
```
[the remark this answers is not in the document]
jaireddjawed: Thanks for the suggestion. I used it in the following [PR](https://github.com/hashicorp/vault-secrets-operator/pull/1159) that will be merged after this one, where I acutally update/remove secrets based on the events that are received. I ensured that we only update/remove a secret based on it's own events and not another secret's events.

api/v1beta1/vaultstaticsecret_types.go
```diff
@@ -41,7 +41,7 @@ type VaultStaticSecretSpec struct {
 	// Destination provides configuration necessary for syncing the Vault secret to Kubernetes.
 	Destination Destination `json:"destination"`
 	// SyncConfig configures sync behavior from Vault to VSO
-	SyncConfig *SyncConfig `json:"syncConfig,omitempty"`
+	SyncConfig *VaultStaticSecretSyncConfig `json:"syncConfig,omitempty"`
```
tvoran: I'm a little worried about changing the name of this type, since this is effectively part of the API, and changing the name of the type could break someone who is using this programmatically.

So I'm rethinking my previous suggestion here, perhaps it's better to use `SyncConfig` for all secret CRDs, and if we need to add other options for one of the secret types in the future, we can add a config block inside of `SyncConfig` just for that type (i.e. `DynamicOptions`).

controllers/vaultstaticsecret_controller.go
```diff
@@ -286,13 +279,23 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		// starts listening to events that are relevant to the secret so that updates can be triggered
+		// on a near-instant basis
+		err := EnsureEventWatcher(ctx, &InstantUpdateConfig{
+			Secret:          o,
+			Client:          c,
+			Registry:        r.eventWatcherRegistry,
+			BackOffRegistry: r.BackOffRegistry,
+			SourceCh:        r.SourceCh,
+			Recorder:        r.Recorder,
+			NewClientFunc:   r.newVSSClient,
```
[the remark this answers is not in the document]
jaireddjawed: I believe that it is using `r.ClientFactory` already. I placed the function away from the `EnsureEventWatcher` function because it was getting a bit crowded. Let me know if you'd like me to change it.

```
func (r *VaultStaticSecretReconciler) newVSSClient(ctx context.Context, obj client.Object) (vault.Client, error) {
	vss, ok := obj.(*secretsv1beta1.VaultStaticSecret)
	if !ok {
		return nil, fmt.Errorf("unexpected object type %T", obj)
	}
	return r.ClientFactory.Get(ctx, r.Client, vss)
}
```

controllers/instant_updates.go
```diff
@@ -0,0 +1,388 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
+	"github.com/hashicorp/vault-secrets-operator/consts"
+	"github.com/hashicorp/vault-secrets-operator/vault"
+)
+
+const (
+	// InstantUpdateEventPath is the path to subscribe to Vault events
+	InstantUpdateEventPath = "/v1/sys/events/subscribe/*"
```
tvoran, line 31: Just thinking about the implications of this for users, since this change will require users to change their Vault policies (like you had to in the test, changing the allowed subscription path from `sys/events/subscribe/kv*` to `sys/events/subscribe/*`. It looks like our documentation examples and tutorial all use the `kv*` endpoint, so those will all need to be updated, and we should probably add a warning in the VSO changelog, since this isn't backwards compatible with our previous instructions.

Alternatively, I wonder how hard it would be to subscribe to `kv*` when the object is a VaultStaticSecret? Although if we were to change the subscription connections to be one per VaultAuth instead of one per VaultStaticSecret or VaultDynamicSecret it would probably need to subscribe to `*`.

Anyway, just wanted to call it out here.

controllers/instant_updates.go
```diff
@@ -0,0 +1,388 @@
...
+	// create the websocket that will be used to receive instant update events
+	wsClient, err := cfg.Client.WebsocketClient(InstantUpdateEventPath)
+	if err != nil {
+		return fmt.Errorf("failed to create websocket client: %w", err)
+	}
+
+	conn, err := wsClient.Connect(ctx)
+	if err != nil {
+		return err
+	}
```
tvoran: I'm curious why the connection is being made here, instead of in streamStaticSecretEvents() as it was previously?

If the connection fails here, it looks like it'll never be retried, just emits a k8s event and continues: https://github.com/hashicorp/vault-secrets-operator/blob/b22eb5dfe40236304fdf7e4229e34d0ddf40b203/controllers/vaultstaticsecret_controller.go#L294-L296

The previous way looks like it would retry if there was a connection error, so that's why I'm curious.

controllers/instant_updates.go
```diff
@@ -0,0 +1,388 @@
...
+					logger.Error(err, "Too many errors while watching events, requeuing")
+					cfg.enqueueForReconcile(name)
+					return
+				}
+			}
+		}
+	}
+}
+
+func (cfg *InstantUpdateConfig) streamSecretEvents(ctx context.Context, obj client.Object, _ websocket.MessageType, data []byte) (bool, error) {
```
tvoran: Does this function still need the MessageType parameter?

jaireddjawed: Nope. Just removed it.

controllers/instant_updates.go
jaireddjawed: I originally did it there when I was trying to get the connection to work after my changes and at the time I was running into issues during my testing. I moved it back to ensure that it retries.

controllers/instant_updates.go
jaireddjawed: I'm thinking that there are a couple of ways we can remediate this?

1. We can have the user provide the permission path(s) while they are configuring instant updates. This could keep VSO from over-subscribing to events that are not necessary to look at. This would be a breaking change unless we default to an event path if the user doesn't provide one. Documentation would also still need to be updated.

2. Since `EnsureEventWatcher` is called from the secret's `Reconcile()` we should be able to pass the secret's object to `EnsureEventWatcher` and determine the secret type.

* If the websocket is being initialized for the first time and it was a `VaultStaticSecret`, we can subscribe to `kv*` only. If it's a dynamic secret we subscribe to `*`.

* If we already subscribed to `kv*` earlier and `EnsureEventWatcher()` is being called again by a dynamic secret, we can re-establish the connection with the `*` path. 

This way is a bit less clean code-wise and still requires that we over-subscribe to events, but ensures backward comparability.

I'm thinking route #2 is the best path. Please let me know your suggestion!

controllers/instant_updates_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+
+	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
+	"github.com/hashicorp/vault-secrets-operator/consts"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestStreamSecretEvents(t *testing.T) {
```
tvoran: This should probably be renamed now.


```suggestion
func Test_matchSecretEvent(t *testing.T) {
```

controllers/instant_updates_test.go
```diff
@@ -0,0 +1,161 @@
...
+			eventJSON: []byte(fmt.Sprintf(
+				`{"data":{"event":{"metadata":{"modified":"true"}},"namespace":"/%s"}}`,
+				vssKVV2.Spec.Namespace,
+			)),
+			wantMatch: false,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
tvoran: This doesn't seem necessary?


```suggestion
```

controllers/vaultstaticsecret_controller.go
```diff
@@ -341,254 +344,12 @@ func (r *VaultStaticSecretReconciler) handleDeletion(ctx context.Context, o clie
...
-				if nextBackoff == backoff.Stop {
-					logger.Error(fmt.Errorf("backoff limit reached"), "Backoff limit reached, requeuing")
-					break eventLoop
-				}
-				time.Sleep(retryBackoff.NextBackOff())
-			}
-			err := r.streamStaticSecretEvents(ctx, &o, wsClient)
-			if err != nil {
-				if strings.Contains(err.Error(), "use of closed network connection") ||
-					strings.Contains(err.Error(), "context canceled") {
```
tvoran, line 446: It looks like this error handling was removed in the new code. I'm curious as to why?

controllers/instant_updates.go
```diff
@@ -0,0 +1,376 @@
...
+	message := vaultEventMessage{}
+	if err := json.Unmarshal(data, &message); err != nil {
+		return false, fmt.Errorf("failed to unmarshal event message: %w", err)
+	}
+
+	if message.Data.Event.Metadata.Modified == "" {
+		return false, nil
+	}
+
+	modified, err := strconv.ParseBool(message.Data.Event.Metadata.Modified)
```
tvoran: Why was this changed from using our `parseutil` library? I think in the past we've tried to prefer our parseutil library over the std lib vesion, and in this case `parseutil.ParseBool()` doesn't required the extra check for empty `modified` on line 272.

controllers/instant_updates.go
```diff
@@ -0,0 +1,376 @@
...
+			if reconnectErr != nil {
+				logger.Error(reconnectErr, "Failed to reconnect websocket watcher", "object", name)
+				continue
+			}
+
+			wsClient = newWSClient
+			cfg.Client = newClient
+			if meta, ok := cfg.Registry.Get(name); ok {
+				meta.LastClientID = newClient.ID()
+			}
```
tvoran: Should this be set back in the registry? It looks like that's what the previous code did:

https://github.com/hashicorp/vault-secrets-operator/pull/1146/changes#diff-36fded271f811978c85d6ac71b49b180c8a6160b7d462c2a0c2e305ef92d420fL482-L492

controllers/instant_updates.go
tvoran: Yeah of those I'd probably go with number 2, though we should run it by product before going down that route (it may be acceptable to just have folks change their Vault policies).

But I don't think we should let that hold this PR up, since this is going into a feature branch.

tvoran (commented):
Just some minor questions.

controllers/instant_updates.go
jaireddjawed: @stevealmyHC Please let me know if you have a preference regarding this design choice. As said above, @tvoran and I agree that option 2 is the best of these choices.

controllers/vaultstaticsecret_controller.go
jaireddjawed: That was a miss by me. Fixed it.

controllers/instant_updates.go
jaireddjawed: I discussed with @stevealmyHC and he was saying that there could be a concern with using the `*` path where a user can have 10k+ secrets and only need instant updates for a small fraction of those, leading to the websocket responding to a lot of events that are not necessary.

I've created a ticket where we can decide what the best implementation route is. I came up with some additional approaches.

1. Currently, we add a websocket for each individual secret. We can continue this approach with dynamic secrets and only subscribe to the required events. The trade off for this is that we have the additional overhead of the extra websockets, but the overhead is manageable, especially since there are already customers using it in production systems.

2. We can evaluate the path of each secret that calls the `EnsureEventWatcher` function. From there, we can subscribe to the least global path necessary to get the events that occur on each secret.

For example, if a user has two secrets `postgres/my-db-secret` and `postgres/my-db-secret-2`, we'd determine that the least global path required to get all the events that occur is `postgres/*`

controllers/instant_updates.go
```diff
@@ -0,0 +1,391 @@
...
+					"key", name.String())
+				return
+			}
+
+			meta.LastClientID = newClient.ID()
+			cfg.Registry.Register(name, meta)
+
+			shouldBackoff = false
+			errorCount = 0
+			retryBackoff.Reset()
```
tvoran: At this point streamSecretEvents() has returned an error of some sort, and the error count is under the InstantUpdateErrorThreshold. Should the backoff be reset and the errorCount set to 0? Seems like the backoff will never actually be used?


```suggestion
```

controllers/instant_updates.go
```diff
@@ -0,0 +1,391 @@
...
+				shouldBackoff = false
+				errorCount = 0
+				retryBackoff.Reset()
+				continue
+			}
+			if ctx.Err() != nil {
+				return
+			}
+			if strings.Contains(err.Error(), "use of closed network connection") ||
+				strings.Contains(err.Error(), "context canceled") {
```
tvoran: Are `ctx.Err() != nil` and `strings.Contains(err.Error(), "context canceled")` redundant? Checking ctx.Err() seems like a stronger check, so would it work to combine and use that here?

```suggestion
			if strings.Contains(err.Error(), "use of closed network connection") ||
				ctx.Err() != nil {
```

controllers/instant_updates.go
```diff
@@ -0,0 +1,385 @@
...
+
+			wsClient = newWSClient
+			cfg.Client = newClient
+
+			meta, ok := cfg.Registry.Get(name)
+			if !ok {
+				logger.Error(
+					fmt.Errorf("failed to get event watcher metadata for VaultStaticSecret"),
+					"key", name.String())
+				return
```
tvoran, line 232: Should this be requeued for reconcile on this error? Otherwise it looks like it just gives up at this point.


```suggestion
				cfg.enqueueForReconcile(name)
				return
```

jaireddjawed: Yes, that would have been the case. Changed it!

controllers/instant_updates.go
```diff
@@ -0,0 +1,386 @@
...
+	// Use the same backoff options used for Vault reads in Reconcile()
+	retryBackoff := backoff.NewExponentialBackOff(cfg.BackOffRegistry.opts...)
+
+	shouldBackoff := false
+	errorCount := 0
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.V(consts.LogLevelDebug).Info("Context done, stopping watcher")
```
tvoran: Does this logger have any information set in it for which watcher is being stopped? Otherwise I'm not sure how useful this will be (compared to the previous log message: https://github.com/hashicorp/vault-secrets-operator/pull/1146/changes#diff-36fded271f811978c85d6ac71b49b180c8a6160b7d462c2a0c2e305ef92d420fL431-L432)

controllers/instant_updates.go
```diff
@@ -0,0 +1,386 @@
...
+		return fmt.Errorf("instant update watcher requires a client factory")
+	}
+	if cfg.EventObjectFactory == nil {
+		cfg.EventObjectFactory = defaultEventObjectFactory(cfg.Secret)
+	}
+	return nil
+}
+
+// defaultEventObjectFactory creates a default event object factory for tests
+func defaultEventObjectFactory(template client.Object) func(types.NamespacedName) client.Object {
```
tvoran, line 374: It looks like this is the only EventObjectFactory implementation? But the comment says this is for tests?

controllers/instant_updates.go
```diff
@@ -0,0 +1,386 @@
...
+	return func(key types.NamespacedName) client.Object {
+		objCopy := template.DeepCopyObject()
+		obj, ok := objCopy.(client.Object)
+		if !ok {
+			return template
+		}
+		obj.SetNamespace(key.Namespace)
+		obj.SetName(key.Name)
+		obj.SetResourceVersion("")
```
tvoran: Curious why these need to be set? Does the deep copy not copy them?

PR #1156, opened 2025-11-20
Add Transit Options to Encrypt/Decrypt Helpers

blaumer:
### Summary
This PR adds optional `TransitOption` parameters to EncryptWithTransit and DecryptWithTransit, enabling callers to pass additional Transit API fields (e.g., key version). Includes a helper WithKeyVersion(int) and remains fully backward-compatible.

Right now there isnt any support for decrypt options. New options can be added as needed

### Testing
Was able to successfully make Encrypt calls when specifying key version

### DIFF
--- a/vault/transit_test.go
+++ b/vault/transit_test.go
@@ -12,9 +12,18 @@ func TestWithKeyVersion(t *testing.T) {
 		name string
 		v    uint
 	}{
-		{"key version 1", 1},
-		{"key version 42", 42},
-		{"key version 10000", 10000},
+		{
+			name: "key version 1",
+			v:    1,
+		},
+		{
+			name: "key version 42",
+			v:    42,
+		},
+		{
+			name: "key version 10000",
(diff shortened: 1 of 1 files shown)


vault/transit.go
```diff
@@ -76,3 +84,7 @@ func DecryptWithTransit(ctx context.Context, vaultClient Client, mount, key stri
 
 	return base64.StdEncoding.DecodeString(d.Plaintext)
 }
+
+func WithKeyVersion(v int) TransitOption {
```
digivava: If we're going to have just the singular TransitOption (instead of TransitEncryptOptions, etc. for each operation), can we add a godoc comment to this explaining it's for Encrypt, so that future developers won't accidentally think they can pass WithKeyVersion to the Decrypt call, etc.?

digivava (approved):

benashz (dismissed):
Minor nit, otherwise +1

PR #1189, opened 2025-12-31
Standalone caching client factory

digivava:
There are use cases where our other non-Operator consumers of the VSO client library want to use the VaultAuth and VaultConnection structs but do not have access to actual VaultAuth and VaultConnection resources. Currently, many of the existing caching client factory-related methods have dependencies on the actual resources existing in cluster.

This PR changes that, by allowing developers to designate the caching client factory (and thus, the client) as "standalone", meaning that it does not require VaultAuth or VaultConnection custom resources in the cluster.

## PCI review checklist

### DIFF
--- a/vault/cache_key.go
+++ b/vault/cache_key.go
@@ -139,16 +139,10 @@ func ComputeClientCacheKeyFromMeta(ctx context.Context, client ctrlclient.Client
 
 // ComputeClientCacheKey for use in a ClientCache. It is derived by combining instances of
 // VaultAuth, VaultConnection, and a CredentialProvider UID.
-//
-// When isStandalone is false (normal VSO operation with K8s resources):
-// - Uses K8s resource UIDs and generations from authObj, connObj, and providerUID
-// - Validates that all UIDs are exactly 36 characters
-// - Checks for duplicate UIDs
-//
-// When isStandalone is true (standalone mode without K8s resources):
-// - Uses content-based hashes of authObj.Spec and connObj.Spec instead of UIDs
-// - Allows empty UIDs from objects not fetched from K8s API
-// - Generation is always 1 since objects aren't K8s resources
+// All of these elements are summed together into a SHA256 checksum,
+// and prefixed with the VaultAuth method. The chances of a collision are extremely remote,
+// since the inputs into the hash should always be unique. For example, we use the UUID
+// from three different sources as inputs.
 //
 // The resulting key will resemble something like: kubernetes-2a8108711ae49ac0faa724, where the prefix
 // is the VaultAuth.Spec.Method, and the remainder is the concatenation of the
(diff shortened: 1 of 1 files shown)


vault/cache_key.go
```diff
@@ -153,38 +153,63 @@ func ComputeClientCacheKeyFromMeta(ctx context.Context, client ctrlclient.Client
...
-		if _, ok := seen[uid]; ok {
-			errs = errors.Join(errs, fmt.Errorf("%w %s", errorDuplicateUID, uid))
+
+		if errs != nil {
+			return "", errs
+		}
+
+		authSpecHash := helpers.HashString(fmt.Sprintf("%+v", authObj.Spec))
+		connSpecHash := helpers.HashString(fmt.Sprintf("%+v", connObj.Spec))
+
```
benashz: Would it make sense to validate the various Spec? Could the "repr" be empty or missing required values?

vault/cache_key.go
```diff
@@ -153,38 +153,63 @@ func ComputeClientCacheKeyFromMeta(ctx context.Context, client ctrlclient.Client
...
+		// Format: "authHash-1.connHash-1.providerUID"
+		// (generation is always 1 for standalone since we didn't fetch the auth and conn objects from a K8s resource)
+		input = fmt.Sprintf("%s-%d.%s-%d.%s",
+			authSpecHash, 1,
```
benashz: nit: is there a particular reason to interpolate the generation since it is always 1?

benashz (commented):
In general, I like the extension  of the client factory support the non-api approach. Although I wonder if we should implement this feature as a `standAloneClientFactory` rather than extending the ClientFactory interface? For some reason that feels a bit clearer to me.

vault/cache_key.go
```diff
@@ -153,38 +153,63 @@ func ComputeClientCacheKeyFromMeta(ctx context.Context, client ctrlclient.Client
...
 	}
 
-	// only used for duplicate UID detection, all values are ignored
-	seen := make(map[types.UID]int)
-	requireUIDLen := 36
-	validateUID := func(name string, uid types.UID) {
-		if len(uid) != requireUIDLen {
-			errs = errors.Join(errs, fmt.Errorf("%w %d, must be %d", errorInvalidUIDLength, len(uid), requireUIDLen))
+	var input string
+	if isStandalone {
```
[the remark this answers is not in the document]
digivava: I thought about it, but got a bit scared off by the "read-only" description of the UID field (since traditionally the field is set by the API server).

I like your idea about making this whole thing a `standaloneClientFactory` instead though; that may clean up all of this and also allow us to more definitively avoid breaking existing code.

digivava:
Closing in favor of https://github.com/hashicorp/vault-secrets-operator/pull/1199

PR #1194, opened 2026-01-08
fix: invalid secret HMAC validation on static creds

benashz:
Fixes an issue with secret data drift detection and when reconciling static creds. 

The HMAC included fields from the secret data were not being filtered and could include non-idempotent data fields.

## PCI review checklist


digivava (approved):
Makes sense to me!

PR #1195, opened 2026-01-08
Allow for custom cache key function in client factory

digivava:
This PR makes it so other tools can use VSO's client factory code as a library even when they don't necessarily need or have access to everything that VSO's cache key requires.

## PCI review checklist


digivava:
trying a different approach off of the approle-extension branch, so will open a new PR

PR #1255, opened 2026-04-24
fix: Update VDS static credentials metadata handling.

Balaji2198:
Fixes a static-creds reconcile bug where callback-driven ForceSyncs could keep using stale `ttl` when the synced Secret payload was unchanged, causing the next horizon to be computed from an old TTL.
The fix refreshes static-creds status metadata before the HMAC early return, and adds regression coverage for both the metadata-only syncSecret path and the higher-level ForceSync Reconcile path.

## PCI review checklist


controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -570,6 +701,204 @@ func TestVaultDynamicSecretReconciler_syncSecret(t *testing.T) {
 	}
 }
 
+func TestVaultDynamicSecretReconciler_syncSecret_staticCredsRefreshesStatusOnMatchingHMAC(t *testing.T) {
```
siyer-corp: Can we document what cases are being covered here?

siyer-corp, line 680: Can we document what cases are being covered here?

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -570,6 +701,204 @@ func TestVaultDynamicSecretReconciler_syncSecret(t *testing.T) {
...
+		LeaseDuration: 0,
+		Renewable:     false,
+	}, lease)
+	assert.Equal(t, int64(540), obj.Status.StaticCredsMetaData.TTL)
+	assert.Equal(t, int64(600), obj.Status.StaticCredsMetaData.RotationPeriod)
+	assert.Equal(t, time.Date(2026, 4, 21, 12, 0, 0, 0, time.UTC).Unix(), obj.Status.StaticCredsMetaData.LastVaultRotation)
+	assert.Equal(t, 1, len(vClient.Requests))
+}
+
+func TestVaultDynamicSecretReconciler_Reconcile_forceSyncStaticCredsUsesRefreshedTTL(t *testing.T) {
```
siyer-corp, line 752: Can we document what cases are being covered here?

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -33,6 +38,132 @@ import (
...
+type reconcileTestVaultClient struct {
+	*vault.MockRecordingVaultClient
+	cacheKey vault.ClientCacheKey
+}
+
+func (c *reconcileTestVaultClient) Init(context.Context, client.Client, *secretsv1beta1.VaultAuth, *secretsv1beta1.VaultConnection, string, *vault.ClientOptions) error {
+	return nil
+}
+
+func (c *reconcileTestVaultClient) Login(context.Context, client.Client) error {
```
siyer-corp: May I know why we need these functions that aren't called in the tests?

Balaji2198: These methods are stubbed so that `reconcileTestVaultClient` implements the full `vault.Client` interface, which is what the `ClientFactory` returns in tests. Without these stubs, the fake client couldn’t be used where `Reconcile` expects a `vault.Client`.

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -570,6 +635,252 @@ func TestVaultDynamicSecretReconciler_syncSecret(t *testing.T) {
...
+//   - LastVaultRotation is identical to the stored status (no rotation happened)
+//   - awaitVaultSecretRotation enters the backoff loop (inLastSyncRotation=true)
+//     and makes a second Vault request, which returns the same data
+//   - the stored SecretMAC matches the newly computed HMAC
+//   - syncSecret therefore does not rewrite the Secret (`updated == false`)
+//   - syncSecret still refreshes status metadata: TTL is updated from the
+//     fresh Vault response even though LastVaultRotation is unchanged
+//
+// Previous status: TTL=600, LastVaultRotation=12:00
+// Refreshed Vault metadata: TTL=540, LastVaultRotation=12:00 (unchanged)
```
siyer-corp: nit: can we remove _unchanged_ here, since the TTL is actually changed here?

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -570,6 +635,252 @@ func TestVaultDynamicSecretReconciler_syncSecret(t *testing.T) {
...
+//   - awaitVaultSecretRotation enters the backoff loop (inLastSyncRotation=true)
+//     and makes a second Vault request, confirming the same stable state
+//   - the destination Secret already contains the same data, so the secret
+//     payload is not rewritten
+//   - Reconcile persists the refreshed TTL from the latest Vault response
+//   - computePostSyncHorizon uses the refreshed TTL for RequeueAfter
+//   - the SyncRegistry entry is removed after a successful reconcile
+//
+// Previous status: TTL=600, LastVaultRotation=12:00
+// Refreshed Vault metadata: TTL=540, LastVaultRotation=12:00 (unchanged)
```
siyer-corp: nit: can we remove _unchanged_ here, since the TTL is actually changed here?

controllers/vaultdynamicsecret_controller_test.go
Balaji2198: The “unchanged” was intended to highlight that `LastVaultRotation` didn’t change. To avoid any confusion, I’ll go ahead and remove it.

controllers/vaultdynamicsecret_controller_test.go
Balaji2198: The “unchanged” was intended to highlight that `LastVaultRotation` didn’t change. To avoid any confusion, I’ll go ahead and remove it.

siyer-corp (approved):
LGTM!

PR #1259, opened 2026-04-28
Detect TTL reset for uneven rotation schedules with ttl rollover bug

siyer-corp:
## Description

The awaitVaultSecretRotation loop used `newTTL >= lastSyncTTL` to detect when Vault resets the TTL at a rotation boundary (the TTL rollover bug). This works for even schedules (e.g. "*/1 * * * *") where every interval is identical, but fails for uneven schedules (e.g. "30 3,16 * * *") when a longer interval is followed by a shorter one. At the rotation boundary, Vault resets the TTL upward to the next interval's full duration, but because that duration is shorter than the previous interval's (which lastSyncTTL recorded), the ">=" check never fires and VSO returns stale credentials without retrying.

Add a second detection path: compute the expected countdown TTL based on time elapsed since last sync, and if it is near zero (within a 5-second tolerance) while the observed TTL is still large, treat it as a TTL reset and retry. This correctly handles both even and uneven schedules without affecting mid-cycle remediation (e.g. K8s secret deletion), where expectedTTL is far from zero and no retry is needed.

Add unit tests covering:
- Even schedule with TTL == lastSyncTTL (regression)
- Even schedule with TTL > lastSyncTTL / classic rollover (regression)
- Uneven schedule with TTL < lastSyncTTL / the bug (new)
- Mid-cycle remediation on uneven schedule / no false retry (new)

## PCI review checklist


controllers/vaultdynamicsecret_controller.go
```diff
@@ -639,10 +639,28 @@ func (r *VaultDynamicSecretReconciler) awaitVaultSecretRotation(ctx context.Cont
...
+					//
+					// 1. Classic TTL rollover (even schedule): newTTL >= lastSyncTTL,
+					//    meaning Vault reset it upward to the start of a new period.
+					//
+					// 2. Uneven schedule (e.g. alternating 2-min/1-min): newTTL resets
+					//    to a SHORTER interval, so the ">=" check misses it. We detect
+					//    this by checking whether the expected countdown TTL is near
+					//    zero (we're at the rotation boundary) while the observed TTL
+					//    is still large (it was reset to a new period).
+					const ttlResetTolerance = int64(5)
```
Balaji2198, line 658: Can we add a short comment documenting how we arrived at 5s?

benashz (dismissed):
LGTM

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -1531,6 +1531,293 @@ func TestVaultDynamicSecretReconciler_awaitRotation(t *testing.T) {
...
+			},
+			wantRequestCount: 2,
+		},
+		{
+			// Bug: uneven rotation schedule (e.g. "30 3,16 * * *") where the next
+			// interval is shorter than the last sync TTL. Vault resets the TTL for
+			// the shorter interval (39600 < 46800) before updating LastVaultRotation
+			// due to DB lag. The old ">= lastSyncTTL" check misses this case —
+			// the loop exits immediately with stale credentials instead of retrying.
+			// See: test/integration/modules/vault/VSO Bug
```
Balaji2198: Could this(VSO) be a typo? I could only see this structure `test/integration/modules/vault/`

siyer-corp: It's a reference to a local file, removed it.

Balaji2198 (approved):

PR #1279, opened 2026-06-03
Fix Dynamic Secrets Integration tests.

Balaji2198:
Fixes three issues in the `TestVaultDynamicSecret_InstantUpdates` enterprise integration test:

- `rotation_period` assertion: Enterprise Vault static-creds responses include a `rotation_period` field. Added it to the `assertDynamicSecret` expected map to match actual secret data.

- Parallel test conflict: Both `TestVaultDynamicSecret` and `TestVaultDynamicSecret_InstantUpdates` created a `VaultAuth` named default in the operator namespace, causing an already exists conflict when `testInParallel=true`. Changed `TestVaultDynamicSecret_InstantUpdates` to create its own `VaultAuth` in the app namespace and set `VaultAuthRef` on the VDS spec accordingly.

- Missing Vault event subscription policy: The test's Vault policy did not include permissions for the database event WebSocket endpoint `sys/events/subscribe/database*`, causing `SubscribeToEvents` to fail silently and the EventWatcherStarted Kubernetes event to never be emitted. Added a new `vault_policy.db-events` Terraform resource granting `subscribe` on the DB mount and read on `sys/events/subscribe/database*`, and included it in the token policies for the auth role.

## PCI review checklist

### DIFF
--- a/test/integration/vaultdynamicsecret/terraform/postgres.tf
+++ b/test/integration/vaultdynamicsecret/terraform/postgres.tf
@@ -158,6 +158,7 @@ resource "vault_database_secret_backend_static_role" "postgres-scheduled" {
   ]
 }
 resource "vault_policy" "db" {
+  count     = var.use_events ? 0 : 1
   namespace = local.namespace
   name      = "${local.auth_policy}-db"
   policy    = <<EOT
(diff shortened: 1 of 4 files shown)


test/integration/vaultdynamicsecret/terraform/postgres.tf
```diff
@@ -173,6 +173,21 @@ path "${vault_database_secrets_mount.db.path}/static-creds/${vault_database_secr
 EOT
 }
 
+resource "vault_policy" "db-events" {
+  namespace = local.namespace
```
siyer-corp, line 179: Should we use the pattern from VSS tests which uses `use_events`? Reference [here](https://github.com/hashicorp/vault-secrets-operator/blob/main/test/integration/vaultstaticsecret/terraform/main.tf#L113)

Balaji2198: Updated the tests.

siyer-corp (approved):

PR #1281, opened 2026-06-08
Code for requeue event once even loop exit, registry cleanup an…

harshit-nema:
**Bug #1: No Requeue Event After Event Loop Exit**
Fix: Call notifySubscribersOfStop() to trigger reconciliation when event loop exits and then requeue event is sent.

**Bug #2: Missing Registry Cleanup**
Fix: Add OnStop callback to subscriber that deletes registry entry when WebSocket stops.

**Bug #3: Orphaned Registry Detection**
Fix: Add health check - verify WebSocket exists and is healthy before trusting registry entry. If dead/missing, delete registry entry and create new WebSocket.

**Verified behaviors**
- Websocket read failure triggers reconnect retry
- Reconnect failures are counted per websocket instance
- Threshold is achieved as per declared maxReconnectErrorThreshold.
- Event loop exits on threshold exhaustion
- Requeue event is sent
- Reconcile recreates a new websocket
- New websocket starts cleanly after Vault recovery
- The cycle can repeat multiple times

**Logs collected using command :** 
`kubectl logs -n vault-secrets-operator-system deploy/vault-secrets-operator-controller-manager -f --timestamps \
  | stdbuf -oL tee full-controller-up.log \
  | stdbuf -oL grep '"name":"vaultstaticsecret-sample-tenant-1"' \
  | stdbuf -oL grep -E "SharedWebSocket|Event loop|Failed to reconnect|Too many reconnect|Sent requeue|Created new SharedWebSocket|Successfully reconnected to WebSocket" \
  | tee websocket.log`

image

For database and lease event also backoff retry logic is tested. 
Commands : 
```
kubectl logs -n vault-secrets-operator-system \
  deploy/vault-secrets-operator-controller-manager \
  -c manager -f --timestamps \
  | stdbuf -oL tee full-lease-test.log \
  | stdbuf -oL grep '"name":"vaultdynamicsecret-instant-updates"' \
  | stdbuf -oL grep -E 'SharedWebSocket|Event loop|Failed to reconnect|Too many reconnect|Sent requeue|Created new SharedWebSocket|Successfully reconnected|WebSocket stopped|orphaned' \
  | stdbuf -oL grep '"eventType":"lease"' \
  | tee lease-websocket.log
```
image

`kubectl logs -n vault-secrets-operator-system \
  deploy/vault-secrets-operator-controller-manager \
  -c manager -f --timestamps \
  | stdbuf -oL tee full-database-test.log \
  | stdbuf -oL grep '"name":"vaultdynamicsecret-instant-updates"' \
  | stdbuf -oL grep -E 'SharedWebSocket|Event loop|Failed to reconnect|Too many reconnect|Sent requeue|Created new SharedWebSocket|Successfully reconnected|WebSocket stopped|orphaned' \
  | stdbuf -oL grep '"eventType":"database"' \
  | tee database-websocket.log`

image

## PCI review checklist

### DIFF
--- a/controllers/event_watcher_registry_test.go
+++ b/controllers/event_watcher_registry_test.go
@@ -4,57 +4,617 @@
 package controllers
 
 import (
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
 )
+
+// Helper functions for event watcher registry tests
+
+// createTestRegistry creates a new event watcher registry for testing
+func createTestRegistry() *eventWatcherRegistry {
+	return newEventWatcherRegistry()
+}
+
+// createTestMeta creates test event watcher metadata
+func createTestMeta(generation int64, clientID string) *eventWatcherMeta {
+	return &eventWatcherMeta{
+		LastGeneration: generation,
+		LastClientID:   clientID,
+	}
+}
+
+// createTestNamespacedName creates a test NamespacedName
+func createTestNamespacedName(name, namespace string) types.NamespacedName {
+	return types.NamespacedName{Name: name, Namespace: namespace}
+}
+
+// registerResource registers a resource in the registry with given metadata
+func registerResource(registry *eventWatcherRegistry, objKey types.NamespacedName, generation int64, clientID string) {
+	meta := createTestMeta(generation, clientID)
+	registry.Register(objKey, meta)
+}
+
+// registerMultipleResources registers multiple resources in the registry
+func registerMultipleResources(registry *eventWatcherRegistry, resources []types.NamespacedName, generation int64, clientID string) {
+	for _, res := range resources {
+		registerResource(registry, res, generation, clientID)
+	}
+}
+
+// verifyRegistryCount asserts the registry has expected number of items
+func verifyRegistryCount(t *testing.T, registry *eventWatcherRegistry, expected int, msgAndArgs ...interface{}) {
+	assert.Equal(t, expected, registry.registry.ItemCount(), msgAndArgs...)
+}
+
+// verifyResourceExists asserts a resource exists in the registry
+func verifyResourceExists(t *testing.T, registry *eventWatcherRegistry, objKey types.NamespacedName) *eventWatcherMeta {
+	got, ok := registry.Get(objKey)
+	require.True(t, ok, "expected resource %s to exist in registry", objKey)
+	require.NotNil(t, got, "expected non-nil metadata for resource %s", objKey)
+	return got
+}
+
+// verifyResourceNotExists asserts a resource does not exist in the registry
+func verifyResourceNotExists(t *testing.T, registry *eventWatcherRegistry, objKey types.NamespacedName) {
+	got, ok := registry.Get(objKey)
+	assert.False(t, ok, "expected resource %s to not exist in registry", objKey)
+	assert.Nil(t, got, "expected nil metadata for resource %s", objKey)
+}
+
+// deleteMultipleResources deletes multiple resources from the registry
+func deleteMultipleResources(registry *eventWatcherRegistry, resources []types.NamespacedName) {
+	for _, res := range resources {
+		registry.Delete(res)
+	}
+}
+
+// verifyMultipleResourcesNotExist verifies multiple resources don't exist in registry
+func verifyMultipleResourcesNotExist(t *testing.T, registry *eventWatcherRegistry, resources []types.NamespacedName) {
+	for _, res := range resources {
+		verifyResourceNotExists(t, registry, res)
+	}
+}
+
+// verifyMultipleResourcesExist verifies multiple resources exist in registry
+func verifyMultipleResourcesExist(t *testing.T, registry *eventWatcherRegistry, resources []types.NamespacedName) {
+	for _, res := range resources {
+		verifyResourceExists(t, registry, res)
+	}
+}
 
 // Test the event watcher registry basics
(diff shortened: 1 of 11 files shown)


vault/client.go
```diff
@@ -1086,7 +1087,17 @@ func (c *defaultClient) getOrCreateWebSocket(
...
+		if ws.IsHealthy() {
+			return ws, nil
+		}
+		// WebSocket exists but is dead, remove it from map
+		logger := log.FromContext(ctx).WithValues(
+			"clientID", c.id,
+			"eventType", eventType,
+		)
+		logger.Info("Removing dead WebSocket from map before creating new one")
+		delete(c.websockets, eventType)
```
Balaji2198, line 1160: To ensure the socket is properly closed and keep lifecycle management consistent. Should we use `c.closeWebSocket(eventType)` instead of `delete(c.websockets, eventType)`? This also maintains thread safety.

harshit-nema: L1085 acquires the write lock, so c.closeWebSocket(eventType) should not be called here as L1127 tries to acquire the same lock again. 
So, added ws.Close() to close the ws connection and to maintain thread safety. [Close() doesn't acquire lock, closeWebsocket does].

controllers/vaultstaticsecret_controller.go
```diff
@@ -359,17 +359,29 @@ func (r *VaultStaticSecretReconciler) ensureEventWatcher(ctx context.Context, o
...
+				logger.V(consts.LogLevelDebug).Info("Event subscription already active",
+					"namespace", o.Namespace, "name", o.Name)
+				return nil
+			}
+			// The subscription exists but metadata or vault client has changed, unsubscribe first
+			logger.V(consts.LogLevelDebug).Info("Unsubscribing due to metadata or client change",
+				"namespace", o.Namespace, "name", o.Name)
+			r.unWatchEvents(o, c)
+		} else {
+			// WebSocket is dead or missing - orUnsubscribing due to metadataphaned registry entry detected
```
siyer-corp: This comment seems to be garbled. Can we correct this?

vault/event_subscription.go
```diff
@@ -26,9 +26,15 @@ import (
...
-	maxReconnectElapsedTime = 10 * time.Minute
+	initialReconnectDelay = 1 * time.Second
+	maxReconnectDelay     = 60 * time.Second
+	// maxReconnectElapsedTime is set to 0 to retry indefinitely (until context is cancelled).
+	// This matches the main branch behavior where WebSocket goroutines keep retrying
+	// until they succeed or are explicitly stopped. The SharedWebSocket will keep
+	// retrying to reconnect when Vault is down, and automatically recover when Vault
+	// comes back up, without requiring external triggers (requeue events or periodic
+	// reconciliation).
+	maxReconnectElapsedTime = 0
```
siyer-corp: Can we confirm if this is correct? Does the existing logic retry indefinitely? It looks like we have an [errorThreshold](https://github.com/hashicorp/vault-secrets-operator/blob/main/controllers/vaultstaticsecret_controller.go#L430) set and it breaks the loop post that.

controllers/event_watcher_registry_test.go
```diff
@@ -4,57 +4,617 @@
...
+	// Simulate OnStop callback being called when event loop exits
+	// This mimics the behavior in vaultstaticsecret_controller.go:398-403
+	onStopCallback := func() {
+		// Clean up registry entry
+		registry.Delete(objKey)
+
+		// Send requeue event to trigger reconciliation
+		select {
+		case reconcileCh <- event.GenericEvent{
+			Object: &secretsv1beta1.VaultStaticSecret{
```
siyer-corp: We might need to handle VDS once, we merge the dynamic secrets branch into this. Something to lookout for.

vault/event_subscription.go
harshit-nema: Code added to replace infinite retry with errorCount threshold in eventLoop.

controllers/event_watcher_registry_test.go
harshit-nema: Code added for VDS

vault/client.go
```diff
@@ -1139,14 +1140,35 @@ func (c *defaultClient) getOrCreateWebSocket(
...
 	}
 
 	// Create new SharedWebSocket
 	ws, err := NewSharedWebSocket(ctx, c, eventType)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create shared websocket: %w", err)
 	}
+	// Remove this websocket from the client registry when its event loop stops.
+	ws.onStop = func() {
```
Balaji2198, line 1188: `ws.Close()` will eventually cause the event loop to invoke `onStop()`, while the registry entry is already removed here. Although the `current == ws` check makes this safe. Can we add a comment on when the double cleanup is required?

controllers/vaultstaticsecret_controller_test.go
```diff
@@ -217,3 +219,265 @@ func TestVaultStaticSecretReconciler_Reconcile_vaultIndex(t *testing.T) {
...
+	capturedOnStop()
+
+	// Registry entry must be gone — next reconcile will re-subscribe.
+	_, ok = r.eventWatcherRegistry.Get(key)
+	assert.False(t, ok, "registry entry must be deleted after OnStop fires")
+}
+
+// Test_VSS_unWatchEvents_UnsubscribesAndClearsRegistry verifies that
+// unWatchEvents calls UnsubscribeFromEvents for KV and removes the registry entry.
+func Test_VSS_unWatchEvents_UnsubscribesAndClearsRegistry(t *testing.T) {
```
Balaji2198, line 394: Can we add a test case for `unWatchEvents` where `UnsubscribeFromEvents` returns an error?

harshit-nema: I think it won't be worth to add the test where `UnsubscribeFromEvents` returns an error, as the error is simply the logger and registry deletion is anyways happening with/without error. 
The err != nil in `unWatchEvent()`, has no recoverable behaviour — it logs and moves on. 
Let me know your thoughts.

vault/client.go
harshit-nema: comments updated.

siyer-corp (approved):

PR #1282, opened 2026-06-08
Add Instant updates integration tests

Balaji2198:
Adds two integration tests validating end-to-end `SyncConfig.InstantUpdates=true` behaviour for `VaultDynamicSecret`.

- `TestVaultDynamicSecret_InstantUpdates_DynamicCreds`: Creates a VDS with dynamic (leased) credentials, `InstantUpdates: true`, and `RefreshAfter: 1h`. Revokes the active lease via the Vault API and asserts that `SecretLease.ID` changes within 30s, proving the operator reacted to the lease lifecycle WebSocket event rather than polling.
- `TestVaultDynamicSecret_InstantUpdates_WatcherRestartsOnVaultAuthChange `: Creates a VDS with `InstantUpdates: true` then updates the referenced `VaultAuth`, forcing the operator to acquire a new Vault client. Asserts that a second `EventWatcherStarted` event is emitted (confirming re-subscription on the new client), then force-rotates the role and verifies the new watcher delivers the event end-to-end.

## PCI review checklist


test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1677,3 +1676,506 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
 		}
 	}
 }
+
```
siyer-corp, line 1514: Can we combine this test with `TestVaultDynamicSecret_InstantUpdates`, and also see if we can extract the infra setup for these tests?

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1677,3 +1676,506 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
...
+		TerraformDir: tfDir,
+		Vars: map[string]interface{}{
+			"k8s_config_context":         k8sConfigContext,
+			"k8s_vault_namespace":        k8sVaultNamespace,
+			"k8s_vault_service_account":  "vault",
+			"name_prefix":                "vds-events-dyn",
+			"vault_address":              os.Getenv("VAULT_ADDRESS"),
+			"vault_token":                os.Getenv("VAULT_TOKEN"),
+			"vault_token_period":         120,
+			"vault_db_default_lease_ttl": 60,
```
siyer-corp: Should this be increased to (ex 600), since the test will wait for 60s for the lease to update. With a short lease_ttl, it could get updated with the normal renewal cycle.

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1677,3 +1676,506 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
...
+		TerraformDir: tfDir,
+		Vars: map[string]interface{}{
+			"k8s_config_context":         k8sConfigContext,
+			"k8s_vault_namespace":        k8sVaultNamespace,
+			"k8s_vault_service_account":  "vault",
+			"name_prefix":                "vds-events-auth",
+			"vault_address":              os.Getenv("VAULT_ADDRESS"),
+			"vault_token":                os.Getenv("VAULT_TOKEN"),
+			"vault_token_period":         120,
+			"vault_db_default_lease_ttl": 30,
```
siyer-corp: Same here, should the lease ttl be increased?

Balaji2198: This test uses static credentials (`static-creds/` path, `AllowStaticCreds: true`), so the database mount’s default lease TTL has no impact. Static role credentials rotate only on the role’s `rotation_period` or explicit rotate-role API calls, not via background lease renewal.

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1526,11 +1509,41 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
...
-		require.NoError(t, crdClient.Create(ctx, o))
-		created = append(created, o)
+	return tfOptions, outputs
+}
+
+// TestVaultDynamicSecret_InstantUpdates validates that a VaultDynamicSecret with
+// SyncConfig.InstantUpdates=true receives near-instant credential updates driven
+// by WebSocket events rather than polling. It covers both static role credentials
+// (database* events) and dynamic credentials (lease* events) in parallel subtests
+// that share the same Terraform infrastructure.
```
siyer-corp: Is this wording still valid? There seems to be a separate test for _DynamicCreds

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1677,3 +1700,416 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
...
+		t.Skip("Skipping because this test requires Vault Enterprise >= 1.16.3")
+	}
+
+	operatorNS := os.Getenv("OPERATOR_NAMESPACE")
+	require.NotEmpty(t, operatorNS, "OPERATOR_NAMESPACE is not set")
+
+	ctx := context.Background()
+	crdClient := getCRDClient(t)
+
+	tfOptions, outputs := setupInstantUpdatesInfra(t, "vds-events-dynamic", 120)
```
siyer-corp: I feel 120s might still be low to test if the instant updates is actually used in refreshing the credential and not the regular reconcile loop. Should we increase it to 600 to ensure that instant updates kicks in event if the retries take some time?

test/integration/vaultdynamicsecret_integration_test.go
```diff
@@ -1677,3 +1700,416 @@ func TestVaultDynamicSecret_InstantUpdates(t *testing.T) {
...
+	var vdsBefore secretsv1beta1.VaultDynamicSecret
+	require.NoError(t, backoff.Retry(func() error {
+		if err := crdClient.Get(ctx, objKey, &vdsBefore); err != nil {
+			return err
+		}
+		if vdsBefore.Status.SecretLease.ID == "" {
+			return fmt.Errorf("waiting for SecretLease.ID to be set on %s", objKey)
+		}
+		return nil
+	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Second), 60)))
```
siyer-corp, line 1813: We can reduce the retry seconds here to make it tighter to ensure the rotation is triggered and captured using the instant updates.

siyer-corp (approved):

PR #1284, opened 2026-06-11
VSS: Add key templating for transformed secret data

jenneron:
We want to implement the use of this operator, however we have a lot of secrets that have keys like `db`, `connection`, etc which we need to result in kubernetes secrets with corresponding keys `db.json`, `connection.json` to mount in our pods.

Current transformation only allows to hardcode new name for a specific key, but doesn't allow to dynamically change names of all keys in the secret.

This change allows that by adding `keyTemplate` to secret transformations so source secret data keys can be renamed with a Go template before being written to the destination Kubernetes Secret.

for example:

```
spec:
  destination:
    create: true
    name: test
    overwrite: false
    transformation:
      keyTemplate: "{{ .Name }}.json"
```

This makes:
`db` -> `db.json`
`connection` -> `connection.json`

---

## PCI review checklist

- [ ] ~~If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.~~

- [ ] ~~If applicable, I've documented the impact of any changes to security controls.~~

### DIFF
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -54,6 +54,12 @@ type RolloutRestartTarget struct {
 }
 
 type Transformation struct {
+	// KeyTemplate is a Go text template used to transform each included source
+	// secret data key before writing it to the destination Secret. The template is
+	// executed once per source key with .Name set to the original source key. For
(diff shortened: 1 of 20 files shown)


jenneron:
We no longer need this change, so I'm closing the PR

PR #1285, opened 2026-06-15
Add X-Vault-Index Header Support for Stale-Read Protection and Event Reconciliation

Balaji2198:
When a Vault WebSocket event triggers an instant-update reconcile, extract the `vault_index` from the event metadata and attach it as `X-Vault-Index` on the subsequent read request. This ensures VSO always reads the latest version of the secret, even when connected to a Performance Standby that may not have replicated the write yet.

## PCI review checklist


consts/consts.go
harshit-nema, line 1: Should we need to add the changelog entry in CHANGELOG.md file ?

harshit-nema, line 1: Similary would it be good to update the documentation for X-Vault-Index-feature under docs/ ?

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -2311,3 +2312,86 @@ func Test_resolveEventType(t *testing.T) {
...
+				require.NotNil(t, req.Headers, "expected X-Vault-Index header on request")
+				assert.Equal(t, []string{"42"}, req.Headers[vsoconsts.HeaderVaultIndex])
+			} else {
+				if req.Headers != nil {
+					assert.Empty(t, req.Headers[vsoconsts.HeaderVaultIndex])
+				}
+			}
+		})
+	}
+}
```
harshit-nema, line 2696: Should there be unit testcase on **r.pendingVaultIndex.LoadAndDelete(req.NamespacedName)** ? vault_index gets removed after use (One time consumption)

harshit-nema, line 2696: Should we add a negative testcase too where if instant update is not enabled, reconcile proceed normally without special headers?

controllers/vaultstaticsecret_controller.go
```diff
@@ -154,7 +161,11 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}, nil
 	}
 
-	kvReq, err := newKVRequest(o.Spec)
+	var kvHeaders http.Header
+	if idx, ok := r.pendingVaultIndex.LoadAndDelete(req.NamespacedName); ok {
```
harshit-nema, line 165: Would be better to add test cases for coverage validation check for vault-index. (As vault-index is base64-encoded check, so validation on similar note) ?

controllers/vaultstaticsecret_controller_test.go
harshit-nema, line 1: Would it be better to add complete end-to-end integration test for vaultIndex functionality for VSS where vault_index be fetched from event instance and passed to the Vault API call  headers?

controllers/vaultdynamicsecret_controller.go
harshit-nema, line 1: Would it be better to add complete end-to-end integration test for vaultIndex functionality for VDS where vault_index be fetched from event instance and passed to the Vault API call  headers?

controllers/vaultstaticsecret_controller_test.go
harshit-nema, line 1: Current test case validates newKVRequest. Should we add more unit testcases on : vault_index stored in pendingVaultIndex, LoadAndDelete during reconciliation, X-Vault-Index header sent to Vault, vault_index cleared after use?

harshit-nema (commented):
It would be good to update changelog.md file and documentation for the changes done in the VSO. Provided other comments as well.

controllers/vaultdynamicsecret_controller_test.go
Balaji2198: The existing tests without instant updates already covers this use case.

controllers/vaultstaticsecret_controller_test.go
Balaji2198: The existing integration test of `TestVaultStaticSecret - events-both` covers this scenario, we can not validate specically for `vault_index` in the e2e integration test. However, these scenarios are already covered in the unit testcase `TestVaultStaticSecretReconciler_Reconcile_vaultIndex`

controllers/vaultdynamicsecret_controller.go
Balaji2198: The existing integration test of `TestVaultDynamicSecret_InstantUpdates_DynamicCreds` covers this scenario, we can not validate specically for `vault_index` in the e2e integration test. However, these scenarios are already covered in the unit testcase `TestVaultDynamicSecretReconciler_syncSecret_vaultIndex`

consts/consts.go
Balaji2198: Added the changelog for it

controllers/vaultdynamicsecret_controller.go
```diff
@@ -630,7 +642,7 @@ func (r *VaultDynamicSecretReconciler) awaitVaultSecretRotation(ctx context.Cont
 		backoff.WithMaxInterval(time.Second*2))
 	if err := backoff.Retry(
 		func() error {
-			resp, err = r.doVault(ctx, c, o)
+			resp, err = r.doVault(ctx, c, o, nil)
```
siyer-corp, line 651: Should we pass the headers here as well? Any reason that we're passing nil here?

controllers/vaultstaticsecret_controller_test.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// TestVaultStaticSecretReconciler_Reconcile_vaultIndex verifies the full
+// vault_index lifecycle at the Reconcile level:
+//  1. A vault_index stored in pendingVaultIndex (as routeEvent() would do on an
+//     instant-update event) is forwarded as X-Vault-Index on the KV read.
+//  2. LoadAndDelete consumes the entry exactly once — the map is empty after
+//     the reconcile.
+//  3. A subsequent reconcile without a stored index sends no X-Vault-Index.
+func TestVaultStaticSecretReconciler_Reconcile_vaultIndex(t *testing.T) {
```
siyer-corp, line 131: Can we add a similar test for Dynamic Secrets?

controllers/vaultstaticsecret_controller_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	secretsv1beta1 "github.com/hashicorp/vault-secrets-operator/api/v1beta1"
+	"github.com/hashicorp/vault-secrets-operator/consts"
+	vsoconsts "github.com/hashicorp/vault-secrets-operator/consts"
```
siyer-corp: nit: There are duplicate imports used here, can we just use vsoconsts?

controllers/vaultdynamicsecret_controller_test.go
siyer-corp, line 760: Can we also add docs for the vaultIndex tests that are included in this one?

controllers/vaultdynamicsecret_controller.go
Balaji2198: `nil` is expected here because the `X-Vault-Index` header is already consumed via `LoadAndDelete` during the initial `Reconcile()` call to ensure the first read isn't stale.

The polling inside `awaitVaultSecretRotation` is waiting for a future rotation event, so passing the original event's index wouldn't have any useful effect (and the index is no longer available anyway).

controllers/vaultstaticsecret_controller_test.go
Balaji2198: Done

siyer-corp (approved):

PR #1291, opened 2026-06-23
Added flag to skip RBAC resource creation for restricted-permission VSO installs

Siddharth-Kannan:
Problem: Right now its not possible to install VSO in a cluster with restricted user permissions since it does not support a way to skip RBAC resources creation (ClusterRole, ClusterRoleBinding, Role, RoleBinding).

Fix: Added an option in the charts to skip RBAC resource creation using a flag which gates RBAC charts (defaults to true to keep previous implementation consistent).
## PCI review checklist


siyer-corp:
The implementation looks directionally correct, but I think we should make the flag’s contract explicit before approval.

For controller.rbac.enabled=false, the intended behavior seems to be:

- the chart does not create RBAC authorization resources (ClusterRole, ClusterRoleBinding, Role, RoleBinding)
- the chart still creates ServiceAccounts, the controller Deployment, and hook Jobs 
- equivalent RBAC for those service accounts must be provisioned out of band by a cluster admin
Could we please codify that contract in two places?

Documentation/comment in chart/values.yaml

1. Update the comment for controller.rbac.enabled to explicitly say that disabling it skips only RBAC resources, while ServiceAccounts / Deployment / hook Jobs still render and require pre-provisioned RBAC.
2. Unit tests in test/unit/rbac.bats

In addition to the existing “RBAC does not render” assertions, please add tests showing that with controller.rbac.enabled=false:

- the controller ServiceAccount, Deployment, and pre-delete hook Job from chart/templates/deployment.yaml still render
- the upgrade hook ServiceAccount and Job from chart/templates/hook-upgrade-crds.yaml still render
- the upgrade hook ClusterRole and ClusterRoleBinding from chart/templates/hook-upgrade-crds.yaml do not render

siyer-corp (approved):

PR #1292, opened 2026-06-23
Implement Event routing - Dynamic event subscription

Balaji2198:
VaultDynamicSecret with InstantUpdates previously hardcoded the database event stream for all mounts. This caused silent event misses for LDAP mounts, custom plugin mounts, and any non-database engine, the operator would subscribe to the wrong WebSocket stream and never receive rotation notifications.

This replaces the static hardcoded type with dynamic mount-type resolution, a safe 3-tier fallback strategy, and correct event routing for any Vault secret engine

## PCI review checklist


controllers/vaultdynamicsecret_controller.go
```diff
@@ -1039,21 +1048,78 @@ func (r *VaultDynamicSecretReconciler) ensureEventWatcher(
...
+					ReconcileCh:       r.SourceCh,
+					PendingVaultIndex: &r.pendingVaultIndex,
+				}
+				if err := c.SubscribeToEvents(ctx, vault.EventTypeLease, leaseSubscriber); err != nil {
+					logger.V(consts.LogLevelWarning).Info("Failed to subscribe to lease events", "error", err)
+				}
+			}
 		}
+		return fmt.Errorf("failed to resolve mount type for %q: %w", o.Spec.Mount, mountTypeErr)
+	}
```
siyer-corp, line 1122:
```suggestion
	default:
        // No prior type and lookup failed — subscribing without knowing the
        // type would target the wrong WebSocket and silently miss events.
        // Emit a clear, actionable warning and skip engine-event subscription.
        r.Recorder.Eventf(o, corev1.EventTypeWarning, consts.ReasonEventWatcherError,
            "Cannot determine Vault mount type for %q: %s. "+
                "Ensure the VaultAuth policy grants read on sys/mounts/*. "+
                "Engine-event subscription skipped; will retry next reconcile.",
            o.Spec.Mount, mountTypeErr)

        resultErr := fmt.Errorf("failed to resolve mount type for %q: %w", o.Spec.Mount, mountTypeErr)

        // Lease events are mount-type-independent; still subscribe if applicable.
        if !o.Spec.AllowStaticCreds && currentLeaseID != "" {
            if !(hasMeta && meta.LastLeaseID == currentLeaseID && meta.LastClientID == c.ID()) {
                leaseSubscriber := &vault.Subscriber{
                    ResourceKey:       name,
                    VaultPath:         currentLeaseID,
                    ResourceType:      "VaultDynamicSecret",
                    ReconcileCh:       r.SourceCh,
                    PendingVaultIndex: &r.pendingVaultIndex,
                }
                if err := c.SubscribeToEvents(ctx, vault.EventTypeLease, leaseSubscriber); err != nil {
                    logger.V(consts.LogLevelWarning).Info("Failed to subscribe to lease events", "error", err)
                    // Surface this too, instead of only logging it, so a compound
                    // failure (mount type AND lease subscribe both broken) is visible.
                    resultErr = errors.Join(resultErr, fmt.Errorf("failed to subscribe to lease events: %w", err))
                } else {
                    // Record what we did establish so the next reconcile can detect
                    // "lease-only subscription already active" and skip re-subscribing
                    // on every cycle while the mount-type lookup keeps failing.
                    // LastEventType is left empty so Step 2's staleness check still
                    // forces a full re-subscribe once GetMountType succeeds.
                    r.eventWatcherRegistry.Register(name, &eventWatcherMeta{
                        LastClientID:   c.ID(),
                        LastGeneration: o.GetGeneration(),
                        LastLeaseID:    currentLeaseID,
                        LastEventType:  "",
                    })
                }
            }
        }
        return resultErr
```

controllers/vaultdynamicsecret_controller.go
siyer-corp, line 431: Should we handle the case when mountType is unresolved, and apply a horizon based reconcile so that it does not have to wait for the next scheduled cycle? (In case the permissions are fixed for that mount later)

controllers/vaultdynamicsecret_controller.go
siyer-corp: Can you check on this suggestion? Current code was skipping the r.eventWatcherRegistry.Register call, which is happening in the happy path scenario. Also, the error handling could be improved so that the user does not see unrelated logs.

controllers/vaultdynamicsecret_controller.go
Balaji2198: Updated the logic to handle this scenario.

siyer-corp (approved):

PR #1299, opened 2026-07-02
VDS: Fix rolloutRestart triggered on every reconcile for static roles

itsjiyapatel:
Bug: When using VaultDynamicSecret with allowStaticCreds: false to consume a static role (e.g. ldap/static-creds/my-role), the controller triggered rolloutRestartTargets on every reconciliation interval even when credentials
had not changed. This caused unnecessary pod restarts at every reconcile.

The root cause was that static creds detection was gated on spec.allowStaticCreds. With allowStaticCreds: false the HMAC check was skipped entirely, so every sync returned updated=true and triggered a restart regardless of whether the actual credentials had changed.

Fix: Static creds detection is now based on the Vault response metadata rather than spec.allowStaticCreds. isStaticCreds() inspects the StaticCredsMetaData parsed from the response (last_vault_rotation, rotation_period, rotation_schedule). If those fields are present, checkStaticCredsHMAC() is called to compare the credentials against the stored MAC. If the MAC matches, updated=false is returned and no restart is triggered. Metadata-only changes (ttl, rotation_period, last_vault_rotation) are excluded from the HMAC via staticTransOpt so they never cause a restart on their own.

Tested:
- refreshAfter interval elapsed: pod did not restart (HMAC matched)
- metadata-only change (rotation_period, rotation_policy): pod did not restart
- Vault rotation with ttl=3m: pod restarted correctly after rotation

## PCI review checklist


controllers/vaultdynamicsecret_controller.go
```diff
@@ -494,6 +496,42 @@ func (r *VaultDynamicSecretReconciler) doVault(ctx context.Context, c vault.Clie
...
 }
 
+// checkStaticCredsHMAC performs HMAC-based drift detection for static credentials.
+// Returns true if sync should be skipped (credentials unchanged), false if sync needed.
+func (r *VaultDynamicSecretReconciler) checkStaticCredsHMAC(
+	ctx context.Context,
+	o *secretsv1beta1.VaultDynamicSecret,
+	data map[string][]byte,
+	staticCredsMeta *secretsv1beta1.VaultStaticCredsMetaData,
+	logPrefix string,
```
siyer-corp: `logPrefix` seems to be unused.

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -744,6 +744,106 @@ func TestVaultDynamicSecretReconciler_syncSecret_staticCredsRefreshesStatusOnMat
 	assert.Len(t, vClient.Requests, 2)
 }
 
+// Test that metadata field changes don't trigger sync when allowStaticCreds=false
+// and credentials haven't changed. This is a regression test for the rolling restart
+// bug where configuration changes to rotation settings would unnecessarily restart pods.
+func TestVaultDynamicSecretReconciler_syncSecret_allowStaticCredsFalse_metadataChangesDoNotTriggerSync(t *testing.T) {
```
siyer-corp: Is it possible to convert both static-creds tests into a table-driven test, with shared fixture, varying `AllowedStaticCreds` value+ expected request count

controllers/vaultdynamicsecret_controller.go
```diff
@@ -52,7 +52,9 @@ var (
 	staticTransOpt           = &helpers.SecretTransformationOption{
 		Excludes: []string{
 			fmt.Sprintf("^%s$", strings.Join([]string{
-				"ttl", "rotation_schedule", "rotation_period", "last_vault_rotation", helpers.SecretDataKeyRaw,
+				"ttl", "rotation_schedule", "rotation_period", "rotation_policy",
+				"last_vault_rotation", "next_vault_rotation", "rotation_window",
```
siyer-corp, line 56: Should we add a test for the added exclude fields as well?

siyer-corp (commented):
Thank you, provided some comments. What scenarios have we tested manually to ensure these changes don't cause any regressions?

controllers/vaultdynamicsecret_controller.go
itsjiyapatel: The test `TestVaultDynamicSecretReconciler_syncSecret_staticCreds_metadataChanges` takes care of that

controllers/vaultdynamicsecret_controller_test.go
itsjiyapatel: Added the `TestVaultDynamicSecretReconciler_syncSecret_staticCreds_metadataChanges` test that will take care of this

controllers/vaultdynamicsecret_controller_test.go
```diff
@@ -672,76 +672,159 @@ func newStaticCredsFixture(t *testing.T) *staticCredsFixture {
...
-// Exercises syncSecret in isolation.
-//
-// Status before: TTL=600  |  Vault response: TTL=540 (same LastVaultRotation)
-func TestVaultDynamicSecretReconciler_syncSecret_staticCredsRefreshesStatusOnMatchingHMAC(t *testing.T) {
+// Table-driven test for static credentials metadata changes.
+// Tests that metadata field changes don't trigger sync when credentials haven't changed,
+// for both allowStaticCreds=true and allowStaticCreds=false.
+// This is a regression test for the rolling restart bug where configuration changes
+// to rotation settings would unnecessarily restart pods.
+func TestVaultDynamicSecretReconciler_syncSecret_staticCreds_metadataChanges(t *testing.T) {
```
Balaji2198, line 681: This testcase majorly covers the scenario in the `syncSecret`. Can we also add a Reconcile-level regression test for the `allowStaticCreds=false` + `rolloutRestartTargets` scenario to verify that unchanged credentials do not trigger a rollout restart on periodic reconciles? This would provide end-to-end coverage of the reported issue.

Balaji2198 (changes requested):

controllers/vaultdynamicsecret_controller_test.go
itsjiyapatel: added TestVaultDynamicSecretReconciler_Reconcile_forceSyncStaticCredsUsesRefreshedTTL test case

Balaji2198 (dismissed):

siyer-corp (approved):
LGTM

PR #1315, opened 2026-07-23
Remove the aws-sdk v1 dependency

tejashwiniingalagi:
## PCI review checklist

Screenshot 2026-08-03 at 2 46 26 PM
Screenshot 2026-08-03 at 2 45 45 PM

Screenshot 2026-08-03 at 2 45 25 PM
96e-4845-ae5e-009573be55e5" />


credentials/vault/aws.go
```diff
@@ -21,6 +34,152 @@ import (
...
+			return stsSigningEndpoint{}, fmt.Errorf("invalid custom STS endpoint URL %q", endpointURL)
+		}
+		return stsSigningEndpoint{
+			requestURL:    uri.String(),
+			signingName:   stsSigningName,
+			signingRegion: region,
+		}, nil
+	}
+	return stsSigningEndpoint{
+		requestURL:    fmt.Sprintf("https://sts.%s.amazonaws.com", region),
```
siyer-corp: Can we use AWS SDK v2's native resolver instead of hardcoded string formatting here?

credentials/vault/aws.go
```diff
@@ -144,19 +304,18 @@ func (l *AWSCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.
...
 	}
 
-	// TODO: convert logr to something compatible with hclog for use in the
-	// awsutil functions
 	config.Logger = hclog.Default()
 	config.Logger.SetLevel(hclog.Debug)
 
-	creds, err := config.GenerateCredentialChain(awsutil.WithSkipWebIdentityValidity(true))
+	// GenerateCredentialChain returns *aws.Config (SDK v2).
+	awsCfg, err := config.GenerateCredentialChain(ctx)
```
siyer-corp, line 328: The [previous code](https://github.com/hashicorp/vault-secrets-operator/blob/main/credentials/vault/aws.go#L152) bypassed web identity token validity checking via awsutil.WithSkipWebIdentityValidity(true). Is there a reason why we're skipping it? Or is it already handled somewhere?

tejashwiniingalagi: this option doesn't exist in awsutil v2 . The v2 SDK handles web identity tokens differently (it calls GetIdentityToken() lazily at sign time, so there's no upfront validity check to skip).

credentials/vault/aws.go
tejashwiniingalagi: done

credentials/vault/aws_test.go
```diff
@@ -65,3 +76,308 @@ func Test_getIRSAConfig(t *testing.T) {
...
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tc.wantURL, ep.requestURL)
+			assert.Equal(t, tc.wantRegion, ep.signingRegion)
+			assert.Equal(t, stsSigningName, ep.signingName)
+		})
+	}
+}
+
+func Test_buildSignedGetCallerIdentityRequest(t *testing.T) {
```
Balaji2198, line 141: Could we add a test covering temporary AWS credentials with a non-empty `SessionToken`? This is important for IRSA, node role, and instance profile authentication flows, and would help prevent regressions in forwarding/signing `X-Amz-Security-Token`.

credentials/vault/aws.go
```diff
@@ -21,6 +34,165 @@ import (
...
+// generateLoginData builds the Vault AWS IAM login payload by constructing and
+// signing a sts:GetCallerIdentity HTTP request using AWS SDK v2.
+// It replicates the behaviour of the now-removed awsutil.GenerateLoginData from
+// go-secure-stdlib/awsutil v0.
+func generateLoginData(ctx context.Context, awsConfig *aws.Config, headerValue, stsEndpoint string) (map[string]interface{}, error) {
+	if awsConfig == nil || awsConfig.Credentials == nil {
+		return nil, fmt.Errorf("AWS credentials are not configured")
+	}
+
+	credentials, err := awsConfig.Credentials.Retrieve(ctx)
```
Balaji2198, line 89: Could we validate the retrieved credentials with `credentials.HasKeys()` before proceeding? If `Retrieve()` returns empty credentials without an error, the request could be signed with an empty `AccessKeyID`, resulting in a silent authentication failure.

credentials/vault/aws.go
```diff
@@ -21,6 +34,165 @@ import (
...
+	req.Header.Set("Content-Type", stsContentType)
+	if headerValue != "" {
+		req.Header.Set(iamServerIDHeader, headerValue)
+	}
+
+	signingRegion := endpoint.signingRegion
+	if signingRegion == "" {
+		signingRegion = region
+	}
+	if signingRegion == "" {
```
Balaji2198, line 179: Could we simplify this fallback chain? `resolveSTSSigningEndpoint` and the `generateLoginData` already guarantee a non-empty `signingRegion`, so these fallbacks appear unreachable in the production path.

credentials/vault/aws.go
```diff
@@ -144,19 +310,22 @@ func (l *AWSCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.
...
 	config.Logger = hclog.Default()
 	config.Logger.SetLevel(hclog.Debug)
 
-	creds, err := config.GenerateCredentialChain(awsutil.WithSkipWebIdentityValidity(true))
+	// GenerateCredentialChain returns *aws.Config (SDK v2).
+	// Note: the awsutil v0 option WithSkipWebIdentityValidity(true) has no
+	// equivalent in v2. The SDK v2 stscreds.WebIdentityRoleProvider retrieves
+	// the token lazily at signing time via GetIdentityToken(), so there is no
+	// upfront validity window to bypass.
+	awsCfg, err := config.GenerateCredentialChain(ctx)
```
siyer-corp, line 321: Could we test AWS authentication with only irsaServiceAccount configured - no AWS credentials, web-identity environment variables, or shared profiles?

Expected: VSO calls AssumeRoleWithWebIdentity using the requested service-account token and role.

Concern: the new wrapper may ignore the inline token and use node credentials instead, or fail if node credentials are unavailable.

siyer-corp: Could we also test with AWS_ROLE_ARN set and node/instance metadata credentials available, but no static credentials, shared profiles, or web-identity configuration?

Expected: VSO uses the node credentials to assume the configured role, then uses the assumed-role credentials for Vault authentication.

Concern: the new wrapper may skip AssumeRole and authenticate directly using the node credentials

tejashwiniingalagi: > Could we test AWS authentication with only irsaServiceAccount configured - no AWS credentials, web-identity environment variables, or shared profiles?
> 
> Expected: VSO calls AssumeRoleWithWebIdentity using the requested service-account token and role.
> 
> Concern: the new wrapper may ignore the inline token and use node credentials instead, or fail if node credentials are unavailable.

Tested this, Root cause, for the record: awsutil supplies the inline token via
`config.WithWebIdentityRoleCredentialOptions`, which only *customizes* a
`WebIdentityRoleProvider` the SDK has already decided to build from
`AWS_WEB_IDENTITY_TOKEN_FILE`. With no such env var no provider is built and the option is
silently discarded.

credentials/vault/aws.go
```diff
@@ -206,6 +379,110 @@ func (l *AWSCredentialProvider) getCredentialsConfig(credsSecret *corev1.Secret,
...
+				for _, k := range sortedKeys(credsConfig.RoleTags) {
+					o.Tags = append(o.Tags, ststypes.Tag{
+						Key:   aws.String(k),
+						Value: aws.String(credsConfig.RoleTags[k]),
+					})
+				}
+			})
+	}
+
+	awsCfg.Credentials = aws.NewCredentialsCache(provider)
```
siyer-corp: This unconditionally replaces the selected credentials whenever a role ARN is configured, including one inherited from the operator’s environment. On main, valid Secret, environment, and shared credentials take precedence over role providers.

For example, when a VaultAuth uses secretRef but the operator pod has IRSA environment variables, this override discards the Secret credentials and authenticates using the operator’s role instead. Similarly, environment credentials plus AWS_ROLE_ARN now use assumed-role credentials instead of the environment credentials.

Could we preserve the existing credential precedence and add tests for these combinations? Please verify which credentials sign the Vault login request, not just whether login succeeds. Let's add tests for these scenarios in aws_test.go

credentials/vault/aws_test.go
```diff
@@ -65,3 +94,1448 @@ func Test_getIRSAConfig(t *testing.T) {
 		})
 	}
 }
+
+func Test_resolveSTSSigningEndpoint(t *testing.T) {
```
Balaji2198, line 100: Can we add a test to verify the full precedence hierarchy: `explicit spec > AWS_REGION > AWS_DEFAULT_REGION > default (us-east-1)`

credentials/vault/aws_test.go
```diff
@@ -65,3 +94,1448 @@ func Test_getIRSAConfig(t *testing.T) {
...
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tc.wantURL, ep.requestURL)
+			assert.Equal(t, tc.wantRegion, ep.signingRegion)
+			assert.Equal(t, stsSigningName, ep.signingName)
+		})
+	}
+}
+
+func Test_buildSignedGetCallerIdentityRequest(t *testing.T) {
```
Balaji2198: Can we add a testcase for an AWS error response (e.g., 403 AccessDenied, ExpiredToken) from STS causes `GetCreds` to return a non-nil error for both node-role and IRSA paths?

tejashwiniingalagi: done

credentials/vault/aws_test.go
tejashwiniingalagi: done

credentials/vault/aws.go
tejashwiniingalagi: taken care

credentials/vault/aws.go
```diff
@@ -144,19 +314,23 @@ func (l *AWSCredentialProvider) GetCreds(ctx context.Context, client ctrlclient.
...
 	config.Logger = hclog.Default()
 	config.Logger.SetLevel(hclog.Debug)
 
-	creds, err := config.GenerateCredentialChain(awsutil.WithSkipWebIdentityValidity(true))
+	// GenerateCredentialChain returns *aws.Config (SDK v2).
+	// Note: the awsutil v0 option WithSkipWebIdentityValidity(true) has no
+	// equivalent in v2. The SDK v2 stscreds.WebIdentityRoleProvider retrieves
+	// the token lazily at signing time via GetIdentityToken(), so there is no
+	// upfront validity window to bypass.
+	awsCfg, err := config.GenerateCredentialChain(ctx)
```
siyer-corp: The shared AWS credentials files fail to load. VSO leaves Filename empty, and the v2 wrapper overrides normal file lookup with an empty path.

Authentication using AWS_SHARED_CREDENTIALS_FILE fails for both default and named profiles. Main supports these configurations. Can we test and confirm this?

tejashwiniingalagi: the issue exists on main too

tejashwiniingalagi: fixed

tejashwiniingalagi: works for main, fixed in this branch too

credentials/vault/aws.go
```diff
@@ -177,10 +353,24 @@ func (l *AWSCredentialProvider) getCredentialsConfig(credsSecret *corev1.Secret,
...
 	if l.authObj.Spec.AWS.STSEndpoint != "" {
-		config.STSEndpoint = l.authObj.Spec.AWS.STSEndpoint
+		config.STSEndpointResolver = &customSTSEndpointResolver{endpointURL: l.authObj.Spec.AWS.STSEndpoint}
 	}
 	if l.authObj.Spec.AWS.IAMEndpoint != "" {
-		config.IAMEndpoint = l.authObj.Spec.AWS.IAMEndpoint
+		config.IAMEndpointResolver = &customIAMEndpointResolver{endpointURL: l.authObj.Spec.AWS.IAMEndpoint}
+	}
+
+	// awsutil passes CredentialsConfig.Filename to
```
siyer-corp: When both environment access keys and a valid shared credentials profile exist, the v2 wrapper explicitly selects the profile, causing its credentials to override the environment credentials. Main uses the environment credentials first. Can we get this precendence fixed and add the sub-test in the appropriate test case?

credentials/vault/aws.go
```diff
@@ -206,6 +396,154 @@ func (l *AWSCredentialProvider) getCredentialsConfig(credsSecret *corev1.Secret,
 	return config, nil
 }
 
+// applyCredentialsOverride replaces awsCfg.Credentials with a role-assumption
```
siyer-corp: Could we shorten the comments to describe the function contract and non-obvious constraints, and move the detailed migration/dependency investigation into the PR description?

siyer-corp (commented):
The test refactor is much improved. As a follow-up cleanup, could we simplify tables that repeat identical inputs with separate assertion callbacks? Expected-value fields and shared assertions would make the scenarios easier to compare.

Could we also add a changelog entry for this?

credentials/vault/aws.go
```diff
@@ -206,6 +390,152 @@ func (l *AWSCredentialProvider) getCredentialsConfig(credsSecret *corev1.Secret,
...
+//     CredentialsConfig.RoleARN from the operator pod's own AWS_ROLE_ARN, set
+//     whenever VSO itself runs under IRSA, so this stands down when the chain
+//     already resolved static credentials - they outrank role assumption, and
+//     acting on an inherited role would authenticate as the operator rather
+//     than the configured identity. An irsaServiceAccount on the VaultAuth is
+//     an explicit request and still wins.
+//   - The provider resolves lazily, so a failed role assumption surfaces as an
+//     error from GetCreds. awsutil v0 logged a warning and continued with the
+//     unassumed credentials; failing loudly is deliberate, since a silent
+//     fallback would authenticate under an unexpected identity.
```
siyer-corp: This intentionally changes main’s behavior: a failed AssumeRole could previously fall back to node credentials, whereas it now fails authentication. Should we preserve compatibility in this SDK migration and address stricter failure behavior separately?

CHANGELOG.md
```diff
@@ -1,13 +1,20 @@
...
 
 Fix:
+* VaultAuth: fix AWS authentication with `spec.aws.irsaServiceAccount`. The IRSA ServiceAccount token was not used to call `sts:AssumeRoleWithWebIdentity`, so logins silently fell through to EC2 instance metadata credentials, or failed outright when none were available ([#1315](https://github.com/hashicorp/vault-secrets-operator/pull/1315))
+* VaultAuth: fix AWS authentication using a shared credentials file. `AWS_SHARED_CREDENTIALS_FILE` and the default `~/.aws/credentials` were both ignored, for the default and named profiles alike, causing credential resolution to fail ([#1315](https://github.com/hashicorp/vault-secrets-operator/pull/1315))
+* VaultAuth: preserve AWS credential precedence. Credentials from `spec.aws.secretRef`, the environment, and shared profiles are no longer displaced by a role assumed from the operator pod's own `AWS_ROLE_ARN`, and environment credentials again take precedence over a shared profile ([#1315](https://github.com/hashicorp/vault-secrets-operator/pull/1315))
```
siyer-corp: These are issues that were raised during the changes being made, not existing bugs that are fixed.

CHANGELOG.md
```diff
@@ -1,13 +1,20 @@
 ## Unreleased
 
 Enhancements:
+* VaultAuth: AWS auth now honors `spec.aws.stsEndpoint` when retrieving AWS credentials, not just when signing the Vault login request. Previously the configured endpoint was ignored during credential retrieval, so IRSA and assumed-role flows contacted the default AWS STS endpoint. ([#1315](https://github.com/hashicorp/vault-secrets-operator/pull/1315))
```
siyer-corp: Same here. The existing flow already has this.

credentials/vault/aws.go
tejashwiniingalagi: taken care of comments. will update the PR description too

PR #1323, opened 2026-07-31
FIX: Return error on Vault failure paths to prevent VSS from dropping off work queue

tejashwiniingalagi:
Description:

Problem
After a Vault HA event (leader re-election, pod restart), VSO emits transient errors like:

"local node not active but active cluster node not found"

This is expected and self-resolving — once the new leader is established, Vault recovers. However, VSO permanently stopped reconciling the affected VaultStaticSecret resources. Days later, secrets were stale with zero controller log entries since the failure. Restarting the VSO pod alone did not recover them; the only workaround was a manual spec-touch to force a re-enqueue.

Root Cause
Two error return paths in VaultStaticSecretReconciler.Reconcile returned nil error alongside RequeueAfter:

// Before — both error paths
return ctrl.Result{RequeueAfter: horizon}, nil

controller-runtime treats a nil error + RequeueAfter as a one-shot delayed retry. Once that timer fires and Vault has not yet recovered, no further retries are scheduled — the resource is silently dropped from the work queue. 

A secondary issue: when hmacSecretData=true and the Vault read succeeds but the secret data is unchanged (doSync=false), the SecretSynced condition was never updated — leaving it permanently False even after full recovery.

Fix
1. Return the non-nil error on both Vault error paths (ClientFactory.Get and c.Read):

// After
return ctrl.Result{RequeueAfter: horizon}, err

Returning a non-nil error alongside RequeueAfter tells controller-runtime to place the resource back into the rate-limiter queue. It will keep retrying with exponential backoff via BackOffRegistry (5s → 60s plateau) indefinitely until Vault recovers — no spec-touch required.

2. Write SecretSynced=True in the doSync=false branch:

When HMAC comparison shows the secret data is unchanged, the reconciler previously wrote no condition at all — leaving a stale SecretSynced=False from the prior failure. Now both the sync and no-sync paths always write the condition, accurately reflecting the healthy post-recovery state.

## PCI review checklist

Screenshot 2026-08-04 at 8 59 22 AM


controllers/vaultstaticsecret_controller.go
```diff
@@ -289,6 +286,14 @@ func (r *VaultStaticSecretReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	} else {
 		logger.V(consts.LogLevelDebug).Info("Secret sync not required")
+		// Secret data is unchanged; use a distinct Reason so observers can
+		// tell the difference between "data written now" and "verified up-to-date".
+		// This also overwrites any previous SecretSynced=False error condition
+		// because updateConditions deduplicates on Type, replacing the old entry.
+		conditions = append(conditions,
```
siyer-corp, line 291: Writing a condition in the doSync=false branch seems correct. But Reason="SecretUpToDate" is a breaking change. Before this PR, every healthy VSS always reported Reason="Synced". Changing this for all hmacSecretData=true resources will silently break scripts, alerting rules, and Argo CD health checks that match on that reason value. What if we reuse Reason="Synced" with a different message?

controllers/vaultstaticsecret_controller_test.go
```diff
@@ -0,0 +1,294 @@
...
+	err error
+}
+
+func (f *errClientFactory) Get(_ context.Context, _ client.Client, _ client.Object) (vault.Client, error) {
+	return nil, f.err
+}
+
+func (f *errClientFactory) RegisterClientCallbackHandler(vault.ClientCallbackHandler) {}
+
+// errReadVaultClient is a vault.Client whose Read always returns an error.
```
siyer-corp: A few suggestions for the test

-  Can we use a single table-driven test for related scenarios instead of separate test functions
- Can we construct the reconciler inline rather than behind helper functions, so the test setup is immediately visible
- Can we add an err field to the existing reconcileTestClientFactory (defined in vaultdynamicsecret_controller_test.go, accessible across the package since both files share package controllers) instead of introducing a new errClientFactory type that duplicates the same concept?
- Replace okReadVaultClient / errReadVaultClient with MockRecordingVaultClient with canned responses/errors, consistent with how VDS tests fake Vault client behaviour?

controllers/vaultstaticsecret_controller.go
tejashwiniingalagi: I have put back the reason as synced

controllers/vaultstaticsecret_controller_test.go
tejashwiniingalagi: done

siyer-corp:
Can we update the title of the PR with the actual fix being made?

siyer-corp (approved):
LGTM!

sanvikam79 (approved):

PR #1332, opened 2026-08-17
fix: Resolve concurrency, websocket leaks, and client optimizations.

Balaji2198:
- Goroutine leak: `NewSharedWebSocket` now derives from the caller's ctx instead of `context.Background()`, so event-loop goroutines exit on operator shutdown
- Data race: added connMu `sync.RWMutex` to protect `ws.conn` across goroutines
- Duplicate Vault calls: `GetMountType` now holds a single write-lock across the cache-miss path, preventing concurrent sys/mounts requests
- Blocking callback: `vaultClientCallback` sends to SourceCh non-blocking (select/default) to match routeEvent's existing pattern
- Stale reads on retry: `awaitVaultSecretRotation` now threads `X-Vault-Index` headers through its backoff retry loop; redirect cycle detection added to WebsocketClient.Connect; context.TODO() replaced with real context in `UnsubscribeFromEvents`; metric Help string corrected; dead `buildLeaseEventKey` function removed

## PCI review checklist

### DIFF
--- a/controllers/vaultstaticsecret_controller.go
+++ b/controllers/vaultstaticsecret_controller.go
@@ -500,7 +500,7 @@ func (r *VaultStaticSecretReconciler) SetupWithManager(mgr ctrl.Manager, opts co
 		},
 	)
 
-	r.SourceCh = make(chan event.GenericEvent)
+	r.SourceCh = make(chan event.GenericEvent, 4)
 	r.eventWatcherRegistry = newEventWatcherRegistry()
 	ctrlmetrics.Registry.MustRegister(prometheus.NewGaugeFunc(
 		prometheus.GaugeOpts{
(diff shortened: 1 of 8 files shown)


controllers/vaultstaticsecret_controller.go
```diff
@@ -500,14 +500,14 @@ func (r *VaultStaticSecretReconciler) SetupWithManager(mgr ctrl.Manager, opts co
 		},
 	)
 
-	r.SourceCh = make(chan event.GenericEvent)
+	r.SourceCh = make(chan event.GenericEvent, 4)
```
siyer-corp, line 503: Should we add this for VDS as well?

siyer-corp (approved):

PR #1352, opened 2026-09-13
Migrate from kube-rbac-proxy to Controller-Runtime’s feature WithAuthenticationAndAuthorization

Siddharth-Kannan:
Images provided under _gcr.io/kubebuilder/_ are unavailable since March 18, 2025.This migration from `kube-rbac-proxy `drops the sidecar and protects /metrics in-process with controller-runtime's `filters.WithAuthenticationAndAuthorization`, which runs the same TokenReview/SubjectAccessReview checks. The manager now serves /metrics over HTTPS on 8443 itself.

- `--metrics-secure` and `--enable-http2` flags (both default false); HTTP/2 off by default on the metrics and webhook servers.
-  The named port had to move onto the manager. The Service uses targetPort: https, a name the sidecar owned. Miss it and the Service resolves to no endpoints — metrics break silently.

Testing:
- *Chart rendering* — `deployment.bats` 89/89; confirmed the rendered Deployment has exactly one
  container carrying `--metrics-secure=true` and `containerPort: 8443` named `https`, and that
  the metrics Service still resolves via `targetPort: https`
- *Authn/authz (envtest)* — no token → `401`, malformed token → `500`, valid token without the
  reader role → `403`, token with the role → `200`
- *TLS/HTTP2* — all four `--metrics-secure` × `--enable-http2` combinations, including confirming
  hostile h2c clients are rejected and plain HTTP against the TLS port fails
- *Integration suite* — full run on kind via both the kustomize and Helm paths, and on OpenShift
  `amd64`/`s390x`.
- *Live scrape* — `Service` endpoints resolve to `…:8443`, metrics returned over HTTP/1.1; on
  OpenShift the cluster's own `prometheus-k8s` ServiceAccount scrapes successfully

## PCI review checklist


chart/values.yaml
```diff
@@ -161,28 +161,6 @@ controller:
         # @type: boolean
         edit: false
 
-  # Settings related to the kubeRbacProxy container. This container is an HTTP proxy for the
-  # controller manager which performs RBAC authorization against the Kubernetes API using SubjectAccessReviews.
-  kubeRbacProxy:
```
siyer-corp, line 166: Can we pair this removal with a `{{- fail }}` guard in `chart/templates/deployment.yaml`? Deleting the kubeRbacProxy block means anyone still setting `controller.kubeRbacProxy.*` gets no error. Helm upgrade has no --strict, so stale values are silently ignored. Someone who'd tuned kubeRbacProxy.resources or repointed the image to an internal registry would have no idea it stopped applying.

test/unit/deployment.bats
siyer-corp, line 685: Can we add a test asserting --metrics-secure=true and --metrics-bind-address=:8443 are present on the manager container? The existing tests only check argument indices, so if someone later drops the secure flag here or the FilterProvider, we'd ship an unauthenticated metrics endpoint with every test still green.

config/default-openshift/manager_ubi_auth_proxy_patch.yaml
```diff
@@ -1,9 +1,12 @@
...
-# proxy for the controller manager, it performs RBAC authorization against the
-# Kubernetes API using SubjectAccessReviews.
+# This patch sets OpenShift-specific configuration for the controller manager.
+#
+# NOTE: The kube-rbac-proxy sidecar (previously pinned here to a UBI-based
+# registry.redhat.io image) has been removed. The metrics endpoint is now
+# protected in-process by controller-runtime's
+# filters.WithAuthenticationAndAuthorization, so no proxy image is required.
 apiVersion: apps/v1
 kind: Deployment
```
siyer-corp: Do we need to keep this file anymore?

siyer-corp (commented):
Provided some initial comments.

chart/templates/deployment.yaml
```diff
@@ -3,6 +3,19 @@
...
 */}}
 
+{{/*
+Fail if controller.kubeRbacProxy.* is still set. The sidecar has been removed, and
+Helm silently ignores unknown values, so without this guard a stale setting would
+appear to apply but have no effect.
+*/}}
+{{- if hasKey .Values.controller "kubeRbacProxy" }}
+{{- fail (printf "\n%s\n%s\n%s"
+  "controller.kubeRbacProxy.* is no longer supported."
```
vijayavelsekar: Do we need a test to cover this new code ?

chart/templates/deployment.yaml
```diff
@@ -46,28 +59,11 @@ spec:
...
-          name: https
-          protocol: TCP
-        resources: {{- toYaml .Values.controller.kubeRbacProxy.resources | nindent 10 }}
-        securityContext:
-          {{- toYaml .Values.controller.securityContext | nindent 10 }}
       - name: manager
         args:
         - --health-probe-bind-address=:8081
-        - --metrics-bind-address=127.0.0.1:8080
+        - --metrics-bind-address=:8443
```
Balaji2198, line 65: Can users override reserved metrics flags (like `--metrics-secure` or `--metrics-bind-address`) via `extraArgs`? If so, does this allow them to expose unauthenticated metrics?

Siddharth-Kannan: extraArgs renders after the hardcoded flags, and Go's flag package is last-wins, so a user-supplied --metrics-secure=false takes precedence. So it is possible.

Only --metrics-secure=false removes authentication. The endpoint then serves plain HTTP with no auth on 8443, still reachable through the Service. Overriding --metrics-bind-address alone keeps HTTPS + authn/authz and just breaks the Service path. So exposing metrics requires an explicit opt-out; it can't happen by accident.

